PR #20, opened 2022-07-12
openSUSE refactor

dcermak:
This is proposal how to remove the need for many:
```python
if self.is_opensuse:
    return "foo"
return "bar"
```
(the PR will not merge cleanly as I rebased it on main)

### DIFF
--- a/src/bci_build/templates.py
+++ b/src/bci_build/templates.py
@@ -70,10 +70,10 @@ KIWI_TEMPLATE = jinja2.Template(
             <label name="org.opensuse.reference" value="{{ image.reference }}"/>
             <label name="org.openbuildservice.disturl" value="%DISTURL%"/>
 {% if not image.is_opensuse %}            <label name="com.suse.supportlevel" value="{{ image.support_level }}"/>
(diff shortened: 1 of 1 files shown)


dcermak:
I agree that it's slightly larger, but when compared to `main` it's not that much of a difference (and most of that is documentation).
```
❯ git diff origin/main origin/opensuse --stat
 src/bci_build/package.py   | 77 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++------------------
 src/bci_build/templates.py | 20 ++++++++++----------
 2 files changed, 69 insertions(+), 28 deletions(-)

❯ git diff origin/main opensuse-refactor --stat
 src/bci_build/package.py   | 109 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++-----------------
 src/bci_build/templates.py |  20 ++++++++++----------
 2 files changed, 102 insertions(+), 27 deletions(-)

```

Vogtinator:
I don't mind either, so feel free to merge. Code looks good, but I haven't worked with this style of python before.

PR #25, opened 2022-07-28
Update to rust1.62

Firstyear:


dcermak:
Thanks, when will Rust 1.62 hit SLE?

Firstyear:
It's been submitted, not sure on when it will be accepted ...

dcermak:
Please talk to maintenance about this, if they are not aware that this should go out soon, it'll stay in their queue for quite a bit.

PR #27, opened 2022-08-11
Add PROMETHEUS_CONTAINERS

witekest:
Add entry for Prometheus container images.


src/bci_build/package.py
```diff
@@ -1364,6 +1364,22 @@ def _get_openjdk_kwargs(
     )
 ]
 
+PROMETHEUS_CONTAINERS = [
+    ApplicationStackContainer(
```
dcermak, line 1368: It needs a `version`

witekest: Done

src/bci_build/package.py
```diff
@@ -1364,6 +1364,23 @@ def _get_openjdk_kwargs(
...
+        package_name="prometheus-image",
+        os_version=os_version,
+        is_latest=os_version in CAN_BE_LATEST_OS_VERSION,
+        name="prometheus",
+        pretty_name="Prometheus",
+        package_list=["golang-github-prometheus-prometheus"],
+        version="2.32.1",
```
dcermak, line 1375: This is not going to work, this is just the version in Factory, but in SLE we are at 2.18 at the moment.

Please use a replacement for this like it is e.g. done with the postgress container. That uses https://github.com/openSUSE/obs-service-replace_using_package_version in the background.

witekest: Where does it pull the packages from? We have 2.32.1 in [SLE-15-SP5:GA](https://build.suse.de/package/show/SUSE:SLE-15-SP5:GA/golang-github-prometheus-prometheus).

src/bci_build/package.py
```diff
@@ -1369,6 +1369,31 @@ def _get_openjdk_kwargs(
...
+PROMETHEUS_PACKAGE_NAME = "golang-github-prometheus-prometheus"
+PROMETHEUS_CONTAINERS = [
+    ApplicationStackContainer(
+        package_name="prometheus-image",
+        os_version=os_version,
+        is_latest=os_version in CAN_BE_LATEST_OS_VERSION,
+        name="prometheus",
+        pretty_name="Prometheus",
+        package_list=[PROMETHEUS_PACKAGE_NAME],
+        version="%%prometheus_version%%",
```
dcermak, line 1381: Hm, this is in hindsight not ideal, because now the container name is very weird via the cli. Could you maybe change the version to `2` and add the actual version into the `additional_versions` list?

witekest: Could you elaborate on this? The container image name is _prometheus_ and does not include version number, right?  
99.9% of users have Prometheus 2.x and adding version 2 to the image name would be confusing, I guess.  
Ideally, I'd like to keep the same image name and version tags as the [upstream images](https://hub.docker.com/r/prom/prometheus/tags).

dcermak: I wanted to keep the "main" version something static, because the dockerfile generator would allow you to write the package to obs via `$name-$version-$os_version`, which would result in the very ugly `Prometheus-%%prometheus_version%%-sp4` name. However, this is a purely non-technical and rather silly limitation, which is why I have created #33.
Could you please rebase on main and add the parameter `version_in_uid=False` to the constructor of `ApplicationStackContainer`?

witekest: Done

dcermak:
Thanks!

dcermak:
Submitted to the devel projects: https://build.opensuse.org/request/show/999001, https://build.opensuse.org/request/show/999003

PR #45, opened 2022-09-26
Add OCI registry container image

dirkmueller:


src/bci_build/package.py
```diff
@@ -473,11 +480,35 @@ def packages(self) -> str:
...
+    def volumes_kiwi(self) -> str:
+        """The volumes for this image as xml elements that are inserted into
+        a container.
+        """
+
+        res = ""
+        if self.volumes:
+            res += "\n" + " " * 8 + f"<volumes>\n"
+            for v in self.volumes:
+                res += " " * 10 + f'<volume name="{v}\ />\n'
```
dcermak, line 493: Typo?
```suggestion
                res += " " * 10 + f'<volume name="{v}" />\n'
```

src/bci_build/package.py
```diff
@@ -1737,6 +1772,42 @@ def init_container_kwargs(os_version: OsVersion):
...
+                "ca-certificates-mozilla",
+            )
+        ],
+        entrypoint=["/usr/bin/registry"],
+        cmd=["serve", "/etc/registry/config.yml"],
+        build_recipe_type=BuildType.KIWI,
+        volumes=["/var/lib/docker-registry"],
+        exposes_tcp=[5000],
+    )
+    for os_version in ALL_OS_VERSIONS
```
dcermak, line 1808: Do you really want to create the container for 15 SP3 as well and not just SP4?

src/bci_build/package.py
dirkmueller: yes, thanks

src/bci_build/package.py
dirkmueller: changed to `CAN_BE_LATEST_OS_VERSION`

PR #51, opened 2022-09-27
workaround more issues with the SLE build environment

dirkmueller:


dcermak:
Does this package build anywhere?

PR #128, opened 2022-11-22
Python devel install

dirkmueller:
context here is https://github.com/SUSE/BCI-tests/pull/184

### DIFF
--- a/src/bci_build/package.py
+++ b/src/bci_build/package.py
@@ -1087,7 +1087,7 @@ def _get_python_kwargs(
     py3_ver: Literal["3.6", "3.8", "3.9", "3.10"], os_version: OsVersion
 ):
     is_system_py: bool = py3_ver == (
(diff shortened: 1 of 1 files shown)


src/bci_build/package.py
```diff
@@ -1106,8 +1106,7 @@ def _get_python_kwargs(
         "version": py3_ver,
         "additional_versions": ["3"],
         "env": {"PYTHON_VERSION": py3_ver_replacement, "PIP_VERSION": pip3_replacement},
-        "package_list": [py3, pip3, "curl", "git-core"]
-        + (
+        "package_list": [f"{py3}-devel", pip3, "curl", "git-core"] + (
```
dcermak, line 1109: I am not convinced that this is really necessary. Only some packages require python3-devel to be build and this makes the container bigger.

dirkmueller: it makes the container bigger, but it also makes the container more stable because mismatches between repo and container (think about people pinning a container) no longer breaks by default

PR #143, opened 2022-11-29
Add main python package to python containers

dcermak:
Otherwise certain modules like sqlite3 are missing.
This fixes: https://bugzilla.suse.com/show_bug.cgi?id=1205850


dirkmueller (approved):
good catch

dirkmueller:
confirmed working

PR #197, opened 2023-01-02
Properly quote the changelog entry in /vc command

dcermak:


.github/workflows/vc-command.yml
```diff
@@ -53,7 +53,7 @@ jobs:
               -vvvv --os-version $OS_VERSION \
```
dirkmueller, line 53: while being at it, how about quoting this as well?

.github/workflows/vc-command.yml
```diff
@@ -53,7 +53,7 @@ jobs:
...
                 add_changelog_entry \
                 --user $user \
```
dirkmueller, line 55: and here?

.github/workflows/vc-command.yml
dcermak: This one will never contain any spaces

.github/workflows/vc-command.yml
dcermak: Usernames should not contain spaces so it shouldn't be necessary :crossed_fingers:

dirkmueller: quoting handles more than just spaces (for example usernames like '-miska-' are valid, but need to be quoted in shell)

PR #201, opened 2023-01-10
Remove 15 SP3 from container generation

dcermak:


dirkmueller (approved):
quite an impressive code cleanup! LGTM, matches what we decided on.

PR #204, opened 2023-01-11
Disable systemd colors in the bci-init image

dirkmueller:
That makes automation on the image slightly easier via expect()

### DIFF
--- a/src/bci_build/package.py
+++ b/src/bci_build/package.py
@@ -6,6 +6,7 @@ import asyncio
 import datetime
 import enum
 import os
+import textwrap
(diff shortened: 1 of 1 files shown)


dcermak (changes requested):
I still see colors in `registry.opensuse.org/home/defolos/bci/staging/sle-15-sp4/sle15-sp4-eoqfo/containerfile/bci/bci-init:latest` by executing `systemctl status` inside that container. Could you double check that please @dirkmueller?

dcermak (approved):

PR #221, opened 2023-02-08
More 🦆-tape around BuildVersion

dcermak:
We have to swap out the version and buildversion for language stack containers where the version is higher than buildversion. Otherwise images without buildversion that are older will order as newer than images that are actually newer, but have the buildversion set.

### DIFF
--- a/src/bci_build/package.py
+++ b/src/bci_build/package.py
@@ -1026,8 +1026,17 @@ class LanguageStackContainer(BaseContainerImage):
             # for non PEP440 versions, we'll get an exception and just return
             # the parent's classes build_version
             try:
-                version.parse(str(self.version))
-                return f"{build_ver}.{self.version}"
+                # more fun!
+                # if the language stack version has a higher version that
+                # build_version, then we must put the stack version first and
+                # append the build version, so that it will order as newer than
+                # previously published images without buildversion
+                stack_version = version.parse(str(self.version))
+                build_version = version.parse(build_ver)
+                if stack_version >= build_version:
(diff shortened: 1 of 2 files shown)


Vogtinator (commented):
IMO this hack is only necessary because of historical releases of certain images, where the version didn't start with the SLE version and was >= 15.x. So hardcoding exceptions here is IMO safer and easier to reason about.

With this approach, e.g. openjdk-14 gets `15.4.14` while openjdk-16 ends up as `16.15.4`...

dcermak:
> IMO this hack is only necessary because of historical releases of certain images, where the version didn't start with the SLE version and was >= 15.x. So hardcoding exceptions here is IMO safer and easier to reason about.
> 
> With this approach, e.g. openjdk-14 gets `15.4.14` while openjdk-16 ends up as `16.15.4`...

Isn't that what we wanted to achieve though? Also, note that this monstrosity only applies if build_version is not None (and it is None for everything but sle 15 sp4)

Vogtinator:
> > IMO this hack is only necessary because of historical releases of certain images, where the version didn't start with the SLE version and was >= 15.x. So hardcoding exceptions here is IMO safer and easier to reason about.
> > With this approach, e.g. openjdk-14 gets `15.4.14` while openjdk-16 ends up as `16.15.4`...
> 
> Isn't that what we wanted to achieve though? Also, note that this monstrosity only applies if build_version is not None (and it is None for everything but sle 15 sp4)

Kind of. I'm not too happy about the order switching between package versions. My brain is currently not fully working so I'm not sure what practical consequences that has.

PR #247, opened 2023-02-24
pcp: fix pmlogger_daily not purging old archives

djoreilly:
The pmlogger_daily.service was failing silently because it runs as user pcp which did not have permissions on `/var/log/pcp/pmlogger`. Normally pcp running on the host creates `/var/log/pcp/pmlogger` and also does `chown pcp:pcp /var/log/pcp/pmlogger` if it does not exist. But `VOLUME /var/log/pcp/pmlogger` in the Dockerfile causes the directory to exist with owner `root:root` in the image.


dcermak:
Superseded by https://github.com/SUSE/BCI-dockerfile-generator/pull/248

PR #254, opened 2023-02-28
Add missing push filter

hennevogel:
We want to trigger the services only on merges, not on every possible event.


dcermak:
The file you are modifying is autogenerated in https://github.com/SUSE/BCI-dockerfile-generator/blob/d19a5b4b940baa58d05e94d1194f8c44772b201c/src/staging/bot.py#L309

PR #259, opened 2023-03-02
Add PHP BCIs

dcermak:


src/bci_build/package.py
```diff
@@ -1423,6 +1424,160 @@ def _get_openjdk_kwargs(
...
+            **extra_env,
+        },
+        extra_files={
+            "/usr/local/bin/docker-php-entrypoint": _php_entrypoint(php_variant),
+            "/usr/local/bin/docker-php-source": _EMPTY_SCRIPT,
+            "/usr/local/bin/docker-php-ext-configure": _EMPTY_SCRIPT,
+            "/usr/local/bin/docker-php-ext-enable": _EMPTY_SCRIPT,
+            "/usr/local/bin/docker-php-ext-install": f"""#!/bin/sh
+set -eox pipefail
+zypper -n in --no-recommends php{php_version}-$1
```
dirkmueller, line 1565: please install with recommends as that adds a couple of package-specified extra extensions

src/bci_build/package.py
```diff
@@ -1423,6 +1424,160 @@ def _get_openjdk_kwargs(
...
+            "APACHE_CONFDIR": "/etc/apache2",
+            "APACHE_ENVVARS": "/usr/sbin/envvars",
+        }
+        cmd = ["apache2-foreground"]
+        custom_end = (
+            common_end
+            + """STOPSIGNAL SIGWINCH
+
+# create our own apache2-foreground from the systemd startup script
+RUN cat /usr/sbin/start_apache2 | sed 's|^exec $apache_bin|exec $apache_bin -DFOREGROUND|' > /usr/local/bin/apache2-foreground
```
dirkmueller, line 1477: I wonder if we can do that by setting /etc/sysconfig/apache2 define. would have a slightly better optics

src/bci_build/package.py
```diff
@@ -1423,6 +1424,160 @@ def _get_openjdk_kwargs(
...
+        + extra_pkgs,
+        replacements_via_service=[
+            Replacement("%%composer_version%%", package_name="php-composer2")
+        ],
+        cmd=cmd,
+        entrypoint=["docker-php-entrypoint"],
+        env={
+            "PHP_VERSION": php_version,
+            "PHP_INI_DIR": f"/etc/php{php_version}/",
+            "PHPIZE_DEPS": f"php{php_version}-devel autoconf gawk gcc make file",
```
dirkmueller, line 1554: do we still need this? do we need to install those packages?

src/bci_build/package.py
```diff
@@ -1423,6 +1424,160 @@ def _get_openjdk_kwargs(
...
+        cmd = ["php-fpm"]
+        custom_end = r"""WORKDIR /srv/www/htdocs
+
+RUN set -eux; \
+	cd /etc/php8/fpm/; \
+        cp php-fpm.d/www.conf.default php-fpm.d/www.conf; \
+        cp php-fpm.conf.default php-fpm.conf; \
+	{ \
+		echo '[global]'; \
+		echo 'error_log = /proc/self/fd/2'; \
```
dirkmueller, line 1500: so this logs to stderr. does that work with apache2 doing that as well?

src/bci_build/package.py
```diff
@@ -1423,6 +1428,163 @@ def _get_openjdk_kwargs(
...
+            "COMPOSER_VERSION": "%%composer_version%%",
+            **extra_env,
+        },
+        extra_files={
+            "/usr/local/bin/docker-php-entrypoint": _php_entrypoint(php_variant),
+            "/usr/local/bin/docker-php-source": _EMPTY_SCRIPT,
+            "/usr/local/bin/docker-php-ext-configure": _EMPTY_SCRIPT,
+            "/usr/local/bin/docker-php-ext-enable": _EMPTY_SCRIPT,
+            "/usr/local/bin/docker-php-ext-install": f"""#!/bin/sh
+set -eox pipefail
```
dirkmueller, line 1571: `set -euo pipefail`

src/bci_build/package.py
dcermak: Not really I think

src/bci_build/package.py
```diff
@@ -982,7 +986,7 @@ async def write_file_to_dest(fname: str, contents: Union[str, bytes]) -> None:
 
         for fname, contents in self.extra_files.items():
             files.append(fname)
-            tasks.append(asyncio.ensure_future(write_file_to_dest(fname, contents)))
+            tasks.append(write_file_to_dest(fname, contents))
```
dirkmueller, line 989: whats this change about? is this intentional?

dirkmueller (dismissed):
LGTM, question inline (on the asyncio.ensure* removal)

dirkmueller (dismissed):

dirkmueller (approved):

PR #267, opened 2023-03-07
Use a valid email address for automated commits & changelog entries

dcermak:


dirkmueller:
@dcermak if you want me to review something, please give me done time to discover the email. I am not reading GitHub notifications 24/7 but only once or twice a day 

I think we have a team list (there bugzilla default assigned list) that we could do here.

src/bci_build/package.py
```diff
@@ -975,7 +975,7 @@ async def write_file_to_dest(fname: str, contents: Union[str, bytes]) -> None:
                     write_file_to_dest(
                         changes_file_name,
                         f"""-------------------------------------------------------------------
-{datetime.datetime.now(tz=datetime.timezone.utc).strftime("%a %b %d %X %Z %Y")} - SUSE Update Bot <noreply@suse.com>
+{datetime.datetime.now(tz=datetime.timezone.utc).strftime("%a %b %d %X %Z %Y")} - SUSE Update Bot <dcermak@suse.com>
```
dirkmueller, line 978: That's just misleading now (you're not an update bot)

How about simply using the team mailing list?

PR #271, opened 2023-03-09
Specify full path for the postgresql entrypoint

dcermak:


dirkmueller (approved):
does it need a chmod +x also?

the other containers have this as custom_end:

```
{DOCKERFILE_RUN} chmod +x /usr/local/bin/entrypoint.sh
```

~~it is missing here. that looks inconsistent.~~ its there.

PR #278, opened 2023-03-10
Add HEALTCHECK to the postgresql containers

dcermak:


src/bci_build/package.py
```diff
@@ -1841,6 +1841,8 @@ def _create_php_bci(
     sed -ri "s|^#?(listen_addresses)\s*=\s*\S+.*|\1 = '*'|" /usr/share/postgresql{ver}/postgresql.conf.sample
 
 STOPSIGNAL SIGINT
+HEALTHCHECK --interval=10s --start-period=10s --timeout=5s \
+    CMD psql postgresql://${{POSTGRES_USER:-postgres}}:$POSTGRES_PASSWORD@localhost:5432
```
dirkmueller, line 1845: why not doing something like

`pg_isready -U ${{POSTGRES_USER:-postgres}} `

that script was made for healthchecking.

dcermak: very simple: I didn't know this was a thing…

dirkmueller (approved):

PR #283, opened 2023-03-13
Add comment by the bot how to launch BCI tests on the staging project

dcermak:


dirkmueller:
Don't really have an opinion on this. Code looks good to me. Only question I have is why not run the bci testsuite for the user as part of the ci for the PR ? 

I think especially if we would stop rebuilding the world that might actually be a nice quality of life improvement without eating too much into our CI minutes budget.

PR #291, opened 2023-03-14
VOLUME goes last

dcermak:


src/bci_build/package.py
```diff
@@ -1838,7 +1838,8 @@ def _create_php_bci(
 {DOCKERFILE_RUN} chmod +x /usr/local/bin/docker-entrypoint.sh; \
     ln -s su /usr/bin/gosu; \
     mkdir /docker-entrypoint-initdb.d; \
-    sed -ri "s|^#?(listen_addresses)\s*=\s*\S+.*|\1 = '*'|" /usr/share/postgresql{ver}/postgresql.conf.sample
+    sed -ri "s|^#?(listen_addresses)\s*=\s*\S+.*|\1 = '*'|" /usr/share/postgresql{ver}/postgresql.conf.sample; \
+    mkdir -p "$PGDATA" && chown -R postgres:postgres "$PGDATA" && chmod 0777 "$PGDATA"
```
dirkmueller, line 1842: chmod 0777? Why?

dcermak: So that the entrypoint still works when run as a non-root user.

dirkmueller: I don't understand this. We're chowning it explicitly to the user just before  and it anyway is a volume that has its own inode for this directory, so why does this make any difference?

There's gotta be a better way. Also, as the docker entry point sets up the permission also (if invoked as root), can we not start it as root?

dirkmueller (changes requested):
Please don't describe what you did but why you did it. It for example isn't clear from the description on WHY volume needs to be last

dirkmueller:
Permission on the data  is the one thing that matters when it comes to databases. Happy to help finding a better solution.

src/bci_build/package.py
dcermak: I have removed the `chmod 0777`, it didn't fix anything in particular, you still can't run the database process as an arbitrary user, but you can run it as `postgres` without the `chmod`. So from my pov it's not needed.

dirkmueller:
@dcermak thank you for updating the commit descriptions, this helps a lot! I'm not 100% sure that the volume behavior is like you describe when a volume is mounted (rather than autocreated) from the outside on invocation. As such I don't think we should rely on "VOLUME has to be the last statement"  to have previous chown or copys persist. 

don't particularly care where we position volume as long as we reviewed that none of the other containers that use volume are affected by this change. 

I think 
https://github.com/SUSE/BCI-dockerfile-generator/pull/296 might be a simpler fix for the issue that you're addressing here (I think). the docker-entrypoint script takes care of setting up the data/ directory, and when it is below the home directory the database starts up just fine here.

src/bci_build/package.py
```diff
@@ -1838,7 +1838,8 @@ def _create_php_bci(
 {DOCKERFILE_RUN} chmod +x /usr/local/bin/docker-entrypoint.sh; \
     ln -s su /usr/bin/gosu; \
     mkdir /docker-entrypoint-initdb.d; \
-    sed -ri "s|^#?(listen_addresses)\s*=\s*\S+.*|\1 = '*'|" /usr/share/postgresql{ver}/postgresql.conf.sample
+    sed -ri "s|^#?(listen_addresses)\s*=\s*\S+.*|\1 = '*'|" /usr/share/postgresql{ver}/postgresql.conf.sample; \
+    mkdir -p "$PGDATA" && chown -R postgres:postgres "$PGDATA"
```
dirkmueller, line 1842: Small nitpick, I would prefer

```
install -d -m 700 - owner postres $PGDATA

```

Here

dirkmueller (approved):
Lgtm. I wonder if we should set the entry point user to postres also as that would make it slightly easier to use but that can be done as a follow-up

src/bci_build/templates.py
```diff
@@ -41,9 +41,9 @@
 {%- if image.entrypoint_user %}USER {{ image.entrypoint_user }}{% endif %}
```
dirkmueller, line 41: this needs to be moved after the "dockerfile_custom_end" as well

src/bci_build/package.py
dirkmueller: actually all of these lines are unnecessary if USER is set to postgres - the volume mounts are working then as expected.

dirkmueller (changes requested):
at least USER needs to be moved down as well. which means the whole "custom_end" is not really  an "end" anymore. maybe needs refactoring

dcermak:
Closing in favor of https://github.com/SUSE/BCI-dockerfile-generator/pull/296

PR #296, opened 2023-03-15
Use ~postgres/data as PG_DATA

dirkmueller:
the SUSE SLE packages are expecting to run postgresql below the postgres user home, and locating the data dir inside makes the docker entrypoint set permissions on the volume on first start.


dirkmueller:
wow this container is broken :/ the "su" hack is not correct.

dcermak:
> the SUSE SLE packages are expecting to run postgresql below the postgres user home, and locating the data dir inside makes the docker entrypoint set permissions on the volume on first start.

Interesting, the previous iteration worked as well :thinking: though.

dcermak (changes requested):
This submission breaks multiple use cases:
- running the container as `--user postgres` is now broken
- running with a different PGDATA location is broken

dirkmueller:
> * running the container as `--user postgres` is now broken

it appears this is an ordering issue. `-v something:/somewhere --user postgres` fails while `--user postgres -v something:/somewhere` is working. the --user setting has influence on how the volume mountpoint directory is created. 

> * running with a different PGDATA location is broken

fixed.

dcermak (changes requested):
The unit tests are now broken

PR #304, opened 2023-03-20
Add Python 3.11 container

dirkmueller:


dcermak (changes requested):
There should be one for Tumbleweed as well.

dcermak (approved):
LGTM, thanks!

PR #312, opened 2023-03-21
Restore still existing PostgreSQL versions for Tumbleweed

dirkmueller:


src/bci_build/package.py
```diff
@@ -1845,6 +1845,7 @@ def _create_php_bci(
     for ver, os_version in list(
         product([15, 14], [OsVersion.SP5, OsVersion.TUMBLEWEED])
     )
+    + [(pg_ver, OsVersion.TUMBLEWEED) for pg_ver in (13, 12, 10)]
```
dcermak, line 1848: Per https://en.wikipedia.org/wiki/PostgreSQL#Release_history version 10 reached EoL already last year.
```suggestion
    + [(pg_ver, OsVersion.TUMBLEWEED) for pg_ver in (13, 12)]
```

PR #323, opened 2023-04-03
This project is formatted with black 23.x

dirkmueller:


dcermak (approved):
This is not really necessary as we already pin black to 23.3 in `poetry.lock`, but it won't hurt either.

PR #369, opened 2023-05-02
NodeJS 14 is EOL as of April 30 2023

dirkmueller:


dcermak (changes requested):
That's not the right place, you have to remove it from line 1350 as well.

dcermak (approved):
LGTM!

PR #370, opened 2023-05-02
Track end of support dates for Node and Rust containers

dirkmueller:
These follow upstream discontinuation lifecycle, so add it early on as a label


Vogtinator:
Maybe add it to the kiwi template for completeness?

dirkmueller:
> Maybe add it to the kiwi template for completeness?

done, however untested

dcermak:
> > Maybe add it to the kiwi template for completeness?
> 
> done, however untested

Tests are now in place.

PR #391, opened 2023-05-05
Add supported_until dates to all remaining containers

dcermak:


src/bci_build/package.py
```diff
@@ -152,6 +152,12 @@ def pretty_print(self) -> str:
 
 ALL_OS_VERSIONS = [OsVersion.SP4, OsVersion.SP5, OsVersion.TUMBLEWEED]
 
+ALL_OS_VERSIONS_AND_EOL: list[tuple[OsVersion, datetime.date | None]] = [
+    (OsVersion.SP4, datetime.date(2023, 12, 31)),
```
dirkmueller, line 156: Do we know that date already?

src/bci_build/package.py
```diff
@@ -1184,6 +1190,7 @@ def _get_python_kwargs(
         **_get_python_kwargs("3.6", os_version),
         package_name="python-3.6-image",
         support_level=SupportLevel.L3,
+        supported_until=datetime.date(2028, 7, 31),
```
dirkmueller, line 1193: This date is not yet clear?

src/bci_build/package.py
```diff
@@ -1202,12 +1209,14 @@ def _get_python_kwargs(
...
     support_level=SupportLevel.L3,
     is_latest=True,
+    supported_until=datetime.date(2023, 12, 31),
     **_get_python_kwargs("3.10", OsVersion.SP4),
 )
 
 PYTHON_3_11_SP5 = LanguageStackContainer(
     package_name="python-3.11-image",
     support_level=SupportLevel.TECHPREVIEW,
+    supported_until=datetime.date(2024, 12, 31),
```
dirkmueller, line 1219: No, this is wrong. Python 3.11 will be the next long term supported python on sle15. Its the sle15 eol date which we do not know yet (and we should maybe use a global sle15_general_support_end variable for all these places)

src/bci_build/package.py
```diff
@@ -1392,6 +1402,11 @@ def _get_openjdk_kwargs(
         "custom_end": f"""{DOCKERFILE_RUN} [ -d $JAVA_HOME ]; [ -d $JAVA_BINDIR ]; [ -f "$JAVA_BINDIR/java" ] && [ -x "$JAVA_BINDIR/java" ]""",
     }
 
+    if java_version == 11:
+        common["supported_until"] = datetime.date(2026, 12, 31)
```
dirkmueller, line 1406: I think its sle15 eol? Where did you find that date?

src/bci_build/package.py
```diff
@@ -1392,6 +1402,11 @@ def _get_openjdk_kwargs(
...
+    elif java_version == 17:
+        common["supported_until"] = datetime.date(2027, 12, 31)
```
dirkmueller, line 1408: Same here

src/bci_build/package.py
```diff
@@ -1898,6 +1921,8 @@ def _create_php_bci(
         package_name="alertmanager-image",
         os_version=os_version,
         is_latest=os_version in CAN_BE_LATEST_OS_VERSION,
+        support_level=SupportLevel.TECHPREVIEW,
+        supported_until=eol,
```
dirkmueller, line 1925: Setting supportend on techpreview doesn't make sense imho?

dcermak:
> Thanks for the diligence here. I left those dates out because they are not clear to me currently (and I wonder if we should simply set them to today+4weeks or something like that)

I took all dates from https://confluence.suse.com/display/ENGCTNRSTORY/SLE+BCI+Image+Overview

dirkmueller:
I like the thoroughness of the PR that is done here, thanks for pulling that all together, @dcermak. I still think we shouldn't do this for dates that are far out in the future and that might change. I think the other two pull requests (link the sle15 specific lifecycle page and only define support.until dates for things that are absolutely certain, aka #393 and maybe a followup on certain parts where we're certain because it is less than a year out) is the only thing we should be doing. I think its too much busy work to keep it correct and has a risk of misleading users).

PR #393, opened 2023-05-12
Switch lifecycle url to the SLE15 specific url

dirkmueller:


dcermak (commented):
We do not set the equivalent of `com.suse.lifecycle-url` for opensuse, so I am a bit hesitant to add another unused property to `ImageProperties`

dirkmueller:
We could set it for opensuse as well in a follow-up PR

dcermak:
> We could set it for opensuse as well in a follow-up PR

I have done that.

vmoutoussamy (commented):
LGTM!

PR #396, opened 2023-05-12
Update NodeJS support end of life dates

dirkmueller:
Actually we do not follow upstream but simply expand or shorten it by SPx support cycle.


dcermak (approved):

dirkmueller:
(rebase after #407 )

PR #406, opened 2023-05-17
Move latest tag to python311 container on SP5

dirkmueller:
We already accidentally released python 311 from SP5, so this one needs to be special ;-/


src/bci_build/package.py
```diff
@@ -1212,7 +1212,8 @@ def _get_python_kwargs(
         support_level=SupportLevel.L3,
         # https://peps.python.org/pep-0664/ defines 2027/10/31, SUSE offers until end of the year
         supported_until=datetime.date(2027, 12, 31),
-        is_latest=os_version in CAN_BE_LATEST_OS_VERSION,
+        # we accidentally released from SP5 already so we should set latest in SP5
+        is_latest=os_version == OSVersion.SP5,
```
dcermak, line 1216:
```suggestion
        is_latest=os_version == OsVersion.SP5,
```

dirkmueller:
decision was to delete the SP5 containers and we'll do with SP4 going forward.

PR #413, opened 2023-05-22
Add a space constraint for rmt-mariadb-image

brunoleon:
Looks like the build does require at least 9.5GB.
This should ensure only proper worker are selected.


dcermak:
@brunoleon Please rebase on main, the CI should fix itself then

dcermak:
https://github.com/SUSE/BCI-dockerfile-generator/pull/419

PR #430, opened 2023-05-31
Set the password_encryption to md5 when host auth is md5

dcermak:
Systems with libpg < 10 cannot talk to the BCIs as they do not support SCRAM for authentication. Setting only `POSTGRES_HOST_AUTH_METHOD` to md5 is insufficient, as recent postgres versions will use SCRAM for password encryption. Thus we have to override the `password_encryption` in that case as well.

See also:
https://stackoverflow.com/questions/62807717/how-can-i-solve-postgresql-scram-authentication-problem/62808487#62808487

### DIFF
--- a/src/bci_build/postgres/entrypoint.sh
+++ b/src/bci_build/postgres/entrypoint.sh
@@ -88,8 +88,10 @@ docker_init_database_dir() {
 		set -- --waldir "$POSTGRES_INITDB_WALDIR" "$@"
 	fi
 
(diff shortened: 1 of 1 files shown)


vmoutoussamy:
@dcermak this LGTM but I can't merge as it's a draft.

dcermak:
Made obsolete by https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/17364

PR #445, opened 2023-06-13
Switch go & rust from versioned to stable & oldstable

dcermak:


dirkmueller:
This also generates the "#!BuildTag: bci/golang:stable-%RELEASE%" in addition, do we really want that?

dirkmueller:
okay, I see it is consisten with how we do it elsewhere. so fine.

dirkmueller (approved):

PR #462, opened 2023-06-20
Set BuildRelease for oldstable images

dirkmueller:
When we switch a version from "stable" to "oldstable", we need to tell OBS the ci counter of the stable image so that the %RELEASE% remains incremental.

### DIFF
--- a/src/bci_build/package.py
+++ b/src/bci_build/package.py
@@ -408,6 +408,9 @@ class BaseContainerImage(abc.ABC):
     #: :command:`zypper` in :file:`Dockerfile`
     no_recommends: bool = True
 
+    #: minimum build release number to use for the %RELEASE% tag expansion
+    min_build_release: Optional[str] = None
+
     _image_properties: ImageProperties = field(default=_SLE_IMAGE_PROPS)
 
     def __post_init__(self) -> None:
@@ -456,6 +459,12 @@ class BaseContainerImage(abc.ABC):
     def build_version(self) -> Optional[str]:
         if self.os_version in (OsVersion.SP4, OsVersion.SP5):
             return f"15.{int(self.os_version.value)}"
+        return None
+
+    @property
+    def build_release(self) -> Optional[str]:
+        if self.os_version in CAN_BE_LATEST_OS_VERSION and self.min_build_release:
+            return self.min_build_release
         return None
 
     @property
@@ -1312,6 +1321,8 @@ def _get_golang_kwargs(ver: _GO_VER_T, os_version: OsVersion):
     is_stable = ver == _GOLANG_VERSIONS[-1]
     stability_tag = "stable" if is_stable else "oldstable"
     go = f"go{ver}"
+    # Set to the latest cicount of -stable- when that version becomes oldstable
+    oldstable_min_build_release = None
     return {
         "os_version": os_version,
         "package_name": f"golang-{stability_tag}-image",
@@ -1319,6 +1330,7 @@ def _get_golang_kwargs(ver: _GO_VER_T, os_version: OsVersion):
         "name": "golang",
         "additional_versions": [stability_tag],
         "pretty_name": f"Golang {ver}",
+        "min_build_release": oldstable_min_build_release if not is_stable else None,
         "is_latest": (is_stable and (os_version in CAN_BE_LATEST_OS_VERSION)),
         "version": ver,
         "env": {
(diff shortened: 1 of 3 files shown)


src/bci_build/templates.py
```diff
@@ -13,6 +13,7 @@
 #!BuildTag: {{ tag }}
 {% endfor -%}
 {% if image.build_version %}#!BuildVersion: {{ image.build_version }}{% endif %}
+{% if image.build_release %}#!BuildRelease: {{ image.build_release }}{% endif %}
```
dcermak, line 16: Can you please modify the template so that we don't get a mass change in the Dockerfiles?

dirkmueller: Done I think.

src/bci_build/package.py
```diff
@@ -458,6 +461,12 @@ def build_version(self) -> Optional[str]:
             return f"15.{int(self.os_version.value)}"
         return None
 
+    @property
+    def build_release(self) -> Optional[str]:
+        if self.os_version in CAN_BE_LATEST_OS_VERSION and self.min_build_release:
```
dcermak, line 466: Why this condition?

dcermak (commented):
I guess we need this, but I'm not overly happy with the amount of documentation around this new feature.

Additionally the conditions when it gets set feel kinda arbitrary: for `BuildVersion` we set this for the whole Service Pack, but for `BuildRelease` we set it if `min_build_release` is set and the SP can have a `latest` tag? Why not remove this hidden complexity and just have `min_build_release` as the only property and set it in the container definition?

src/bci_build/package.py
dirkmueller: we only want to set the %RELEASE% number on the newest SP, not on the older ones (otherwise they would have the same release number)

dirkmueller:
> Additionally the conditions when it gets set feel kinda arbitrary: for `BuildVersion` we set this for the whole Service Pack, but for `BuildRelease` we set it if `min_build_release` is set and the SP can have a `latest` tag? Why not remove this hidden complexity and just have `min_build_release` as the only property and set it in the container definition?

they do two different things as far as I understood. BuildVersion determines which build result can be aggregated by giving them an order. BuildRelese is modifying the %RELEASE% expansion which is the suffix of the release number. 

We don't need BuildVersion if we remove the aggregate from the publishing logic (which would make sort of more sense anyway .. to me at least)

Vogtinator:
> they do two different things as far as I understood.

Not in that sense. It's exactly like `Version:` and `Release:` in RPMs and used for version comparisons in the same way. When version changes, the OBS internal release counter is reset to 1.1.

> BuildVersion determines which build result can be aggregated by giving them an order. BuildRelease is modifying the %RELEASE% expansion which is the suffix of the release number. but Build Release is not influencing the aggregation order.

It's actually not the order of aggregation. The aggregate just copies all binaries. What finally looks at the version-release is the code that collects all the tags to push to the registry: It picks the provider of a tag with the highest version-release.

So far we made sure that the version is enough to determine which provider wins. If it's the same, OBS looks at the release.

dirkmueller:
> > they do two different things as far as I understood.
> Not in that sense. It's exactly like `Version:` and `Release:` in RPMs and used for version comparisons in the same way. When version changes, the OBS internal release counter is reset to 1.1.

Sure, but on top of that we also make the Release field use-visible by exporting it into an immutable tag (e.g. BuildTag: X.Y-%RELEASE% in the Dockerfile). and that %RELEASE% should remain continuous and incremental from a user perspective as well. 

> It's actually not the order of aggregation. The aggregate just copies all binaries. What finally looks at the version-release is the > code that collects all the tags to push to the registry: It picks the provider of a tag with the highest version-release.

Yes, but for that BuildVersion is sufficient. that's what I meant earlier with the ordering. 

Now we could also be incrementing the BuildVersion on every code change, but that would be a lot of more extra logic on top of that.

Vogtinator:
> > > they do two different things as far as I understood.
> > > Not in that sense. It's exactly like `Version:` and `Release:` in RPMs and used for version comparisons in the same way. When version changes, the OBS internal release counter is reset to 1.1.
> 
> Sure, but on top of that we also make the Release field use-visible by exporting it into an immutable tag (e.g. BuildTag: X.Y-%RELEASE% in the Dockerfile). and that %RELEASE% should remain continuous and incremental from a user perspective as well.

Yep! Which brings us to the root cause: moving tags between source containers (!) is against the design of OBS.

> > It's actually not the order of aggregation. The aggregate just copies all binaries. What finally looks at the version-release is the > code that collects all the tags to push to the registry: It picks the provider of a tag with the highest version-release.
> 
> Yes, but for that BuildVersion is sufficient. that's what I meant earlier with the ordering.
> 
> Now we could also be incrementing the BuildVersion on every code change, but that would be a lot of more extra logic on top of that.

That's basically the intention of having `#!BuildVersion` in the first place. In fact, the kiwi documentation actually describes it as such (IIRC).

Vogtinator:
> > Yep! Which brings us to the root cause: moving tags between source containers (!) is against the design of OBS.
> 
> Where is this documented? I mean, with the same reasoning you could say that having the same tag created by same package in two different projects is against the design.

Almost, but the actual difference there is that normally (container and RPM) repo publish locations are always prefixed with the project + repo name, but that's not the case for the container publish projects.

For RPMs it's also broken to move a binary package from one source to another while keeping the version identical. The new package will have a lower release by default, resulting in the same problems. With the (old) infrastructure, the binary names will also collide with the past ones from the old package, the same issue you'll have here with the container tags.

> The thing is, we so far relied on the magic release number generation (which is simply checkin counter on a package container) to be always correct. and in many cases it isn't, like for example when we switch from one project to another project. that's why maintenance for example added the sleversion as a prefix. we could also be adding a prefix, which would be an alternative to this approach. We could just maintain the continuity in sources, but we don't maintain sources, we generate them...
> 
> > That's basically the intention of having `#!BuildVersion` in the first place. In fact, the kiwi documentation actually describes it as such (IIRC).
> 
> Yes, but again, we're generating %RELEASE% into sources. if we stop doing that then we can stop doing this.

We could do that, by moving the responsibility of rebuilds from OBS to something which also takes over the release counting. Question is whether this turns out better than the status quo in most aspects.

dirkmueller:
I'll take a look at release counter generation

dirkmueller:
While investigating that, we noticed that the build service release counter generation is entirely broken for BCI because it was never able to parse the Dockerfiles we generate. 

so meanwhile I came up with an alternative implementation in https://github.com/SUSE/BCI-dockerfile-generator/pull/473 which hopefully gets more positive reviews (and is not against the OBS design principles)

PR #473, opened 2023-06-22
Prepend the stability_tag in the tag list

dirkmueller:
The first tag that is set is used by the build service to define the binary resulting name. by using the stability tag we have a stable name which avoids constant fixing of aggregates.

### DIFF
--- a/src/bci_build/package.py
+++ b/src/bci_build/package.py
@@ -1055,14 +1055,50 @@ class LanguageStackContainer(BaseContainerImage):
     @property
     def build_tags(self) -> List[str]:
         tags = []
+        # The stability-tags feature in containers. may result in the generation of
+        # identical release numbers for the same version from two different package
+        # containers, such as "oldstable" and "stable."
+        #
+        # When a new version is committed, the check-in counter and the rebuild
+        # counters are reset to "1", resulting in a release number like 1.1.
+        # Subsequent changes will have release numbers like 1.2 (rebuild on the same source)
+        # and 2.1 (new source change without version change).
+        #
+        # Here's an example:
+        #   lang-stable: 1.70-1.1 (first build of the very first commit after version update)
+        #   lang-oldstable: 1.69-5.1 (first build of the fifth source change after version update)
+
+        # After a version rollover occurs from "stable" to "oldstable", the release numbers become:
+
+        #   lang-oldstable: 1.70-1.1
+        #   lang-stable: 1.71-1.1
+
+        # Now there is a conflict with the tags that were previously produced by the lang-stable
+        # container (see lang-stable-1.70-1.1 above).
+        #
+        # In order To resolve this, the solution is to namespace
+        # the tags with a prefix based on the stability ordering. This ensures that we have:
+
+        #   lang-stable: 1.70-1.1.1
+        #   lang-oldstable: 1.69-2.5.1
+
+        # After a version rollover, the numbers now become:
+
+        #   lang-oldstable: 1.70-2.1.1
+        #   lang-stable: 1.71-1.1.1
+
+        # To avoid conflicts, the tags are deconflicted based on the stability ordering.
+        _stability_tag_ordering = (None, "stable", "oldstable")
+        relsuffix = "%RELEASE%"
+        if self.stability_tag and self.stability_tag in _stability_tag_ordering:
+            relsuffix = f"{_stability_tag_ordering.index(self.stability_tag)}.%RELEASE%"
         for name in [self.name] + self.additional_names:
             ver_labels = [self.version_label]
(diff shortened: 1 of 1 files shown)


src/bci_build/package.py
```diff
@@ -1052,10 +1055,18 @@ def uid(self) -> str:
...
     def build_tags(self) -> List[str]:
         tags = []
+        # deconflict tags by the stability ordering
+        _stability_tag_ordering = (None, "stable", "oldstable")
+        relsuffix = "%RELEASE%"
+        if self.stability_tag and self.stability_tag in _stability_tag_ordering:
+            relsuffix = f"{_stability_tag_ordering.index(self.stability_tag)}.%RELEASE%"
         for name in [self.name] + self.additional_names:
-            for ver_label in [self.version_label] + self.additional_versions:
+            ver_labels = [self.version_label]
```
dcermak, line 1064: Wouldn't this be a simpler solution?
```suggestion
            for ver_label in self.additional_versions + [self.version_label]:
```

dcermak:
Atm the tags are added as follows to the `Dockerfile`:
```
#!BuildTag: bci/rust:stable
#!BuildTag: bci/rust:stable-1.%RELEASE%
```
Is that intentional?

src/bci_build/package.py
```diff
@@ -1052,10 +1055,18 @@ def uid(self) -> str:
...
+            if self.stability_tag:
+                ver_labels = [self.stability_tag] + ver_labels
+            for ver_label in ver_labels + self.additional_versions:
                 tags += [f"{self._registry_prefix}/{name}:{ver_label}"] + [
-                    f"{self._registry_prefix}/{name}:{ver_label}-%RELEASE%"
+                    f"{self._registry_prefix}/{name}:{ver_label}-{relsuffix}"
```
dcermak, line 1069: We don't want that necessarily, because now we have tags like this:
```
#!BuildTag: opensuse/bci/golang:stable
#!BuildTag: opensuse/bci/golang:stable-1.%RELEASE%
#!BuildTag: opensuse/bci/golang:1.20
#!BuildTag: opensuse/bci/golang:1.20-1.%RELEASE%
```
The last entry is not correct though (well, technically it is not wrong, but imho it's missleading as it pretends to be a different build than it actually is)

dirkmueller:
> #!BuildTag: bci/rust:stable-1.%RELEASE%
> ```
> Is that intentional?

yes, thats what makes the release numbers non-overlapping no matter how broken the OBS relnumber generation is.

src/bci_build/package.py
dirkmueller: The last entry is correct ( we need to deconflict the release numbers). just setting it on stable is not enough

dcermak:
> > ```
> > #!BuildTag: bci/rust:stable-1.%RELEASE%
> > ```
> > Is that intentional?
> > 
> 
> yes, thats what makes the release numbers non-overlapping no matter how broken the OBS relnumber generation is.

No, I meant whether the quoted entry should be second and the first one is the plain `bci/rust:stable`

src/bci_build/package.py
dcermak: Ah, you're right. That makes sense.

src/bci_build/package.py
```diff
@@ -1052,10 +1055,18 @@ def uid(self) -> str:
     @property
     def build_tags(self) -> List[str]:
         tags = []
+        # deconflict tags by the stability ordering
+        _stability_tag_ordering = (None, "stable", "oldstable")
+        relsuffix = "%RELEASE%"
+        if self.stability_tag and self.stability_tag in _stability_tag_ordering:
+            relsuffix = f"{_stability_tag_ordering.index(self.stability_tag)}.%RELEASE%"
```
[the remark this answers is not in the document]
dirkmueller: There's documentation in the commit message, but I expanded now the commit message also into the source to have it twice and more obvious.

dirkmueller:
> No, I meant whether the quoted entry should be second and the first one is the plain `bci/rust:stable`

I *think* that doesn't matter, but I can change the order, it will just cause a mass change however...

src/bci_build/package.py
```diff
@@ -1052,11 +1055,50 @@ def uid(self) -> str:
     @property
     def build_tags(self) -> List[str]:
         tags = []
+        # The stability-tags feature in containers. may result in the generation of
```
dcermak, line 1058:
```suggestion
        # The stability-tags feature in containers may result in the generation of
```

dcermak (approved):
LGTM!

Feel free to ignore the typo and just merge this or amend & ignore the scratch build.

PR #494, opened 2023-07-05
Also build PHP8 containers for Tumbleweed

dirkmueller:
The test suite wants to test them


dcermak (changes requested):
FTBFS.

The replace using package version service fails to find the php package.

dirkmueller:
> FTBFS.

Fixed (I hope)

> The replace using package version service fails to find the php package.


No, thats just https://github.com/openSUSE/obs-service-replace_using_package_version/pull/56

dcermak:
Thanks, looks good!

PR #508, opened 2023-07-06
Build the base containers with the docker-flavor

dirkmueller:
Otherwise the ftp flavor is used, which is incorrect

### DIFF
--- a/src/bci_build/package.py
+++ b/src/bci_build/package.py
@@ -1342,7 +1342,6 @@ def _get_ruby_kwargs(ruby_version: Literal["2.5", "3.2"], os_version: OsVersion)
             "util-linux",
             "curl",
             "git-core",
-            "distribution-release",
             # additional dependencies to build rails, ffi, sqlite3 gems -->
             "gcc-c++",
             "sqlite3-devel",
@@ -1403,7 +1402,7 @@ def _get_golang_kwargs(
                 regex_in_build_description=golang_version_regex, package_name=go
             )
         ],
-        "package_list": [go, "distribution-release", "make", "git-core"],
+        "package_list": [go, "make", "git-core"],
         "extra_files": {
             # the go binaries are huge and will ftbfs on workers with a root partition with 4GB
             "_constraints": generate_disk_size_constraints(8)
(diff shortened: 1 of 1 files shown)


src/bci_build/package.py
```diff
@@ -2234,6 +2234,9 @@ def _create_php_bci(
         package_list=[
             Package(name, pkg_type=PackageType.BOOTSTRAP)
             for name in (
+                "openSUSE-release-appliance-docker"
+                if os_version == OsVersion.TUMBLEWEED
+                else "sles-release",
```
dcermak, line 2239: I would suggest to split this out into a separate helper function, as this is otherwise exactly duplicated in 3 places.

dirkmueller: good idea. actually we can generalize that outside this code path

src/bci_build/package.py
```diff
@@ -770,6 +770,11 @@ def pkg_filter_func(p: Union[str, Package]) -> bool:
             for pkg_type in PKG_TYPES
         )
 
+        # Ensure that we build base OS containers with the proper product flavor
+        # for correct telemetry reporting
+        if self.is_opensuse and isinstance(self, OsContainer):
+            bootstrap_packages.append("openSUSE-release-appliance-docker")
+
```
dcermak, line 777: No, please no additional hidden magic. I fear that this will be impossible to find down the road why packages are suddenly getting added in some containers and not in others.

dirkmueller: well, in this case it is easy to grep (just search for the package name)

dcermak (approved):
LGTM, but needs a rebase.

dcermak (commented):
Dirk, I think this is a bit of a duplication that we are introducing here: in the SLES BCIs we have a prjconf that picks `sles-release` for `distribution-release`.

WDYT about removing `distribution-release` from the package list entirely instead?

src/bci_build/package.py
```diff
@@ -2302,7 +2313,8 @@ def _get_minimal_kwargs(os_version: OsVersion):
         cmd=["/bin/sh"],
         package_list=[
             Package(name, pkg_type=PackageType.BOOTSTRAP)
-            for name in (
+            for name in _get_os_container_package_names(os_version)
+            + (
                 "busybox",
                 "busybox-links",
                 "distribution-release",
```
dcermak, line 2320:
```suggestion
```

PR #510, opened 2023-07-07
Add php8-readline extension on Tumbleweed

dirkmueller:
It is no longer installed by default and required by php interactive mode (the default entrypoint), otherwise it just logs an assert and exits.

### DIFF
--- a/src/bci_build/package.py
+++ b/src/bci_build/package.py
@@ -1656,7 +1656,9 @@ EXPOSE 9000
 """
         )
     else:
-        extra_pkgs = []
+        extra_pkgs = (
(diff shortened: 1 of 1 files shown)


src/bci_build/package.py
```diff
@@ -1656,7 +1656,9 @@ def _create_php_bci(
 """
         )
     else:
-        extra_pkgs = []
+        extra_pkgs = (
+            [] if OS_VERSION != "tumbleweed" else [f"php{php_version}-readline"]
```
dcermak, line 1660:
```suggestion
            [] if os_version != OsVersion.TUMBLEWEED else [f"php{php_version}-readline"]
```

PR #526, opened 2023-07-12
Add util-linux to the java container (for /usr/bin/getopt)

dirkmueller:
The testsuite depends on being able to install cassandra, which calls getopt


dcermak (commented):
I wouldn't be opposed to fixing the test by installing util-linux before installing Casandra. No need to bloat the image for one test.

dirkmueller:
https://github.com/SUSE/BCI-tests/pull/271

PR #529, opened 2023-07-17
Make the rmt-nginx container also available under "nginx" name

dirkmueller:
This is the cheapest way to get a generic nginx container defined


dcermak (approved):
The labels don't match then really but :shrug:

dcermak:
@dirkmueller please rebase

PR #542, opened 2023-07-20
Create the proper custom descriptions for tumbleweed

dirkmueller:
Test suite is complaining about SLE in the name, so lets adopt it properly

### DIFF
--- a/src/bci_build/package.py
+++ b/src/bci_build/package.py
@@ -362,9 +362,10 @@ class BaseContainerImage(abc.ABC):
 
     #: This string is appended to the automatically generated dockerfile and can
     #: contain arbitrary instructions valid for a :file:`Dockerfile`.
-    #: **Caution** Setting both this property and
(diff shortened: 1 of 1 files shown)


src/bci_build/package.py
```diff
@@ -854,13 +854,17 @@ def description(self) -> str:
         :py:attr:`BaseContainerImage.pretty_name` to generate a description.
 
         """
+
+        description_formatters = {
```
dcermak, line 858: Please mention this in the docstring of the `description`. Also, this will certainly mess with `f""` strings, so I'm not overly happy with this change :-/

dirkmueller: we currently do not have f strings with custom description and I don't predict we'll likely get one soon. so I think its okay (and we can always expand those substitutions, or nest them. that also works).

fcrozat (approved):

PR #548, opened 2023-07-21
split nginx into rmt and plain nginx image

dirkmueller:
This allows us to set the description and pretty names properly for the better registry.suse.com appearance.

Also slightly clean it up while being at it.

### DIFF
--- a/src/bci_build/package.py
+++ b/src/bci_build/package.py
@@ -2118,48 +2118,56 @@ for filename in (
         _NGINX_FILES[filename] = cursor.read(-1)
 
 
-NGINX_CONTAINERS = [
-    ApplicationStackContainer(
-        package_name="rmt-nginx-image",
-        os_version=os_version,
-        is_latest=os_version in CAN_BE_LATEST_OS_VERSION,
-        name="rmt-nginx",
-        additional_names=["nginx"],
-        pretty_name="NGINX for SUSE RMT",
-        version="%%nginx_version%%",
-        version_in_uid=False,
-        replacements_via_service=[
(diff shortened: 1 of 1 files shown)


src/bci_build/package.py
```diff
@@ -2118,48 +2118,56 @@ def _get_os_container_package_names(os_version: OsVersion):
...
-COPY docker-entrypoint.sh /
-{DOCKERFILE_RUN} chmod +x /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
-{DOCKERFILE_RUN} chmod +x /docker-entrypoint.d/20-envsubst-on-templates.sh
-{DOCKERFILE_RUN} chmod +x /docker-entrypoint.d/30-tune-worker-processes.sh
-{DOCKERFILE_RUN} chmod +x /docker-entrypoint.sh
-
+        "package_list": ["gawk", "nginx"],
+        "entrypoint": ["/usr/local/bin/docker-entrypoint.sh"],
+        "cmd": ["nginx", "-g", "daemon off;"],
+        "build_recipe_type": BuildType.DOCKER,
```
dcermak, line 2137: This is actually no longer required
```suggestion
```

dcermak (commented):
This will cause us to have two completely equal containers which only differ in their description… That seems honestly extremely wasteful. Why can't the RMT Helm chart use the general purpose NGINX container?
I will not insist on this, but tbh this doesn't feel correct.

dirkmueller:
@dcermak that's exactly the plan, but we first need to introduce the new container so that the chart can get modified to the new image

PR #566, opened 2023-08-01
Add openqa-single-instance container

dcermak:


okurz (commented):
The change looks ok but is heavily duplicating the upstream container Dockerfile so sooner or later this will also fail on changes. With that I suggest to not merge that until we can come up with a way with less duplications

dcermak:
> The change looks ok but is heavily duplicating the upstream container Dockerfile so sooner or later this will also fail on changes. With that I suggest to not merge that until we can come up with a way with less duplications

@okurz I think you misunderstood my suggestion of this, my idea was to remove the need for you to maintain the `Dockerfile` and instead use the BCI tooling. Hence this is not duplication, but a replacement.

okurz:
Well, that makes more sense but isn't it a good idea to have a Dockerfile in the upstream repo usable by anyone?

dcermak:
> Well, that makes more sense but isn't it a good idea to have a Dockerfile in the upstream repo usable by anyone?

Well that really depends on your expectations as a project. Given that your Dockerfile only works on OBS anyway, I wouldn't see that much of a need. But that is your call to make.

src/bci_build/package.py
```diff
@@ -2530,6 +2530,38 @@ def _get_minimal_kwargs(os_version: OsVersion):
     for os_version in ALL_NONBASE_OS_VERSIONS
 ]
 
+OPENQA_SINGLE_INSTANCE_CONTAINER = ApplicationStackContainer(
+    name="openqa-single-instance",
+    package_name="openqa-single-instance-image",
+    os_version=OsVersion.TUMBLEWEED,
+    pretty_name="openQA single-instance container",
+    custom_description="A complete openQA instance composed of all necessary components to execute openQA tests including an openQA worker",
```
dirkmueller, line 2538: this is missing at least `{based_on_container}` at the end

src/bci_build/package.py
```diff
@@ -2530,6 +2530,38 @@ def _get_minimal_kwargs(os_version: OsVersion):
...
+        "openQA-bootstrap",
+        "qemu-arm",
+        "qemu-ppc",
+        "qemu-x86",
+        "qemu-tools",
+        "sudo",
+        "iputils",
+        "os-autoinst-distri-opensuse-deps",
+    ],
+    custom_end="HEALTHCHECK CMD curl -f http://localhost || exit 1",
```
dirkmueller, line 2562: theoretically the `|| exit 1` seems redundant here. and maybe use curl -sf ?

dirkmueller (changes requested):
needs rebase. do we still go forward with this approach at all?

dcermak:
Unless @okurz would like this container, I'd just close this. No need to provide a container competing with openQA upstream.

okurz:
Agreed. So far we are good with the openQA upstream container definition

PR #571, opened 2023-08-07
Add subpackages of go as well to golang container

dirkmueller:
This prevents mismatches between BCI repo and the container to cause real issues, plus it provides the convenience of having the Golang race detector available out of the box.


ggardet:
This breaks on armv6/7 since `go-race` does not exist there.
So, we should disable build on armv6/7 (or remove go-race for armv6/7).

PR #574, opened 2023-08-08
Remove unnecessary compat package

dirkmueller:
This is only needed if rpm doesn't detect the usrmerge happened, which it doesn't in the kiwi build process. to be investigated


dcermak:
Uh, why is this package in there in the first place? We only add bootstrap packages to the minimal image, hence not even dependencies should end up there

dirkmueller:
@dcermak see commit message. It is pulled in by the filesystem package if rpm(1) doesn't detect that usrmerge happened. And on an empty system it cannot detect that.

There might be a way to fix that in kiwi/zypper, but this is easy enough so that I don't bother about it too much

PR #581, opened 2023-08-10
Add Golang 1.21 as stable

dirkmueller:
There is no golang 1.21 for openssl yet, so we need to separate the two variants stability tags.


dcermak:
You accidentally created `golang-oldstable-openssl-image` for go 1.20 and 1.19. The bot deletes first all files in each package before commiting the new ones, hence the removal appeared to fail because a concurrent removal already took place :grimacing:.

dcermak (approved):

PR #587, opened 2023-08-23
Create a container for kernel module building

dcermak:
For https://jira.suse.com/browse/PED-4822

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -1354,6 +1354,7 @@ from .openjdk import OPENJDK_CONTAINERS
 from .php import PHP_CONTAINERS
 from .rust import RUST_CONTAINERS
 
+from .basecontainers import KERNEL_MODULE_CONTAINERS
 from .basecontainers import MICRO_CONTAINERS
 from .basecontainers import MINIMAL_CONTAINERS
 from .basecontainers import BUSYBOX_CONTAINERS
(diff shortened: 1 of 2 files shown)


dirkmueller (commented):
There's more needed I'd think. Either we need an extra repository added or at least package the ga versions of the kernel packages?

src/bci_build/package.py
```diff
@@ -2564,6 +2564,18 @@ def _get_minimal_kwargs(os_version: OsVersion):
     for os_version in ALL_NONBASE_OS_VERSIONS
 ]
 
+KERNEL_MODULE_CONTAINERS = [
+    OsContainer(
+        name="kernel-module-toolkit",
```
dirkmueller, line 2569: Can we name it a bit more specific? I think something like sle15 should be in the name to make it clear that its not for ANY kernel

src/bci_build/package.py
```diff
@@ -2564,6 +2564,18 @@ def _get_minimal_kwargs(os_version: OsVersion):
...
+        pretty_name="Kernel Module Toolkit",
+        package_name="kernel-module-toolkit",
+        os_version=os_version,
+        is_latest=os_version in CAN_BE_LATEST_OS_VERSION,
+        package_list=["kernel-devel", "kernel-syms", "gcc"],
```
dirkmueller, line 2574: %kernel_module_package_buildreq is expanding to `kmod-compat kernel-syms kernel-syms-azure` in SLE15. so I think we're missing

dcermak: Do we really want to include the kernel symbols for the Azure kernel? That appears a bit out of place here.

dirkmueller (approved):

PR #591, opened 2023-08-24
Enable SLE 15 SP6

dcermak:


dcermak:
staging is green, the build just took ages

PR #603, opened 2023-09-05
Add release suffix also to build version

dirkmueller:
This is needed for ordering when the same version goes from stable->oldstable


msmeissn:
change looks good to me

PR #607, opened 2023-09-06
Only install go-race on the 64bit architectures

dirkmueller:


Vogtinator:
I hate this cleanup code which pollutes everything and I don't really like that this uses `if zypper install ...`, but doing either of those in a better way is more complex. Sigh.

Vogtinator (approved):

PR #631, opened 2023-09-12
Increase build version for postgresql 14 and nodejs 14

dirkmueller:
we had a buildversion downgrade due to the move from SP5 -> SP4, adjust for that.


msmeissn:
change looks good to me

PR #641, opened 2023-09-19
We don't need a buildversion bump for postgresql after all

dirkmueller:
We never actually released the SP5 based postgresql image, so we don't need it here.


msmeissn:
looks good to me!

PR #658, opened 2023-09-22
Add pipx to the tumbleweed containers

dirkmueller:
Also this improves the setup so that both pip and pipx can install user applications that are in the path afterwards

### DIFF
--- a/src/bci_build/package.py
+++ b/src/bci_build/package.py
@@ -1443,13 +1443,18 @@ def _get_python_kwargs(
         "pretty_name": f"Python {py3_ver} development",
         "version": py3_ver,
         "additional_versions": ["3"],
-        "env": {"PYTHON_VERSION": py3_ver_replacement, "PIP_VERSION": pip3_replacement},
+        "env": {
+            "PYTHON_VERSION": py3_ver_replacement,
+            "PATH": "$PATH:/root/.local/bin",
+            "PIP_VERSION": pip3_replacement,
+        },
         "package_list": [f"{py3}-devel", py3, pip3, "curl", "git-core"]
         + (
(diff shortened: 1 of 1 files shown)


src/bci_build/package.py
```diff
@@ -1463,7 +1468,9 @@ def _get_python_kwargs(
         symlink_py_and_pydoc = rf"""ln -s /usr/bin/python{py3_ver} /usr/local/bin/python3; \
     ln -s /usr/bin/pydoc{py3_ver} /usr/local/bin/pydoc"""
 
-        kwargs["config_sh_script"] = symlink_py_and_pydoc
+        kwargs["config_sh_script"] = (
+            "install -d -m 0755 /root/.local/bin; "
```
dcermak, line 1472: Won't pipx do this for you?

dcermak (commented):
Is it intentional that are adding `/root/.local/bin` to `PATH` also on non-TW?

src/bci_build/package.py
dirkmueller: pipx does create the directory for you once you install something that requires it to do so. however it doesn't add it to the $PATH, which is the real change here. and I don't want to add things to the $PATH that haven't been created in the container with proper permissions, I find that unclean.

dirkmueller:
> Is it intentional that are adding `/root/.local/bin` to `PATH` also on non-TW?

yes, because of jsc#PED-5573 . also the path is needed for `pip --user` as well (libraries can have binaries/console scripts attached to it as well)

dcermak:
@dirkmueller The change is not entirely correct, it sets `ENV` for all python versions but only creates the directory on python != system python.

PR #659, opened 2023-09-25
Ensure that the pcp buildname is `suse-pcp-%%pcp-major%%`

dcermak:


dirkmueller:
The issue here is that the name for the `_aggregate` changed?

I'm afraid thie changed the version label, which I think is unwanted. We need some other solution.

Maybe we need to set the buildname as a writeable property rather then guessing it from the first tag?

dirkmueller:
(I'd appreciate some rationale in the commit message, took me a while to guess what we're doing here)

dirkmueller (changes requested):
see inline comment

dmach (approved):
I don't know much about versioning of this package,
but using `pcp_major` as `version` makes sense to me.

dirkmueller:
> I don't know much about versioning of this package, but using `pcp_major` as `version` makes sense to me.

it changes the version number from "5.2.3" to "5". the LABEL versions are supposed to be the full version number, not just the major version. 

the confusing part here might be that %part_patch is not "3" in the above example but "5.2.3" aka all components of the version number until including the patch version.

dcermak:
This was based on a mistaken understanding of what `%pcp_patch%` is

PR #668, opened 2023-09-25
Add the lifecycle information inside the container

dirkmueller:
this allows testing with the zypper lifecycle plugin and we can write a bci-test that verifies the information to be up to date and matching the label


dcermak (changes requested):
`lifecycle-data-sle-module-development-tools` is not available on Tumbleweed

dcermak (approved):

PR #669, opened 2023-09-27
Add a sensible infoheader with pointers to the SLE BCI project

dirkmueller:


src/bci_build/templates.py
```diff
@@ -2,13 +2,32 @@
 descriptions.
 
 """
+import datetime
 import jinja2
 
 
+INFOHEADER_TEMPLATE = f"""
+    Copyright (c) {datetime.datetime.now().year} SUSE LLC
```
dcermak, line 10: Isn't it common practice to have a starting year as well?

dcermak (commented):
I am not opposed to this change, but I'd like to understand what prompted you to include this header? Is it so commonly read?

src/bci_build/templates.py
dirkmueller: the opensuse spec file formatter doesn't do it, and this basically mimics that one. I don't think the date is needed for anything, we can also omit it if thats better.

dirkmueller:
> I am not opposed to this change, but I'd like to understand what prompted you to include this header? Is it so commonly read?

it adds helpful pointers on how to interact with the BCI team (github discussions, bugzilla) and informs that there shouldn't be manual modifications to it (like recently happened with the openjdk 21 switch)

dcermak (approved):

dirkmueller:
automerge is not going to work due to the unavailability of the s390x workers

PR #670, opened 2023-09-27
Modularize package.py

dirkmueller:
This has multiple thousand lines of code. we can split each container in its own package. this starts with the first two.


src/bci_build/package/basalt_base.py
```diff
@@ -0,0 +1,101 @@
+"""Base Container for the Basalt Project"""
+from __future__ import annotations
```
dcermak, line 2: This shouldn't be necessary
```suggestion
```

src/bci_build/package/python.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env python3
+from __future__ import annotations
+
```
dcermak, line 3:
```suggestion
```

dcermak (dismissed):
LGTM

src/bci_build/package/basalt_base.py
dirkmueller: included in the rebase, thanks!

src/bci_build/package/python.py
dirkmueller: included in the rebase.

PR #701, opened 2023-10-11
Enable mariadb for Basalt & rename it to mariadb

dcermak:
But keep rmt-mariadb on SLE for the helm chart compatibility

For [jsc#PED-2742](https://jira.suse.com/browse/PED-2742)

### DIFF
--- a/src/bci_build/package/appcontainers.py
+++ b/src/bci_build/package/appcontainers.py
@@ -134,33 +134,45 @@ with open(
 ) as entrypoint:
     _MARIAD_ENTRYPOINT = entrypoint.read(-1)
 
-MARIADB_CONTAINERS = [
-    ApplicationStackContainer(
-        package_name="rmt-mariadb-image",
-        os_version=os_version,
-        is_latest=os_version in CAN_BE_LATEST_OS_VERSION,
-        name="rmt-mariadb",
-        version="%%mariadb_version%%",
-        version_in_uid=False,
-        replacements_via_service=[
-            Replacement(
-                regex_in_build_description="%%mariadb_version%%",
-                package_name="mariadb",
-                parse_version="minor",
-            )
-        ],
-        pretty_name="MariaDB Server for SUSE RMT",
-        package_list=["mariadb", "mariadb-tools", "gawk", "timezone", "util-linux"],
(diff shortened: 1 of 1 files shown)


src/bci_build/package/appcontainers.py
```diff
@@ -134,33 +134,45 @@
...
+            ],
+            pretty_name="MariaDB Server",
+            package_list=["mariadb", "mariadb-tools", "gawk", "timezone", "util-linux"],
+            entrypoint=["docker-entrypoint.sh"],
+            extra_files={
+                "docker-entrypoint.sh": _MARIAD_ENTRYPOINT,
+                "_constraints": generate_disk_size_constraints(11),
+            },
+            build_recipe_type=BuildType.DOCKER,
+            cmd=["mariadbd"],
```
dirkmueller, line 172: this was `mariadb` before, why did we change to `mariadbd` ? is that the same?

dirkmueller (approved):

src/bci_build/package/appcontainers.py
```diff
@@ -134,33 +134,45 @@
...
-        build_recipe_type=BuildType.DOCKER,
-        cmd=["mariadbd"],
-        volumes=["/var/lib/mysql"],
-        exposes_tcp=[3306],
-        custom_end=rf"""{DOCKERFILE_RUN} mkdir /docker-entrypoint-initdb.d
+MARIADB_CONTAINERS = []
+MARIADB_CLIENT_CONTAINERS = []
+
+for os_version in set(ALL_NONBASE_OS_VERSIONS) | {OsVersion.BASALT}:
+    if os_version == OsVersion.BASALT:
```
dirkmueller, line 141: it should use the "new" name for tumbleweed also, not just for BASALT.

dirkmueller (changes requested):

src/bci_build/package/appcontainers.py
dcermak: Makes sense, I have included Tumbleweed in here too.

dirkmueller (approved):
I'm not a fan of having a basalt mariadb and a sle15 mariadb container but its not the right place to discuss that here..

PR #715, opened 2023-10-16
Split package module further to avoid recursive imports

dirkmueller:
This completes the previous modularization effort by splitting into globals, dataclasses and utils.


src/bci_build/package/appcontainers.py
```diff
@@ -2,19 +2,19 @@
...
-from bci_build.package import Package
-from bci_build.package import PackageType
-from bci_build.package import Replacement
-from bci_build.package import SupportLevel
+from .dataclasses import ApplicationStackContainer
+from .dataclasses import OsContainer
+from .dataclasses import Package
+from .dataclasses import PackageType
+from .dataclasses import Replacement
+from .dataclasses import SupportLevel
```
dcermak, line 10: Please use a different name, this is imho ambiguous as the name clashes with the stdlib & pydantic dataclass module, but provides classes build on top of it and not those compatible with stdlib dataclasses

src/bci_build/package/globals.py
```diff
@@ -0,0 +1,181 @@
+"""This module contains global variables and enums used in the BCI Dockerfile generator."""
```
dcermak, line 1: I'd suggest to call this module constants instead

src/bci_build/package/globals.py
```diff
@@ -0,0 +1,181 @@
...
+from __future__ import annotations
+
+import datetime
+import enum
+from typing import List
+
+
+_BASH_SET = "set -euo pipefail"
```
dcermak, line 9: Since this is an exported symbol, it shouldn't be private
```suggestion
BASH_SET = "set -euo pipefail"
```

dcermak (changes requested):
Building the documentation fails:
```
/__w/BCI-dockerfile-generator/BCI-dockerfile-generator/source/adding_containers.rst:6:py:class reference target not found: bci_build.package.LanguageStackContainer
```

src/bci_build/package/globals.py
dirkmueller: done, that needs followup fixes.

src/bci_build/package/appcontainers.py
dirkmueller: changed to bciclasses.

PR #723, opened 2023-10-17
Add sles-ltss-release package in the bci-base-fips container

dirkmueller:
This is needed for a registration to have the right set of product channels available.


dirkmueller:
> The code looks good to me, but there are no changes to the deployed templates. Is that intentional?

the bci-dockerfile-generator is not running for SP3 anymore. I'm running it manually

PR #727, opened 2023-10-24
Create rundir for the tw containers

dirkmueller:
PostgreSQL moved from /var/run to /run which is usually on tmpfs except inside containers. Outside containers this directory would be created by systemd-tmpfiles generator. Here we need to create it ourselves.


dcermak (approved):
LGTM, but why didn't our tests fail due to the directory missing?

dirkmueller:
> LGTM, but why didn't our tests fail due to the directory missing?

they are failing in https://github.com/SUSE/BCI-tests/pull/269

PR #734, opened 2023-10-26
Introduce golang-openssl 1.21 as stable

dirkmueller:


dirkmueller:
(waiting on go1.21-openssl being released)

PR #735, opened 2023-10-26
Remove app containers from 15 SP4

dirkmueller:


dcermak (commented):
LGTM, but why just git? Just asking because the PR title says "Remove app containers", but it leaves postgres 14 there.

PR #751, opened 2023-11-16
Switch nodejs 20 BCI to SLE 15 SP5 from SP6

dcermak:


src/bci_build/package/node.py
```diff
@@ -24,10 +24,7 @@ def _get_node_kwargs(ver: Literal[16, 18, 20], os_version: OsVersion):
     return {
         "name": "nodejs",
         "os_version": os_version,
-        "is_latest": (
-            (ver == 18 and os_version == OsVersion.SP5)
-            or (ver == 20 and os_version == OsVersion.TUMBLEWEED)
-        ),
+        "is_latest": ver == 20,
```
dirkmueller, line 27: that will add a "latest" entry for the SP6 containers.. I think we only want it on SP5 currently? why not simply change the `ver ==18 and os_version == OsVersion.SP5` to `ver == 20 and os_version == OsVersion.SP5` ?

src/bci_build/package/node.py
```diff
@@ -58,7 +55,7 @@ def _get_node_kwargs(ver: Literal[16, 18, 20], os_version: OsVersion):
         **_get_node_kwargs(18, OsVersion.SP5), support_level=SupportLevel.L3
     ),
     LanguageStackContainer(
-        **_get_node_kwargs(20, OsVersion.SP6), support_level=SupportLevel.L3
+        **_get_node_kwargs(20, OsVersion.SP5), support_level=SupportLevel.L3
```
dirkmueller, line 58: but we also need to have it for SP6.. I suggest to duplicate this for SP5 and keep the SP6 entry.

dirkmueller (changes requested):
unclear why we're removing nodejs20 from sp6 rather than simply adding it to SP5.

src/bci_build/package/node.py
dcermak: I was under the mistaken impression that we did not produce language stack containers for sp6 yet.

src/bci_build/package/node.py
dcermak: Because we have nodejs20 in sp5 now and would like to have that one as `latest` too. I've amended the commit to fix this.

dirkmueller (approved):

PR #776, opened 2023-12-13
Add a trivy container (jsc#PED-7439)

dirkmueller:


dcermak (changes requested):
FTBFS in SLES

dirkmueller:
> FTBFS in SLES

to my surprise trivy is only in package hub and not in container module. so we can't build a bci container against sle. not that it would matter match. so restricted to tumbleweed for now.

dcermak (approved):

PR #783, opened 2023-12-15
Remove com.suse.supportlevel and com.suse.eula from base and language

dirkmueller:
containers

In case the derived containers are not overwriting those (for example they're not done by suse), they're a confusing


dcermak (changes requested):
No, this needs to be discussed with PM first.

dirkmueller:
> No, this needs to be discussed with PM first.

Dan, this is a *draft* PR that I made *for* *the* *purpose* of discussing it with the related stakeholders.

PR #793, opened 2023-12-20
Only support MARIADB_RANDOM_ROOT_PASSWORD on Tumbleweed

dcermak:
pwqgen is not available in SLE and hence we should not support/advertise it there

### DIFF
--- a/src/bci_build/package/appcontainers.py
+++ b/src/bci_build/package/appcontainers.py
@@ -15,6 +15,7 @@ from bci_build.package import Package
 from bci_build.package import PackageType
 from bci_build.package import Replacement
 from bci_build.package import SupportLevel
+from jinja2 import Template
 
 
 _PCP_FILES = {}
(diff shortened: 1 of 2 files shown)


dirkmueller:
Thanks for the PR. I think this script is an outdated copy of the official mariadb entry point. I was hoping to be able to merge it and make it auto update from upstream. My plan was to add a patch that makes it work on SLE instead of cutting out the feature.

If we think it's generally unattractive as you want to set the pw from orchestration we should probably patch it out globally, as we're currently not testing it

dcermak:
> Thanks for the PR. I think this script is an outdated copy of the official mariadb entry point. I was hoping to be able to merge it and make it auto update from upstream. My plan was to add a patch that makes it work on SLE instead of cutting out the feature.
> 
> If we think it's generally unattractive as you want to set the pw from orchestration we should probably patch it out globally, as we're currently not testing it

Makes sense, let's fix this properly. I have created https://github.com/SUSE/BCI-dockerfile-generator/issues/794 for that

PR #808, opened 2024-01-02
Fix unit tests failure in 2024

dirkmueller:
programmatically replace it with the current year to avoid a similar breakage in 2025.


dirkmueller:
unit tests are passing now, which is hopefully sufficient as a merge criteria

PR #850, opened 2024-01-12
Include url to the Readme and to the logo of all BCIs

dcermak:
TODO:
- [x] improve the README

### DIFF
--- a/src/staging/bot.py
+++ b/src/staging/bot.py
@@ -235,11 +235,7 @@ class StagingBot:
         where the build recipes are checked out by default.
 
         """
-        return (
-            str(self.os_version)
-            if self.os_version in (OsVersion.TUMBLEWEED, OsVersion.BASALT)
-            else f"sle15-sp{str(self.os_version)}"
-        )
+        return self.os_version.deployment_branch_name
 
     @property
     def package_names(self) -> list[str] | None:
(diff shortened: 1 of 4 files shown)


tests/test_build_recipe.py
```diff
@@ -32,6 +32,7 @@
 LABEL org.opencontainers.image.created="%BUILDTIME%"
 LABEL org.opencontainers.image.vendor="SUSE LLC"
 LABEL org.opencontainers.image.source="%SOURCEURL%"
+LABEL io.artifacthub.package.readme-url="https://raw.githubusercontent.com/SUSE/BCI-dockerfile-generator/sle15-sp4/.readme/test-image/README.md"
```
dirkmueller, line 35: what's the purpose of this extra `/.readme` directory hierarchy? wouldn't it be easier to maintain to store this alongside the Dockerfile in the source directory (so that everything is together)?

tests/test_build_recipe.py
```diff
@@ -426,6 +436,7 @@
                 entrypoint=["/usr/bin/emacs"],
                 cmd=["/usr/bin/gcc"],
                 maintainer="invalid@suse.com",
+                logo_url="https://suse.com/assets/emacs-logo.svg",
```
dirkmueller, line 439: I wonder if we need to move that to a repository e..g under opensource.suse.com/bci/assets/ and maintain it there..

tests/test_build_recipe.py
dcermak: Including the README in the package directory means that it ends up on OBS/IBS. Currently the README is pretty useless for most packages, hence I thought about not adding it to the main package. But I don't insist on that.

tests/test_build_recipe.py
dcermak: Couldn't that make us liable for the content and the licenses? If we just link to it, we should be good. IANAL applies of course

dirkmueller: most logos are CC, so you only have trademark. and you have that issue in either case.

src/staging/bot.py
```diff
@@ -1484,7 +1487,7 @@ async def find_missing_packages_on_obs(self) -> list[str]:
         deployment_branch_head = repo.commit(f"origin/{self.deployment_branch_name}")
         pkgs_in_deployment_branch = {
             tree.name for tree in deployment_branch_head.tree
-        } - {".obs", ".github", "_config"}
+        } - {".obs", ".github", "_config", ".readme"}
```
dirkmueller, line 1490: this part is now obsolete?

dcermak: It is, thanks for the catch!

dirkmueller (approved):

PR #877, opened 2024-01-29
Drop black & reorder-python-imports for ruff

dcermak:
black & reorder-python-imports started conflicting in their formating, so let's just use one tool that does both: ruff


pyproject.toml
```diff
@@ -18,18 +18,35 @@ aiofiles = ">=22.1"
...
+    "F",
+    # Pycodestyle
+    "E",
+    "W",
+    # isort
+    "I"
+]
+ignore = [
+    # too long lines…
+    "E501"
```
dirkmueller, line 45: why? we should set the line-length to the "right" value (whatever we consider right, BCI-tests uses 79)

dirkmueller: ok, I see we ignored this so far.

pyproject.toml
```diff
@@ -18,18 +18,35 @@ aiofiles = ">=22.1"
...
+]
+
+[tool.ruff]
+src = ["src", "tests", "source"]
```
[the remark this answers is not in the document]
dirkmueller: did so now

dirkmueller (approved):

PR #890, opened 2024-02-01
Default OCI annotations to more sensible defaults

dirkmueller:
link to BCI landing page


src/bci_build/package/__init__.py
```diff
@@ -1325,6 +1329,8 @@ def title(self) -> str:
 
 @dataclass
 class OsContainer(BaseContainerImage):
+    logo_url: str = ""
```
dcermak, line 1332: Wouldn't it make sense for these to use the BCI logo too?

src/bci_build/package/__init__.py
```diff
@@ -502,7 +502,7 @@ class BaseContainerImage(abc.ABC):
     #: URL to the logo of this container image.
     #: This value is added to the ``io.artifacthub.package.logo-url`` label if
     #: present
-    logo_url: str = ""
+    logo_url: str = "https://opensource.suse.com/bci/sle-bci-plain-pos.png"
```
dcermak, line 505: Do we have an svg?

src/bci_build/package/__init__.py
dirkmueller: The challenge I had is that OsContainer is used for a mix of containers that are not BCI, like for example the LTSS or fips containers (or opensuse containers). actually the opensuse issue is general, so I need to find a different solution :/

src/bci_build/package/__init__.py
dirkmueller: I don't unfortunately, I only have the single png (that I captured myself). I'll chase if I can find the orginal source and logos somewhere.

dcermak (approved):

PR #902, opened 2024-02-08
🤖: Update build recipes from templates

dirkmueller:


dirkmueller:
/vc user=dancermak switch golang stable to 1.22, oldstable to 1.21

PR #913, opened 2024-02-26
Add openssh-clients and catatonit to git

weyfonk:
openssh-clients should enable git repositories to be cloned via SSH.
Catatonit is needed for some use cases involving jobs, such as [1].

[1]: https://github.com/rancher/gitjob/commit/b2a3eaf53efd

### DIFF
--- a/src/bci_build/package/appcontainers.py
+++ b/src/bci_build/package/appcontainers.py
@@ -517,7 +517,11 @@ GIT_CONTAINERS = [
         license="GPL-2.0-only",
         package_list=[
             Package(name, pkg_type=PackageType.BOOTSTRAP)
-            for name in ("git-core",)
+            for name in (
+                "catatonit",
+                "git-core",
+                "openssh-clients",
+            )
             + (() if os_version == OsVersion.TUMBLEWEED else ("skelcd-EULA-bci",))
         ],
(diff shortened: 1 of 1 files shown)


src/bci_build/package/appcontainers.py
```diff
@@ -517,7 +517,11 @@ def _get_nginx_kwargs(os_version: OsVersion):
         license="GPL-2.0-only",
         package_list=[
             Package(name, pkg_type=PackageType.BOOTSTRAP)
-            for name in ("git-core",)
+            for name in (
+                "catatonit",
```
dcermak, line 521: Why catatonit?

dcermak, line 521: The use case from https://github.com/rancher/gitjob/commit/b2a3eaf53efd is imho too niche for a general purpose git image, I hence would not include `catatonit` in the image.
```suggestion
```

dcermak (changes requested):

src/bci_build/package/appcontainers.py
dcermak: Not all containers need a separate process to be PID 1, actually so far it has been the exception rather than the rule.

PR #936, opened 2024-03-04
Fix labelprefix for FIPS LTSS images

### LINKED ISSUE #922: FIPS LTSS 15.3 image uses the labelprefix `com.suse.bci` and not `com.suse.sle`
Test runs of the SLE 15.3 FIPS LTSS image are currently failing on OSD:
- https://openqa.suse.de/tests/13599317#step/_root_BCI-tests_metadata_/2
- https://openqa.suse.de/tests/13599318#step/_root_BCI-tests_metadata_/4

The underlying issue is that the testsuite is assuming that the labelprefix for an LTSS image is `com.suse.sle` and not `com.suse.bci` (as it is currently the case with the FIPS LTSS image: https://github.com/SUSE/BCI-dockerfile-generator/blob/9a2acabbfe76bd1ef276a06091b18ed4dbe8b00c/src/bci_build/package/basecontainers.py#L104

For consistency's sake, it would be probably better to change the labelprefix in the dockerfile generator for images under LTSS.

danishprakash:
Closes #922

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -62,6 +62,7 @@ class ReleaseStage(enum.Enum):
 class ImageType(enum.Enum):
     """Values of the ``image-type`` label of a BCI"""
 
+    LTSS = "ltss"
     SLE_BCI = "sle-bci"
     APPLICATION = "application"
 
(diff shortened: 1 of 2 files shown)


src/bci_build/package/__init__.py
```diff
@@ -1358,6 +1358,13 @@ def reference(self) -> str:
         return f"{self.registry}/{self._registry_prefix}/bci-{self.name}:{self.version_label}"
 
 
+@dataclass
+class OsFIPSContainer(OsContainer):
```
dcermak, line 1362: I would prefer a different name here, e.g.
```suggestion
class LTSSContainer(OsContainer):
```

PR #963, opened 2024-03-22
Add MariaDB healthcheck.sh script

alexandrevicenzi:
The script can run many tests as well be used in the container healthcheck instruction to check health status.

### DIFF
--- a/src/bci_build/package/appcontainers.py
+++ b/src/bci_build/package/appcontainers.py
@@ -130,6 +130,21 @@ HEALTHCHECK --start-period=5m --timeout=5s --interval=5s --retries=2 \
 ]
 
 _MARIADB_ENTRYPOINT = (Path(__file__).parent / "mariadb" / "entrypoint.sh").read_bytes()
+_MARIADB_HEALTHCHECK = (
+    Path(__file__).parent / "mariadb" / "healthcheck.sh"
+).read_bytes()
+_MARIADB_GOSU = b"""#!/bin/bash
+
+u=$1
+shift
+
+if ! id -u $u > /dev/null 2>&1; then
+    echo "Invalid user: $u"
+    exit 1
+fi
+
+setpriv --reuid=$u --regid=$u --clear-groups -- /bin/bash "$@"
+"""
 
 MARIADB_CONTAINERS = []
 MARIADB_CLIENT_CONTAINERS = []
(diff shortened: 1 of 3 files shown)


alexandrevicenzi:
This is related to https://github.com/SUSE/BCI-tests/issues/405

src/bci_build/package/appcontainers.py
```diff
@@ -130,6 +130,20 @@
 ]
 
 _MARIADB_ENTRYPOINT = (Path(__file__).parent / "mariadb" / "entrypoint.sh").read_bytes()
+_MARIADB_HEALTHCHECK = (Path(__file__).parent / "mariadb" / "healthcheck.sh").read_bytes()
+_MARIADB_GOSU = b"""#!/usr/bin/env bash
```
dirkmueller, line 134: why /usr/bin/env bash here when you hardcode /bin/bash later? you could use the same in both, and I think env is unnecessary here. we knwo what the location of bash is in SLE BCI containers

src/bci_build/package/appcontainers.py
```diff
@@ -130,6 +130,20 @@
...
+
+u=$1
+shift
+opts=$*
```
dirkmueller, line 138: this breaks arguments with spaces. suggest to delete this line

src/bci_build/package/appcontainers.py
```diff
@@ -130,6 +130,20 @@
...
+
+if ! id -u $u > /dev/null 2>&1; then
+    echo "Invalid user: $u"
+    exit 1
+fi
+
+setpriv --reuid=$u --regid=$u --clear-groups -- /bin/bash $opts
```
dirkmueller, line 145:
```suggestion
setpriv --reuid=$u --regid=$u --clear-groups -- /bin/bash "$@"
```

dirkmueller (approved):

src/bci_build/package/appcontainers.py
```diff
@@ -163,6 +178,8 @@
             entrypoint=["docker-entrypoint.sh"],
             extra_files={
                 "docker-entrypoint.sh": _MARIADB_ENTRYPOINT,
+                "healthcheck.sh": _MARIADB_HEALTHCHECK,
+                "gosu": _MARIADB_GOSU,
```
tianon, line 182: This is _not_ `gosu`, nor is it even properly "`gosu` compatible" -- please give the binary in the image a different name.

(If full `gosu` compatibility is desired, please see https://github.com/tianon/gosu/pull/143 which has been successfully tested against `gosu`'s test suite for all the parsing edge cases I'm aware of and maintain a test suite for.)

tianon: Thank you! ❤️

PR #974, opened 2024-03-26
Add postgresql-contrib (bsc#1221938)

dirkmueller:


alexandrevicenzi (approved):

dcermak (changes requested):
This makes the postgresql image over 100MB fatter, we shouldn't do that. Let's have a `-full` version instead as agreed in the sync on April 2nd.

PR #1003, opened 2024-04-09
Remove ArchExclusiveLine that doesn't filter anything out

dcermak:
We build only on the specified architectures, hence the filtering is not necessary. At best we mask errors to install the golang race detector.


dirkmueller (changes requested):
I'm building golang containers for armv7 at the request of rancher org, hence this is actually necessary.

PR #1007, opened 2024-04-10
Restore sizes of minimal container on SP6

dirkmueller:
we should not have libpcre1  in the minimal container as libselinux was switching to libpcre2


dcermak (approved):
.6MB less, nice! Thanks

PR #1013, opened 2024-04-11
Be more specific in which files to delete

dirkmueller:
the PCP containers (and some others) are actually broken by the removal of /var/log/* as they install directories there that are missing in the container.


dcermak (approved):

dirkmueller:
x86_64 builds finished, the rest should be the same..

Issue #993, opened 2024-04-04
Add proper README to images


alexandrevicenzi:
Most images currently have a rather small incomplete README or none at all.

Adding a proper README to our images will help users deploy applications quickly.

Containers missing a README:

- [x] MariaDB Server (#1024)
- [x] MariaDB Client (#1342)
- [x] PostgreSQL (#1118)
- [x] 389 Directory Server (#1299)
- [x] Nginx (#1126)
- [x] Git (#2049)
- [x] Helm (#1276)
- [x] Go (#1169)
- [x] OpenJDK (#1344)
- [x] Python (#1195)
- [x] NodeJS (#1241)
- [x] .NET (#1420)
- [x] Rust (#1236)
- [x] Ruby (#1238)
- [x] PHP (#1156)
- [x] Kernel Module (#1542)
- [x] RMT (#1141)

alexandrevicenzi:
MariaDB Server is done.
PostgreSQL is up for review.
Starting to work on Nginx.

Priyankasaggu11929:
I'll work on adding README for Helm.

dcermak:
I'll take PHP

alexandrevicenzi:
I'll take Go.

dcermak:
I'll take the kernel module container

alexandrevicenzi:
I'll take Python.

alexandrevicenzi:
I'll take Rust.

alexandrevicenzi:
I'll take Ruby.

alexandrevicenzi:
I'll take NodeJS.

alexandrevicenzi:
I'll take 389 DS.

alexandrevicenzi:
I'll take Git.

alexandrevicenzi:
I'll take OpenJDK.

alexandrevicenzi:
I'll take .NET.

PR #998, opened 2024-04-04
Add description for Spack container

e4t:
This implements jsc#PED-7118.

PR #1024, opened 2024-04-15
MariaDB Server README

alexandrevicenzi:
This is related to #993.

### DIFF
--- a/src/bci_build/package/appcontainers.py
+++ b/src/bci_build/package/appcontainers.py
@@ -169,7 +169,13 @@ for os_version in set(ALL_NONBASE_OS_VERSIONS) | {OsVersion.BASALT}:
                     regex_in_build_description="%%mariadb_version%%",
                     package_name="mariadb",
                     parse_version="minor",
-                )
+                ),
+                Replacement(
+                    file_name="README.md",
+                    regex_in_build_description="%%mariadb_version%%",
+                    package_name="mariadb",
+                    parse_version="minor",
+                ),
             ],
             pretty_name="MariaDB Server",
             package_list=["mariadb", "mariadb-tools", "gawk", "timezone", "util-linux"],
@@ -226,7 +232,7 @@ COPY gosu /usr/local/bin/gosu
                     regex_in_build_description="%%mariadb_version%%",
                     package_name="mariadb-client",
                     parse_version="minor",
-                )
+                ),
             ],
             pretty_name="MariaDB Client",
             support_level=SupportLevel.L3,
--- a/src/bci_build/package/mariadb/README.md.j2
+++ b/src/bci_build/package/mariadb/README.md.j2
@@ -1,0 +1,64 @@
+# {{ image.pretty_name }} Container Image
+
+MariaDB is an open-source, multi-threaded, relational database management system. It's a backward-compatible branch of the MySQL Community Server that provides a drop-in replacement for MySQL.
+
+## Using the image
+
+By default, the image launches MariaDB with the same configuration that comes with SUSE Linux Enterprise Server, with two exceptions: logging is sent to stdout, and binding to `localhost` is disabled.
+
+The only environment variable required to start the container is the MariaDB root password.
+
+```ShellSession
+$ podman run -it --rm -p 3306:3306 -e MARIADB_ROOT_PASSWORD=my-password {{ image.pretty_reference }}
+```
+
+or:
+
+```ShellSession
+$ podman run -it --rm -p 3306:3306 -e MARIADB_ALLOW_EMPTY_ROOT_PASSWORD=1 {{ image.pretty_reference }}
+```
+
+## Volumes
+
+The database is stored in the directory `/var/lib/mysql`.
+
+When using the MariaDB image, we recommend one of the following options:
+
+* Use a named volume to manage the storage of your database data. This is the default option and it is easy to use. The downside is that the files may be hard to locate for tools and applications that run directly on the host system, i.e. outside containers.
+* Create a data directory on the host system (outside the container) and mount the created directory to a directory accessible from inside the container. This stores the database files in a known location on the host system and makes it easy for tools and applications on the host system to access the files. If you choose this approach, make sure that the specified directory exists, and its permissions along with other security mechanisms on the host system are set up correctly.
+
+To mount a host directory as a volume for your data run the following command:
+
+```ShellSession
+$ podman run -it --rm -v /my/own/datadir:/var/lib/mysql:Z -p 3306:3306 -e MARIADB_ROOT_PASSWORD=my-password {{ image.pretty_reference }}
+```
+
+The `-v /my/own/datadir:/var/lib/mysql:Z` part of the command mounts the `/my/own/datadir` directory from the underlying host system as `/var/lib/mysql` inside the container, where MariaDB will by default write its data files.
+
+## Environment Variables
+
+One of `MARIADB_RANDOM_ROOT_PASSWORD`, `MARIADB_ROOT_PASSWORD_HASH`, `MARIADB_ROOT_PASSWORD` or `MARIADB_ALLOW_EMPTY_ROOT_PASSWORD` (or equivalents, including `*_FILE`), is required.
+
+All other environment variables are optional.
+
+All environment variables are documented in the MariaDB's Knowledge Base [MariaDB Server Docker Official Image Environment Variables](https://mariadb.com/kb/en/mariadb-server-docker-official-image-environment-variables/).
+
+## Health, liveness and readiness
+
+There are no explicit health checks added to the container image. However, you can use the `healthcheck.sh` script to choose from a limited selection of tests to check for what you consider health, liveness, and readiness.
+
+If there is no database initialized when the container starts, a default database is created. This means that incoming connections are not accepted until the initialization is completed. Use the `healthcheck.sh` script to check for container readiness.
+
+To learn about `healthcheck.sh`, how to use it, and what tests are provided, refer to the MariaDB's Knowledge Base [Using Healthcheck.sh](https://mariadb.com/kb/en/using-healthcheck-sh/).
+
+## Usage against an existing database
+
+If you start a MariaDB container instance with a data directory that already contains a database (specifically, a MySQL subdirectory), no environment variables that control initialization are needed or examined, and no pre-existing databases are changed.
+
+The only exception is the `MARIADB_AUTO_UPGRADE` environment variable. If it's set, it may run `mysql_upgrade` or `mariadb-upgrade`, potentially causing changes to the system tables.
+
+## Backup and restore
+
+Information on how to perform backup and restore can be found on the MariaDB Knowledge Base [Container Backup and Restoration](https://mariadb.com/kb/en/container-backup-and-restoration/).
+
+{% include 'licensing_and_eula.j2' %}
--- a/src/bci_build/package/rmt-mariadb/README.md.j2
+++ b/src/bci_build/package/rmt-mariadb/README.md.j2
@@ -1,0 +1,1 @@
+../mariadb/README.md.j2


src/bci_build/templates.py
```diff
@@ -153,6 +153,12 @@
     <param name="regex">{{ replacement.regex_in_build_description }}</param>
     <param name="package">{{ replacement.package_name }}</param>{% if replacement.parse_version %}
     <param name="parse-version">{{ replacement.parse_version }}</param>{% endif %}
+  </service>
+  <service name="replace_using_package_version" mode="buildtime">
+    <param name="file">README.md</param>
+    <param name="regex">{{ replacement.regex_in_build_description }}</param>
+    <param name="package">{{ replacement.package_name }}</param>{% if replacement.parse_version %}
+    <param name="parse-version">{{ replacement.parse_version }}</param>{% endif %}
```
dcermak, line 161: Please no, this is creating needless `_service` entries. Imho we should be doing something like https://github.com/SUSE/BCI-dockerfile-generator/pull/998/commits/905f710a3b9f0932088832e3c6a00f2e4607fefa instead

```suggestion
```

src/bci_build/package/appcontainers.py
```diff
@@ -129,6 +129,9 @@
     for os_version in ALL_NONBASE_OS_VERSIONS
 ]
 
+_MARIADB_README = (
+    Path(__file__).parent / "mariadb" / "README.server.md.j2"
+).read_bytes()
```
dcermak, line 134: Try this:
```suggestion
).read_text()
```

I suspect that because the readme is read as bytes it is then encoded as utf8 in the `readme` property

src/bci_build/package/mariadb/README.server.md.j2
```diff
@@ -0,0 +1,66 @@
+# {{ image.pretty_name }} Container Image
+
+MariaDB is one of the most popular database servers in the world. It’s made by the original developers of MySQL and guaranteed to stay open source.
```
dcermak, line 3: Found it, you've used a fancy apostrophe
```suggestion
MariaDB is one of the most popular database servers in the world. It's made by the original developers of MySQL and guaranteed to stay open source.
```

src/bci_build/templates.py
alexandrevicenzi: Ok, agree, this was not the best solution. The best approach would be to change `Replacement` indeed.

I can add a `file` attribute to specify the file we want to replace something, but this affects the older images which is not ideal. Unless we Docker/Kiwi template by default, but that can have unwanted side effects one day.

I can add a `regex_in_readme` attribute, this will only affect who uses this, and automatically selects README.md, similar to `regex_in_build_description`.

Which one sounds better, or do you have a better idea?

src/bci_build/package/mariadb/README.server.md.j2
alexandrevicenzi: Good catch, I'll fix it.

src/bci_build/package/__init__.py
```diff
@@ -1019,26 +1019,64 @@ def readme_url(self) -> str:
 
     @property
     def readme(self) -> str:
+        """The contents of the image README that is refered by ``io.artifacthub.package.readme-url`` label.
+
+        The text content is obteined from :py:attr:`BaseContainerImage.extra_files` if
```
dcermak, line 1024:
```suggestion
        The text content is obtained from :py:attr:`BaseContainerImage.extra_files` if
```

src/bci_build/package/__init__.py
```diff
@@ -1019,26 +1019,64 @@ def readme_url(self) -> str:
...
+        ``README.md`` or ``README.md.j2`` are defined.
+
+        If no files are explicitly defined in :py:attr:`BaseContainerImage.extra_files`
+        but they are available in the image folder, assuming :py:attr:`BaseContainerImage.name`
+        as the folder name, they are going to be used.
+
+        ``README.md`` will be copied as is while ``README.md.j2`` will be compiled
+        before being returned. ``README.md`` takes precedence over ``README.md.j2``.
```
dcermak, line 1032: I know this just became apparent due to you writing the docs, so feel free to ignore it, but imho this behavior is confusing. If there's a README.md and a README.md.j2, then we should error out. I cannot imagine a situation where having both is **not** a bug.

src/bci_build/package/appcontainers.py
```diff
@@ -129,6 +129,9 @@
     for os_version in ALL_NONBASE_OS_VERSIONS
 ]
 
+_MARIADB_README = (
```
dcermak, line 132: Is this even required? If I understood your code correctly, then the dockerfile generator should discover the readme template by itself

src/bci_build/package/__init__.py
alexandrevicenzi: How it works is:

- Try README.md from `extra_files`
- Try README.md.j2 from `extra_files`
- Try README.md from the image folder
- Try README.md.j2 from the image folder

The search stops once one of these matches. If there are two files, the second is simply ignored.

I'm not sure what is the best option here.

In theory, it could raise an error if out of the 4 items, there is more than one that matches, because it is not clear which README should be used. In practice, it seems unlikely that we have both, but we don't test that.

src/bci_build/package/appcontainers.py
alexandrevicenzi: Yes, the file is called `README.server.md.j2`, the magic works if the file is called `README.md.j2`.

We have a server and client MariaDB image, the idea for this name is that I plan to create a `README.client.md.j2` and keep both readmes in the same folder.

I would not object to creating a new folder for mariadb-client and both file names could be `README.md.j2` if you prefer.

src/bci_build/templates.py
alexandrevicenzi: I picked the changes from Egbert.

src/bci_build/package/__init__.py
dirkmueller: imho we should not have README.md from extra files going forward. it's just in transition as we had things in flight. going forward we should have only one way to specify READMEs. either as a verbatim file or as a .j2 variant in the package subdir.

src/bci_build/package/appcontainers.py
dirkmueller: how about simply an if condition inside the `README.md.j2` that goes between client or server?

that seems much simpler to me. Also if we want to share content, we should simply include it from the  shared templates location.

src/bci_build/package/mariadb/README.server.md.j2
```diff
@@ -0,0 +1,66 @@
+# {{ image.pretty_name }} Container Image
+
+MariaDB is one of the most popular database servers in the world. It's made by the original developers of MySQL and guaranteed to stay open source.
```
dirkmueller, line 3: imho leave this out. SUSE should stay neutral, we also offer postgresql and in future might also be offering further things

src/bci_build/package/mariadb/README.server.md.j2
```diff
@@ -0,0 +1,66 @@
...
+## How to use this image
+
+By default, the image launches MariaDB with the same configuration as the one
+that comes with SUSE Linux Enterprise Server. The differences are that logging is
+sent to stdout and binding to `localhost` is disabled.
+
+The only environment variable that is required to start the container is the MariaDB root password.
+
+```ShellSession
+$ podman run -it --rm -p 3306:3306 -e MARIADB_ROOT_PASSWORD=my-password registry.suse.com/suse/{{ image.name }}:{% if image.is_opensuse %}latest{% else %}{{ image.version }}{% endif %}
```
dirkmueller, line 14: no, opensuse images are not on registry.suse.com. I think there is a full variable that has the actual path. if there isn't then we should add that.

src/bci_build/package/mariadb/README.server.md.j2
```diff
@@ -0,0 +1,66 @@
...
+```
+
+or:
+
+```ShellSession
+$ podman run -it --rm -p 3306:3306 -e MARIADB_ALLOW_EMPTY_ROOT_PASSWORD=1 registry.suse.com/suse/{{ image.name }}:{% if image.is_opensuse %}latest{% else %}{{ image.version }}{% endif %}
```
dirkmueller, line 20: same here

src/bci_build/package/mariadb/README.server.md.j2
```diff
@@ -0,0 +1,66 @@
...
+```
+
+## Volumes
+
+The default data directory location is `/var/lib/mysql`.
```
dirkmueller, line 25: it's not the default, it's the only supported location afaik

src/bci_build/package/mariadb/README.server.md.j2
dcermak:
```suggestion
$ podman run -it --rm -p 3306:3306 -e MARIADB_ROOT_PASSWORD=my-password {{ image.reference | replace("-%RELEASE%") }}
```

src/bci_build/package/appcontainers.py
alexandrevicenzi: I considered that, but both READMES would be so far apart that I decided that two templates would be better. I'll give it a try and see how it looks.

src/bci_build/package/mariadb/README.server.md.j2
alexandrevicenzi: I'm not sure, I believe if you change the settings it might be possible to use another location. I'll see if I can come up with a better sentence.

src/bci_build/package/__init__.py
alexandrevicenzi: Now that we have a license partial template that should be used, I don't see a reason for a verbatim file, everything going forward seems to be a template, and even non-jinja files could be parsed as templates with no issues.

Or there are too many cases where Jinja would not be used?

src/bci_build/package/mariadb/README.server.md.j2
dirkmueller: So we support changing the locations and we're sure that it will never be hardcoded anywhere? I think if noone requested that feature we just don't advertise it for now because it's just more work to test and not provide any additional value IMHO

src/bci_build/package/__init__.py
dirkmueller: No, you're right, I think only supporting `README.md.j2` is enough

src/bci_build/package/mariadb/README.server.md.j2
alexandrevicenzi: I updated the message here to imply that this is the only supported location.

src/bci_build/package/mariadb/README.server.md.j2
alexandrevicenzi: Fixed

src/bci_build/package/mariadb/README.server.md.j2
alexandrevicenzi: I replaced this with the RPM package description.

src/bci_build/package/__init__.py
alexandrevicenzi: I updated the code to accept only `README.md.j2` and for that, I also had to update a few old READMEs that used `README.md` as their name.

src/bci_build/package/tomcat/README.md.j2
```diff
@@ -0,0 +1,82 @@
...
+
+For security reasons, the image runs as the **tomcat** user. This means that
+additional packages cannot be installed via `zypper`, unless the user becomes
+`root`.
+
+To deploy an application, copy the `.war` file file into
+`$CATALINA_BASE/webapps` (either during a container build or by bind-mounting
+the directory), and launch the container using the following command:
+```ShellSession
+$ podman run -d --rm -p 8080:8080 registry.suse.com/suse/tomcat:{{ image.version }}
```
dirkmueller, line 24: this still hardcodes the registry.s.com reference. I think we should add a pretty_reference implementation instead and use that here

src/bci_build/package/tomcat/README.md.j2
```diff
@@ -0,0 +1,82 @@
...
+The container image can be used in rootless mode with Podman. Keep in mind that
+Podman remaps the `tomcat` user in the container to a different user on the
+host. This user will not have write access to the mounted directory. To avoid
+permission issues change permissions of the shared directory to `0777` as
+follows:
+
+```ShellSession
+$ chmod 0777 /path/to/my/app
+$ podman run --rm -d -v /path/to/my/app:{{ image.env.CATALINA_HOME }}/webapps:z \
+      -p 8080:8080 registry.suse.com/suse/tomcat:{{ image.version }}
```
dirkmueller, line 41: same hardcoding of registry url

src/bci_build/package/tomcat/README.md.j2
```diff
@@ -0,0 +1,82 @@
...
+
+Tomcat's runtime options can be configured using the environment variables
+`JAVA_OPTS` and `CATALINA_OPTS`. `JAVA_OPTS` specifies general options used for
+the JVM, whereas `CATALINA_OPTS` specifies Tomcat's flags. You can pass the
+options to the container runtime using the `-e` flag:
+```ShellSession
+$ podman run -it --rm \
+      -e JAVA_OPTS="-Xmx1024m" \
+      -p 8080:8080 \
+      registry.suse.com/suse/tomcat:{{ image.version }}
```
dirkmueller, line 58: same

src/bci_build/package/tomcat/README.md.j2
```diff
@@ -0,0 +1,82 @@
...
+```
+
+The image ships with `CATALINA_HOME` set to `{{ image.env.CATALINA_HOME }}` and `CATALINA_BASE` set to
+`{{ image.env.CATALINA_HOME }}`.
```
dirkmueller, line 62: should be CATALINA.BASE

src/bci_build/package/tomcat/README.md.j2
```diff
@@ -0,0 +1,82 @@
...
+
+
+## Samples
+
+By default, the sample applications shipped with Tomcat are not installed in
+the container image. You can obtain them by installing one of the following
+packages:
+- {{ image.name }}{{ image.version }}-webapps
```
dirkmueller, line 70: no, the name is `tomcat` and `tomcat10`. so this does not work.

dirkmueller:
@alexandrevicenzi sorry I did not realize there is a tomcat change in the "mariadb" PR, so I duplicated the work unfortunately in #1073 . it has a slightly different approach taken, please take a look.

alexandrevicenzi:
@dirkmueller merge yours and I'll rebase it.

src/bci_build/package/mariadb/README.server.md.j2
```diff
@@ -0,0 +1,68 @@
...
+By default, the image launches MariaDB with the same configuration as the one
+that comes with SUSE Linux Enterprise Server. The differences are that logging is
+sent to stdout and binding to `localhost` is disabled.
+
+The only environment variable that is required to start the container is the MariaDB root password.
+
+{%- set reference = ( image.reference | replace("-%RELEASE%", "") ) %}
+
+```ShellSession
+$ podman run -it --rm -p 3306:3306 -e MARIADB_ROOT_PASSWORD=my-password {{reference}}
```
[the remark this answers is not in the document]
dirkmueller: actually I think we should highlight the need of a volume mount  in this command, wdyt?

src/bci_build/package/__init__.py
```diff
@@ -1031,26 +1031,50 @@ def readme_url(self) -> str:
...
-                image=self
-            )
+            return self.render_template(readme_template_fname.read_text())
 
         return f"""# The {self.title} Container image
 
 {self.description}
 """
 
+    def render_template(self, text):
```
dirkmueller, line 1065: not a huge fan of another public method on the base class given that its only used in one place? also #1075  will change this

src/bci_build/package/__init__.py
```diff
@@ -1031,26 +1031,50 @@ def readme_url(self) -> str:
...
+        If no file ``README.md.j2`` is explicitly defined in :py:attr:`BaseContainerImage.extra_files`
+        but the file is available in the image folder, assuming :py:attr:`BaseContainerImage.name`
+        as the folder name, this file going to be used.
+
+        If no proper README is found, a default text based on :py:attr:`BaseContainerImage.title`
+        and :py:attr:`BaseContainerImage.description` will be returned.
+
+        """
+        if "README.md.j2" in self.extra_files:
+            tmpl = self.extra_files["README.md.j2"]
```
dirkmueller, line 1048: so this is the 3rd way to configure a README to use. I would prefer to keep it at the current options that exist (README.md as python rendered content and as jinja2 template). yet another way makes it just confusing rather than helpful.

src/bci_build/package/mariadb/README.server.md.j2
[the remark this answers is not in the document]
alexandrevicenzi: `pretty_reference` is a new addition, this PR was older,  but I can fix it.

Not sure if we need to add a volume here, there's a special section for volumes, and you don't need a volume to start the container, but you need one if you plan to re-use it.

src/bci_build/package/__init__.py
alexandrevicenzi: I think I can skip this change.

src/bci_build/package/__init__.py
alexandrevicenzi: If I go with a file name that matches the image this can be removed.

src/bci_build/package/mariadb/README.server.md.j2
```diff
@@ -0,0 +1,68 @@
...
+
+All other environment variables are optional.
+
+The complete list of environment variables is documented on MariaDB's Knowledge Base [MariaDB Server Docker Official Image Environment Variables](https://mariadb.com/kb/en/mariadb-server-docker-official-image-environment-variables/).
+
+## Health, Liveness and Readiness
+
+There's no explicit health checks added to the container image. However, you can use the `healthcheck.sh` script to choose from a (non-exhaustive) list of tests to check for whatever you consider health/liveness/readiness.
+
+If there is no database initialized when the container starts, then a default database will be created. This means that incoming connections will not be accepted until the initialization completes. Use the `healthcheck.sh` script to check for container readiness.
```
dmpop, line 54:
```suggestion
If there is no database initialized when the container starts, a default database is created. This means that incoming connections are not accepted until the initialization is completed. Use the `healthcheck.sh` script to check for container readiness.
```

src/bci_build/package/mariadb/README.server.md.j2
```diff
@@ -0,0 +1,68 @@
...
+
+To learn about `healthcheck.sh`, how to use it and which exact tests are provided check MariaDB's Knowledge Base [Using Healthcheck.sh](https://mariadb.com/kb/en/using-healthcheck-sh/).
```
dmpop, line 56:
```suggestion
To learn about `healthcheck.sh`, how to use it, and what tests are provided, refer to the MariaDB's Knowledge Base [Using Healthcheck.sh](https://mariadb.com/kb/en/using-healthcheck-sh/).
```

src/bci_build/package/mariadb/README.server.md.j2
```diff
@@ -0,0 +1,68 @@
...
+
+## Environment Variables
+
+One of `MARIADB_RANDOM_ROOT_PASSWORD`, `MARIADB_ROOT_PASSWORD_HASH`, `MARIADB_ROOT_PASSWORD` or `MARIADB_ALLOW_EMPTY_ROOT_PASSWORD` (or equivalents, including `*_FILE`), is required.
+
+All other environment variables are optional.
+
+The complete list of environment variables is documented on MariaDB's Knowledge Base [MariaDB Server Docker Official Image Environment Variables](https://mariadb.com/kb/en/mariadb-server-docker-official-image-environment-variables/).
+
+## Health, Liveness and Readiness
```
dmpop, line 50:
```suggestion
## Health, liveness and readiness
```

src/bci_build/package/mariadb/README.server.md.j2
```diff
@@ -0,0 +1,68 @@
...
+
+There's no explicit health checks added to the container image. However, you can use the `healthcheck.sh` script to choose from a (non-exhaustive) list of tests to check for whatever you consider health/liveness/readiness.
+
+If there is no database initialized when the container starts, then a default database will be created. This means that incoming connections will not be accepted until the initialization completes. Use the `healthcheck.sh` script to check for container readiness.
+
+To learn about `healthcheck.sh`, how to use it and which exact tests are provided check MariaDB's Knowledge Base [Using Healthcheck.sh](https://mariadb.com/kb/en/using-healthcheck-sh/).
+
+## Usage against an existing database
+
+If you start your MariaDB container instance with a data directory that already contains a database (specifically, a MySQL subdirectory), no environment variables that control initialization will be needed or examined, and no pre-existing databases will be changed.
```
dmpop, line 60:
```suggestion
If you start a MariaDB container instance with a data directory that already contains a database (specifically, a MySQL subdirectory), no environment variables that control initialization are needed or examined, and no pre-existing databases are changed.
```

src/bci_build/package/mariadb/README.server.md.j2
```diff
@@ -0,0 +1,68 @@
...
+
+The only exception is the `MARIADB_AUTO_UPGRADE` environment variable, that if set, might cause `mysql_upgrade` or `mariadb-upgrade` to run, which might change the system tables.
```
dmpop, line 62:
```suggestion
The only exception is the `MARIADB_AUTO_UPGRADE` environment variable. If it's set, it may run `mysql_upgrade` or `mariadb-upgrade`, potentially causing changes to the system tables.
```

src/bci_build/package/mariadb/README.server.md.j2
```diff
@@ -0,0 +1,68 @@
...
+
+## Backups and Restores
```
dmpop, line 64:
```suggestion
## Backup and restore
```

src/bci_build/package/mariadb/README.server.md.j2
```diff
@@ -0,0 +1,68 @@
...
+
+Backing up and restoring databases is important in containers too. The documentation on how to do this can be found on the MariaDB Knowledge Base [Container Backup and Restoration](https://mariadb.com/kb/en/container-backup-and-restoration/).
```
dmpop, line 66:
```suggestion
Information on how to perform backup and restore can be found on the MariaDB Knowledge Base [Container Backup and Restoration](https://mariadb.com/kb/en/container-backup-and-restoration/).
```

src/bci_build/package/__init__.py
```diff
@@ -1031,26 +1031,50 @@ def readme_url(self) -> str:
 
     @property
     def readme(self) -> str:
-        if "README.md" in self.extra_files:
-            if isinstance(self.extra_files["README.md"], bytes):
-                return self.extra_files["README.md"].decode("utf-8")
-            return str(self.extra_files["README.md"])
+        """The contents of the image README that is refered by ``io.artifacthub.package.readme-url`` label.
```
dcermak, line 1034:
```suggestion
        """The contents of the image README that is referred by the``io.artifacthub.package.readme-url`` label.
```

src/bci_build/package/__init__.py
dcermak: I do agree with Dirk here, I don't think that we should provide another method to obtain the readme.

dcermak (changes requested):

src/bci_build/package/mariadb/README.server.md.j2
```diff
@@ -0,0 +1,68 @@
+# {{ image.pretty_name }} Container Image
+
+MariaDB is an open-source, multi-threaded, relational database management system. It's a backward compatible, drop-in replacement branch of the MySQL Community Server.
```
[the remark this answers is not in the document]
alexandrevicenzi: Ok, but this was copy-paste from the package description.

src/bci_build/package/mariadb/README.server.md.j2
```diff
@@ -0,0 +1,68 @@
...
+$ podman run -it --rm -p 3306:3306 -e MARIADB_ALLOW_EMPTY_ROOT_PASSWORD=1 {{reference}}
+```
+
+## Volumes
+
+The database is stored in the directory `/var/lib/mysql`.
+
+We encourage users of the MariaDB images to use one of the following options:
+
+* Use a named volume using the container manager to manage the storage of your database data by writing the database files to disk on the host system using its own internal volume management. This is the default and is easy and fairly transparent to the user. The downside is that the files may be hard to locate for tools and applications that run directly on the host system, i.e. outside containers.
```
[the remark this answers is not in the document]
alexandrevicenzi: > Use a named volume to manage the storage of your database data. This is the default option and is easy and fairly transparent to the user.

Does this sound better? There's no need to go too much into details, most users should be aware of named volumes and if not, they can read docker/podman documentation.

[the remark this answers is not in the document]
dmpop: Much better! :+1:

src/bci_build/package/__init__.py
alexandrevicenzi: This bit of code has been dropped.

dirkmueller (approved):
LGTM

src/bci_build/package/appcontainers.py
```diff
@@ -169,7 +169,13 @@
...
                     package_name="mariadb",
                     parse_version="minor",
-                )
+                ),
+                Replacement(
+                    file_name="README.md",
+                    regex_in_build_description="%%mariadb_version%%",
+                    package_name="mariadb",
+                    parse_version="minor",
+                ),
```
dirkmueller, line 178: actually this part seems unnecessary now, there is no `%%mariadb_version%%` as far as I can see

src/bci_build/package/mariadb/README.md.j2
```diff
@@ -0,0 +1,64 @@
...
+* Create a data directory on the host system (outside the container) and mount the created directory to a directory accessible from inside the container. This stores the database files in a known location on the host system and makes it easy for tools and applications on the host system to access the files. If you choose this approach, make sure that the specified directory exists, and its permissions along with other security mechanisms on the host system are set up correctly.
+
+To mount a host directory as a volume for your data run the following command:
+
+```ShellSession
+$ podman run -it --rm -v /my/own/datadir:/var/lib/mysql:Z -p 3306:3306 -e MARIADB_ROOT_PASSWORD=my-password {{ image.pretty_reference }}
+```
+
+The `-v /my/own/datadir:/var/lib/mysql:Z` part of the command mounts the `/my/own/datadir` directory from the underlying host system as `/var/lib/mysql` inside the container, where MariaDB by default will write its data files.
```
dcermak, line 36:
```suggestion
The `-v /my/own/datadir:/var/lib/mysql:Z` part of the command mounts the `/my/own/datadir` directory from the underlying host system as `/var/lib/mysql` inside the container, where MariaDB will by default write its data files.
```

src/bci_build/package/mariadb/README.md.j2
```diff
@@ -0,0 +1,64 @@
...
+
+If there is no database initialized when the container starts, a default database is created. This means that incoming connections are not accepted until the initialization is completed. Use the `healthcheck.sh` script to check for container readiness.
+
+To learn about `healthcheck.sh`, how to use it, and what tests are provided, refer to the MariaDB's Knowledge Base [Using Healthcheck.sh](https://mariadb.com/kb/en/using-healthcheck-sh/).
+
+## Usage against an existing database
+
+If you start a MariaDB container instance with a data directory that already contains a database (specifically, a MySQL subdirectory), no environment variables that control initialization are needed or examined, and no pre-existing databases are changed.
+
+The only exception is the `MARIADB_AUTO_UPGRADE` environment variable. If it's set, it may run `mysql_upgrade` or `mariadb-upgrade`, potentially causing changes to the system tables.
```
dcermak, line 58: It "may run"? That sounds a bit random. I guess this only happens if the mariadb version changed?

dcermak (commented):
LGTM, only added two nitpicks

src/bci_build/package/appcontainers.py
alexandrevicenzi: true, I missed it

alexandrevicenzi: It is still used, the URL created is `registry.suse.com/suse/rmt-mariadb:%%mariadb_version%%`.

The property `pretty_reference` uses the `version` attribute, which is set to `%%mariadb_version%%`.

dirkmueller: Ah thank you.

dirkmueller (approved):

PR #1025, opened 2024-04-16
Add gcc container image

dcermak:


src/bci_build/package/gcc/README.md.j2
```diff
@@ -0,0 +1,95 @@
...
+
+
+## How to use this image
+
+### Compile your application in a `Dockerfile`
+
+Generally you'd want to compile your application and distribute it as part of a
+container image of your own. For that, create a `Dockerfile` as follows which
+builds a container image based on this image, copies the sources to a working
+directory and compiles the application:
```
dmpop, line 16:
```suggestion
Normally, you'd want to compile an application and distribute it as part of a custom
container image. To do this, create a `Dockerfile` similar to the one below. The `Dockerfile` uses this image to build a custom container image, copies the sources to a working directory, and compiles the application:
```

src/bci_build/package/gcc/README.md.j2
```diff
@@ -0,0 +1,95 @@
...
+
+FROM registry.suse.com/bci/bci-micro:latest
+WORKDIR /build/
+COPY --from=builder /src/a.out /build/app
+CMD ["/build/app"]
+```
+
+Note that you must build a static binary to deploy it into bci-micro as
+otherwise shared libraries might be missing. You cannot deploy such an app into
+a `scratch` image as it is not possible to statically link glibc.
```
dmpop, line 44:
```suggestion
Note that you must build a static binary to deploy it into bci-micro;
otherwise shared libraries might be missing. You cannot deploy such an app into
a `scratch` image, as it is not possible to statically link glibc.
```

src/bci_build/package/gcc/README.md.j2
```diff
@@ -0,0 +1,95 @@
...
+
+
+### Available build systems
+
+The Container Image comes preloaded with `make` by default. Other build systems
+and related utilities are available in the repository and can be installed via
+`zypper`, including:
```
dmpop, line 51:
```suggestion
The container image comes with `make` by default. Other build systems
and related utilities are available in the repository, and they can be installed using
`zypper`. This includes the following:
```

src/bci_build/package/gcc/README.md.j2
```diff
@@ -0,0 +1,95 @@
...
+- `meson`
+- `cmake`
+- `ninja`
+- `autoconf` & `automake`
+
+
+### Available compiler frontends
+
+The GNU Compiler Collections supports a multitude of frontends. The container
```
dmpop, line 60:
```suggestion
The GNU Compiler Collections supports a wide range of frontends. The container
```

src/bci_build/package/gcc/README.md.j2
```diff
@@ -0,0 +1,95 @@
...
+image ships the C and C++ frontends available as `gcc` and `g++`
+respectively. The following additional frontends can be installed from the
+repository:
+- `gcc{{ image.version }}-fortran` for Fortran support
```
alexandrevicenzi, line 64: This seems like a maintenance nightmare once versions change and more features are added or removed.

Can we point to our GCC documentation for each release? Something like this https://documentation.suse.com/sbp/server-linux/html/SBP-GCC-12/index.html or another more appropriate if one exists.

src/bci_build/package/gcc/README.md.j2
```diff
@@ -0,0 +1,95 @@
...
+{% endif %}{%- if (image.version >= 13) %}- `gcc{{ image.version }}-m2` for the Modula 2 compiler frontend
+{% endif %}{%- if (image.version >= 14) %}- `gcc{{ image.version }}-rust` for the GNU Rust compiler
+{% endif %}{% endif %}
+
+### Using the container image interactively
+
+You can also utilize this container image in a one-shot fashion where building a
+full container image is not desirable. The easiest way is to mount the working
+directory of your application into the launched container and compile your
+application there:
```
dmpop, line 78:
```suggestion
You can use the image to create ephemeral containers that execute only gcc. This can be useful in situations, where building a
full container image is not practical. One way to do this is to mount the working directory of an application into the launched container and compile the
application there:
```

src/bci_build/package/gcc/README.md.j2
```diff
@@ -0,0 +1,95 @@
...
+```
+or by invoking `make`
+```bash
+podman run --rm -it -v $(pwd):/src/:Z --workdir /src/ \
+    {{ reference_no_release }} \
+    make
+```
+
+Note that the binary built via this process will most likely not work on your
+local machine. It will only work on operating systems that are binary compatible
```
dmpop, line 92:
```suggestion
Note that the binary built using this approach are unlikely to work on a
local machine. They only work on operating systems that are binary-compatible.
```

src/bci_build/package/gcc.py
```diff
@@ -0,0 +1,86 @@
+from typing import Literal
```
dirkmueller, line 1: a module docstring would be nice here

src/bci_build/package/gcc.py
```diff
@@ -0,0 +1,86 @@
...
+        ln -f $gcc_bin $(echo "$gcc_bin" | sed -e 's|/usr/bin/|/usr/local/bin/|' -e 's|-{gcc_version}$||'); \
+    done
+"""
+            if not _is_main_gcc(os_version, gcc_version)
+            else ""
+        ),
+    )
+    for (gcc_version, os_version) in (
+        (7, OsVersion.SP5),
+        (12, OsVersion.SP5),
```
dirkmueller, line 76: 13 should be also available on sp5 ?

src/bci_build/package/gcc.py
```diff
@@ -0,0 +1,86 @@
...
+        (7, OsVersion.SP6),
+        (13, OsVersion.SP6),
+        # (13, OsVersion.SLCC_DEVELOPMENT),
+        # (13, OsVersion.SLCC_PRODUCTION),
+        (7, OsVersion.TUMBLEWEED),
```
dirkmueller, line 81: imho lets not offer 7 anymore on tumbleweed

src/bci_build/package/gcc.py
```diff
@@ -0,0 +1,86 @@
...
+{DOCKERFILE_RUN} for gcc_bin in $(rpm -ql {gcc_pkg} {gpp} |grep ^/usr/bin/ ); do \
+        ln -f $gcc_bin $(echo "$gcc_bin" | sed -e 's|/usr/bin/|/usr/local/bin/|' -e 's|-{gcc_version}$||'); \
+    done
+"""
+            if not _is_main_gcc(os_version, gcc_version)
+            else ""
+        ),
+    )
+    for (gcc_version, os_version) in (
+        (7, OsVersion.SP5),
```
dirkmueller, line 75: is anyone actually asking for 7 being provided?

dcermak: No one is asking for any of this specifically. The requirement is "create a gcc container", which is about as unspecific as it can get.

dcermak: It would make sense to ship 7 for ada support, which is not available for gcc 12 & 13 in the SLE_BCI repo

src/bci_build/package/gcc/README.md.j2
dcermak: > This seems like a maintenance nightmare once versions change and more features are added or removed.

Imho it's really only a mess for Tumbleweed. For SLE we have only Fortran & Ada. Since we don't show the readme for Tumbleweed anywhere, we could also just remove the TW specific part.

> Can we point to our GCC documentation for each release? Something like this https://documentation.suse.com/sbp/server-linux/html/SBP-GCC-12/index.html or another more appropriate if one exists.

Unfortunately we do not have a generic manual, nor one for gcc 13.

src/bci_build/package/gcc.py
```diff
@@ -0,0 +1,88 @@
...
+
+from bci_build.package import CAN_BE_LATEST_OS_VERSION
+from bci_build.package import DOCKERFILE_RUN
+from bci_build.package import LanguageStackContainer
+from bci_build.package import OsVersion
+from bci_build.package import Replacement
+
+
+def _is_latest_gcc(os_version: OsVersion, gcc_version: Literal[7, 12, 13, 14]) -> bool:
+    if os_version == OsVersion.TUMBLEWEED and gcc_version == 14:
```
dirkmueller, line 13: currently this is not yet true, we're still at 13

src/bci_build/package/gcc.py
```diff
@@ -0,0 +1,88 @@
...
+        return True
+    if os_version == OsVersion.SP6 and gcc_version == 13:
+        return True
+    if os_version == OsVersion.SP5 and gcc_version == 12:
```
dirkmueller, line 17: both sp5 and sp6 have gcc 13. the toolchain extension is not service pack agnostic so the version between should be always the same

src/bci_build/package/gcc.py
```diff
@@ -0,0 +1,88 @@
...
+        ln -f $gcc_bin $(echo "$gcc_bin" | sed -e 's|/usr/bin/|/usr/local/bin/|' -e 's|-{gcc_version}$||'); \
+    done
+"""
+            if not _is_main_gcc(os_version, gcc_version)
+            else ""
+        ),
+    )
+    for (gcc_version, os_version) in (
+        (7, OsVersion.SP5),
+        (12, OsVersion.SP5),
```
dirkmueller, line 78: so if we really want to offer 12, which I advise against, we should not list it wtihout a support date end. I think the overlap periods ends in the next few weeks

src/bci_build/package/gcc/README.md.j2
```diff
@@ -0,0 +1,95 @@
...
+```
+
+It is also possible to compile a static binary with gcc as part of a multistage
+build:
+
+```Dockerfile
+FROM {{ reference_no_release }} as builder
+WORKDIR /src/
+COPY . /src/
+RUN gcc -static main.c src1.c src2.c
```
dirkmueller, line 34: I am not sure we actually need -static here or should advertise -static (it interacts badly with glibc). also an `-o app` should be listed

src/bci_build/package/gcc/README.md.j2
```diff
@@ -0,0 +1,95 @@
...
+
+### Available build systems
+
+The container image comes with `make` by default. Other build systems and
+related utilities are available in the repository, and they can be installed
+using `zypper`. This includes the following:
+- `meson`
+- `cmake`
+- `ninja`
+- `autoconf` & `automake`
```
dirkmueller, line 55: usually that's called autotools (it includes libtool)

src/bci_build/package/gcc/README.md.j2
dcermak: the `-o app` shouldn't be necessary, as the `COPY` changes the filename

src/bci_build/package/gcc/README.md.j2
dcermak: autotools is not a valid package name, that's why I listed autoconf & automake explicitly

dirkmueller (approved):

PR #1054, opened 2024-04-30
Use the LTSS base containers for LTSS builds

dirkmueller:
This also fixes the default base container to bci-base rather than sle15

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -756,7 +756,13 @@ exit 0
             return "opensuse/tumbleweed:latest"
         if self.os_version == OsVersion.BASALT:
             return f"{_build_tag_prefix(self.os_version)}/bci-base:latest"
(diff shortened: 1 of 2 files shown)


dcermak (changes requested):
Breaks the tests and causes a giant churn of all dockerfiles

dirkmueller:
> Breaks the tests and causes a giant churn of all dockerfiles

so is the "requests changes" on "giant churn of all dockerfiles" or "breaks the tests" ? it's not clear to me what I can do to get a review from you?

dirkmueller:
iirc changes should go in together with #1069

dirkmueller:
In the hope that the LTSS base fix can go in, I've reverted the bci-base change for now.

dcermak:
> iirc changes should go in together with #1069

That's an automated PR that updated the prjconf

dcermak (approved):

PR #1067, opened 2024-05-02
Also install openjdk21 explicitly

dirkmueller:
-headless subpackage is not actually requiring the main jre like expected, so we need to install that one as well to avoid jdk 11 being pulled in.

### DIFF
--- a/src/bci_build/package/appcontainers.py
+++ b/src/bci_build/package/appcontainers.py
@@ -683,7 +683,11 @@ TOMCAT_CONTAINERS = [
                 else f"tomcat{tomcat_major}"
             )
         ]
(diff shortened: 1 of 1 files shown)


dcermak:
> -headless subpackage is not actually requiring the main jre like expected, so we need to install that one as well to avoid jdk 11 being pulled in.

Isn't that a packaging bug?

dirkmueller:
> > -headless subpackage is not actually requiring the main jre like expected, so we need to install that one as well to avoid jdk 11 being pulled in.
> Isn't that a packaging bug?

It probably is, but it's been that way for a while. I can submit a fix, but I would prefer to get CI to green first.

PR #1080, opened 2024-05-03
Add Apache 2 container image

### LINKED ISSUE #1022: Create a apache container image

alexandrevicenzi:
Fixes #1022.

### DIFF
--- a/src/bci_build/package/apache2/README.md.j2
+++ b/src/bci_build/package/apache2/README.md.j2
@@ -1,0 +1,99 @@
+# {{ image.pretty_name }} Container Image
+
+The Apache HTTP Server Project is an effort to develop and maintain an open-source HTTP server for modern operating systems including UNIX and Windows. The goal of this project is to provide a secure, efficient and extensible server that provides HTTP services in sync with the current HTTP standards.
+
+## How to use this image
+
+This image only contains Apache httpd. To start a simple HTML server run:
+
+{%- set reference = ( image.reference | replace("-%RELEASE%", "") ) %}
+
+```ShellSession
+$ podman run -it --rm -p 8080:80 -v /path/to/html:/srv/www/htdocs/ {{reference}}
+```
+
+## Configuration
+
+The Apache configarion is stored in the directory `/etc/apache2/`.
+
+Apache filesystem layout:
+
+```
+/etc/apache2/
+  |-- charset.conv  . . . . . . . . . . . .  for mod_auth_ldap
+  |-- conf.d/
+  |   |-- apache2-manual.conf . . . . . . .  conf that comes with apache2-doc
+  |   |-- mod_php4.conf . . . . . . . . . .  (example) conf that comes with apache2-mod_php4
+  |   `-- ... . . . . . . . . . . . . . . .  other configuration added by packages
+  |-- default-server.conf
+  |-- errors.conf
+  |-- httpd.conf  . . . . . . . . . . . . .  top level configuration file
+  |-- listen.conf
+  |-- magic
+  |-- mime.types -> ../mime.types
+  |-- mod_autoindex-defaults.conf
+  |-- mod_info.conf
+  |-- mod_log_config.conf
+  |-- mod_mime-defaults.conf
+  |-- mod_perl-startup.pl
+  |-- mod_status.conf
+  |-- mod_userdir.conf
+  |-- mod_usertrack.conf
+  |-- server-tuning.conf
+  |-- ssl-global.conf
+  |-- protocols.conf
+  |-- ssl.crl/  . . . . . . . . . . . . . .  PEM-encoded X.509 Certificate Revocation Lists (CRL)
+  |-- ssl.crt/  . . . . . . . . . . . . . .  PEM-encoded X.509 Certificates
+  |-- ssl.csr/  . . . . . . . . . . . . . .  PEM-encoded X.509 Certificate Signing Requests
+  |-- ssl.key/  . . . . . . . . . . . . . .  PEM-encoded RSA Private Keys
+  |-- ssl.prm/  . . . . . . . . . . . . . .  public DSA Parameter Files
+  |-- global.conf
+  |-- loadmodule.conf
+  |-- uid.conf
+  `-- vhosts.d/ . . . . . . . . . . . . . .  put your virtual host configuration (*.conf) here
+      |-- vhost-ssl.template
+      `-- vhost.template
+```
+
+Configuration examples are available in the directory `/usr/share/doc/packages/apache2/conf/`.
+
+To customize the configuration of the httpd server, first obtain the default configuration from the container:
+
+```ShellSession
+$ podman run --rm --entrypoint /bin/cat {{reference}} /etc/apache2/httpd.conf > my-httpd.conf
+```
+
+Once you made any changes, mount the customized httpd server configuration:
+
+```ShellSession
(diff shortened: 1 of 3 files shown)


src/bci_build/package/apache2/README.md.j2
```diff
@@ -0,0 +1,99 @@
...
+
+{%- set reference = ( image.reference | replace("-%RELEASE%", "") ) %}
+
+```ShellSession
+$ podman run -it --rm -p 8080:80 -v /path/to/html:/srv/www/htdocs/ {{reference}}
+```
+
+## Configuration
+
+The Apache configarion is stored in the directory `/etc/apache2/`.
```
dcermak, line 17:
```suggestion
The Apache configuration is stored in the directory `/etc/apache2/`.
```

src/bci_build/package/apache2/README.md.j2
```diff
@@ -0,0 +1,99 @@
...
+
+Configuration examples are available in the directory `/usr/share/doc/packages/apache2/conf/`.
+
+To customize the configuration of the httpd server, first obtain the default configuration from the container:
+
+```ShellSession
+$ podman run --rm --entrypoint /bin/cat {{reference}} /etc/apache2/httpd.conf > my-httpd.conf
+```
+
+Once you made any changes, mount the customized httpd server configuration:
```
dcermak, line 66:
```suggestion
Once you made your changes, mount the customized Apache server configuration:
```

src/bci_build/package/appcontainers.py
```diff
@@ -724,3 +724,32 @@ def _get_nginx_kwargs(os_version: OsVersion):
...
+            Replacement(
+                regex_in_build_description="%%apache2_version%%",
+                package_name="apache2",
+                parse_version="minor",
+            )
+        ],
+        license="Apache-2.0",
+        package_list=["apache2"],
+        entrypoint=["httpd", "-DFOREGROUND"],
+        volumes=["/srv/www/htdocs"],
```
dcermak, line 747: I don't think that this is a good idea. If I remember it correctly, then this is the documentroot, i.e. where you'll put your html & other static files to be served. However, if you declare this directory to be a volume, then it will be **impossible** to build a container with files in this directory.

This will eliminate the use case of building a container with your static files shipped with the container image.

src/bci_build/package/apache2/README.md.j2
```diff
@@ -0,0 +1,99 @@
...
+The Apache HTTP Server Project is an effort to develop and maintain an open-source HTTP server for modern operating systems including UNIX and Windows. The goal of this project is to provide a secure, efficient and extensible server that provides HTTP services in sync with the current HTTP standards.
+
+## How to use this image
+
+This image only contains Apache httpd. To start a simple HTML server run:
+
+{%- set reference = ( image.reference | replace("-%RELEASE%", "") ) %}
+
+```ShellSession
+$ podman run -it --rm -p 8080:80 -v /path/to/html:/srv/www/htdocs/ {{reference}}
```
dcermak, line 12: You can use the recently added `image.pretty_reference` for this instead:

```suggestion
$ podman run -it --rm -p 8080:80 -v /path/to/html:/srv/www/htdocs/ {{image.pretty_reference}}
```

dcermak (changes requested):
I think it's not a good idea to put the documentroot into a volume

src/bci_build/package/appcontainers.py
```diff
@@ -724,3 +724,32 @@ def _get_nginx_kwargs(os_version: OsVersion):
...
+        replacements_via_service=[
+            Replacement(
+                regex_in_build_description="%%apache2_version%%",
+                package_name="apache2",
+                parse_version="minor",
+            )
+        ],
+        license="Apache-2.0",
+        package_list=["apache2"],
+        entrypoint=["httpd", "-DFOREGROUND"],
```
dcermak, line 746: wdyt about upstream's entrypoint: https://github.com/docker-library/httpd/blob/master/2.4/httpd-foreground ?

src/bci_build/package/apache2/README.md.j2
```diff
@@ -0,0 +1,99 @@
+# {{ image.pretty_name }} Container Image
+
+The Apache HTTP Server Project is an effort to develop and maintain an open-source HTTP server for modern operating systems including UNIX and Windows. The goal of this project is to provide a secure, efficient and extensible server that provides HTTP services in sync with the current HTTP standards.
+
+## How to use this image
+
+This image only contains Apache httpd. To start a simple HTML server run:
+
+{%- set reference = ( image.reference | replace("-%RELEASE%", "") ) %}
```
dirkmueller, line 9:
```suggestion
```

src/bci_build/package/appcontainers.py
```diff
@@ -724,3 +724,32 @@ def _get_nginx_kwargs(os_version: OsVersion):
...
+        version_in_uid=False,
+        replacements_via_service=[
+            Replacement(
+                regex_in_build_description="%%apache2_version%%",
+                package_name="apache2",
+                parse_version="minor",
+            )
+        ],
+        license="Apache-2.0",
+        package_list=["apache2"],
```
dirkmueller, line 745: We offer various MPM modules, should the name at least reflect the MPM that was chosen?

src/bci_build/package/appcontainers.py
```diff
@@ -724,3 +724,32 @@ def _get_nginx_kwargs(os_version: OsVersion):
     )
     for tomcat_major, os_version in product(_TOMCAT_VERSIONS, ALL_BASE_OS_VERSIONS)
 ]
+
+APACHE2_CONTAINERS = [
+    ApplicationStackContainer(
+        name="apache2",
+        os_version=os_version,
+        package_name="apache2-image",
```
dirkmueller, line 732: We offer various MPM modules, should the name at least reflect the MPM that was chosen?

src/bci_build/package/appcontainers.py
alexandrevicenzi: we could use it, but I did not see a lot of value from it.

src/bci_build/package/appcontainers.py
alexandrevicenzi: That is a good question, but I'm afraid I can't answer that. Someone else needs to clarify the purpose of this image.

src/bci_build/package/appcontainers.py
alexandrevicenzi: Which name do you suggest? I assumed that the package name would be a good name for the image, but I'm open to alternatives.

src/bci_build/package/apache2/README.md.j2
alexandrevicenzi: I forgot that, but will fix it.

src/bci_build/package/appcontainers.py
alexandrevicenzi: I'm not sure, but the official Docker Hub image does not define a volume, perhaps it is best to not have it.

dcermak: It does not have it for good reasons. If you put `/srv/www/htdocs` into a volume then the following will yield very surprising behavior, namely that `/srv/www/htdocs/` will be **empty** in whatever you just built:
```Dockerfile
from registry.suse.com/suse/apache:2
copy ./static/ /srv/www/htdocs/
```

alexandrevicenzi:
We agreed that this container is not needed as we currently have the PHP-Apache for the LAMP stack.

PR #1083, opened 2024-05-06
Split redistributable from non-distributable base container

dirkmueller:
The EULA's are different, we should separate them accordingly.


dcermak (commented):
Why do we need this? bci-base is equal to sle15 at the moment.

dirkmueller:
> Why do we need this? bci-base is equal to sle15 at the moment.

yes, but I would like for them to not be equal anymore going forward, so that we can remove container-suseconnect from the redistributable image. It's just dead paperweight there.

PR #1088, opened 2024-05-07
Fix symlinks in gcc container

dirkmueller:


src/bci_build/package/gcc.py
```diff
@@ -46,6 +46,7 @@ def _is_main_gcc(os_version: OsVersion, gcc_version: _GCC_VERSIONS) -> bool:
                 (gcc_pkg := f"gcc{gcc_version}"),
                 (gpp := f"{gcc_pkg}-c++"),
                 "make",
+                "gawk",
```
dcermak, line 49: Not objecting, just asking, but this is unrelated to the `-s` flag that you added, right?

dcermak (approved):

src/bci_build/package/gcc.py
dirkmueller: yes. there was a second commit with the rationale added. "awk" is needed by ./configure.

Issue #1087, opened 2024-05-07
scratch-build-bot: render templates only for sp3 & sp4


dcermak:
Currently the containers for SP3 & SP4 are being created, but the CI doesn't render them.

We should add a CI job that only renders but doesn't build them on OBS

PR #1092, opened 2024-05-07
Render the templates for LTSS code streams but don't build them on OBS

dcermak:
This fixes https://github.com/SUSE/BCI-dockerfile-generator/issues/1087


dirkmueller:
LGTM

PR #1101, opened 2024-05-08
Remove Version identifier for the Tumbleweed BCI containers

dirkmueller:
These are ugly duplications of the distribution prefix already, Tumbleweeda as a rolling release has no version on its own.

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -155,7 +155,9 @@ class OsVersion(enum.Enum):
     @property
     def pretty_os_version_no_dash(self) -> str:
         if self.value == OsVersion.TUMBLEWEED.value:
-            return f"openSUSE {self.value}"
+            # TW has no version by itself and the "openSUSE Tumbleweed" is
+            # already part of the base identifier
+            return ""
         if self.value == OsVersion.BASALT.value:
             return "Adaptable Linux Platform"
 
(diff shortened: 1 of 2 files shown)


src/bci_build/package/__init__.py
```diff
@@ -155,7 +155,9 @@ def pretty_print(self) -> str:
     @property
     def pretty_os_version_no_dash(self) -> str:
         if self.value == OsVersion.TUMBLEWEED.value:
-            return f"openSUSE {self.value}"
+            # TW has no version by itself and the "openSUSE Tumbleweed" is
+            # already part of the base identifer
+            return f""
```
dcermak, line 160:
```suggestion
            return ""
```

src/bci_build/package/basecontainers.py
```diff
@@ -43,7 +43,7 @@ def _get_eula_package_names(os_version: OsVersion) -> tuple[str, ...]:
         package_name="micro-image",
         logo_url="https://opensource.suse.com/bci/SLE_BCI_logomark_green.svg",
         is_latest=os_version in CAN_BE_LATEST_OS_VERSION,
-        pretty_name=f"{os_version.pretty_os_version_no_dash} Micro",
+        pretty_name=f"{os_version.pretty_os_version_no_dash} Micro".strip(),
```
dcermak, line 46: Maybe add the `strip()` to `__post_init__` instead? Imho there's no danger in stripping trailing and leading whitespace.

dirkmueller: ah, good point, I was searching for a place where it is less ugly..

PR #1104, opened 2024-05-08
Fixup README sections

dirkmueller:
Start with a "Description" and then have a "Usage" section. Remove hardcoded distro references as they make the openSUSE images less confusing.


src/bci_build/package/minimal/README.md.j2
```diff
@@ -1,8 +1,10 @@
-# SLE BCI-Minimal: Base Container image without Zypper
+# {{ image.title }]: Base Container image without Zypper
```
dcermak, line 1: That one was tough to spot
```suggestion
# {{ image.title }}: Base Container image without Zypper
```

dmpop (approved):
LGTM! Thank you!

PR #1109, opened 2024-05-08
Fix link syntax in README

dirkmueller:


e4t:
Thanks for spotting this!

PR #1110, opened 2024-05-08
Fix lifecycle data for GCC, PHP and Python

dirkmueller:


src/bci_build/package/python.py
```diff
@@ -98,8 +110,7 @@ def _get_python_kwargs(
         **_get_python_kwargs("3.11", os_version),
         package_name="python-3.11-image",
         support_level=SupportLevel.L3,
-        # https://peps.python.org/pep-0664/ defines 2027/10/31, SUSE offers until end of the year
-        supported_until=datetime.date(2027, 12, 31),
+        supported_until=_PYTHON_SUPPORT_ENDS["3.11"],
```
dcermak, line 113: Would it make sense to move this into `_get_python_kwargs`? Only if it makes the code simpler

dirkmueller: no, I think the supportlevel is product specific, and you would just end up doing product comparisons in `_get_python_kwargs`. you can do this here simpler.

src/bci_build/package/gcc.py
```diff
@@ -41,6 +51,8 @@ def _is_main_gcc(os_version: OsVersion, gcc_version: _GCC_VERSIONS) -> bool:
         package_name=f"gcc-{gcc_version}-image",
         os_version=os_version,
         version=gcc_version,
+        support_level=SupportLevel.L3,
+        supported_until=_GCC_SUPPORT_ENDS.get(gcc_version, None),
```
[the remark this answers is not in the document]
dirkmueller: I'm a friend of explicit over implict, but I've changed it now.

src/bci_build/package/python.py
```diff
@@ -61,6 +72,10 @@ def _get_python_kwargs(
             Replacement(regex_in_build_description=pip3_replacement, package_name=pip3),
         ],
         "os_version": os_version,
+        "support_level": SupportLevel.L3,
+        "supported_until": _SLE_15_PYTHON_SUPPORT_ENDS[py3_ver]
+        if os_version in (OsVersion.SP4, OsVersion.SP5, OsVersion.SP6, OsVersion.SP7)
```
dcermak, line 77: This should be fine as we don't produce sp3 containers anyway?
```suggestion
        if os_version.is_sle15
```

dcermak (commented):
Rust & Node are missing the `os_version.lifecycle_data_pkg` in the package list. You added it to PHP but not to Node and Rust, is that intentional?

dirkmueller:
> Rust & Node are missing the `os_version.lifecycle_data_pkg` in the package list. You added it to PHP but not to Node and Rust, is that intentional?

Rust already had it before  and node doesn't need it because a) it's lifecycle is forcefully aligned to SP boundaries and b) the lifecycle data currently contains no information about node. We could still add it of course

src/bci_build/package/python.py
dirkmueller: Yes  we have several places to clean up should the is_sle15 check merge

src/bci_build/package/python.py
```diff
@@ -61,6 +72,10 @@ def _get_python_kwargs(
             Replacement(regex_in_build_description=pip3_replacement, package_name=pip3),
         ],
         "os_version": os_version,
+        "support_level": SupportLevel.L3,
+        "supported_until": _SLE_15_PYTHON_SUPPORT_ENDS[py3_ver]
+        if os_version.is_sle_15
```
dcermak, line 77:
```suggestion
        if os_version.is_sle15
```

src/bci_build/package/python.py
```diff
@@ -61,6 +72,10 @@ def _get_python_kwargs(
...
+        else None,
```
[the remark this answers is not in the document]
dcermak: Or just replace it with `_SLE_15_PYTHON_SUPPORT_ENDS.get(py3_ver)`, that should be equivalent as the supported until shouldn't matter on openSUSE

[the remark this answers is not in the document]
dirkmueller: we still have lifecycle on SLCC to define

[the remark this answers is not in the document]
dcermak: Makes sense!

src/bci_build/package/__init__.py
```diff
@@ -220,8 +232,9 @@ def lifecycle_data_pkg(self) -> list[str]:
 # End of General Support Dates
 _SUPPORTED_UNTIL_SLE: dict[OsVersion, datetime.date | None] = {
     OsVersion.SP4: datetime.date(2023, 12, 31),
-    OsVersion.SP5: None,  # datetime.date(2024, 12, 31),
+    OsVersion.SP5: datetime.date(2024, 12, 31),
     OsVersion.SP6: None,
+    OsVersion.SP7: datetime.date(2028, 7, 31),
```
dcermak, line 237: Wasn't this extended to 2030?

src/bci_build/package/gcc.py
```diff
@@ -42,6 +50,14 @@ def _is_main_gcc(os_version: OsVersion, gcc_version: _GCC_VERSIONS) -> bool:
         package_name=f"gcc-{gcc_version}-image",
         os_version=os_version,
         version=gcc_version,
+        support_level=(
+            SupportLevel.L3 if not os_version.is_sle15 else SupportLevel.TECHPREVIEW
+        ),
+        supported_until=(
+            _GCC_SLCC_SUPPORTED_UNTIL.get(gcc_version)
+            if not os_version.is_sle15
```
dcermak, line 58: Is the not intentional?
```suggestion
            if not os_version.is_sle15
```

src/bci_build/package/__init__.py
dirkmueller: Not yet public information

src/bci_build/package/gcc.py
dirkmueller: Yes, the lifecycle of gcc is different for sle15 (its 6 months after x+1.2 got released on sle15). As that is impossible to calculate upfront, we need to leave it as None for sle15.

The lifecycle above is the gcc slcc lifecycle which can be calculated upfront because it's keyed on the upstream release date

dirkmueller: That's why the dict has _SLCC_ in the name

dcermak (approved):

dcermak:
Builds were only waiting for s390x

PR #1124, opened 2024-05-14
Set rmt server stability tag

dirkmueller:
To avoid continuously changing the aggregate, we are setting a stability tag.


dcermak (commented):
This feels a bit wrong to be honest, I think we're using a hammer to fix something that needs a screwdriver (figuratively speaking). Especially the extra build tags appear pointless to me

dirkmueller:
> This feels a bit wrong to be honest, I think we're using a hammer to fix something that needs a screwdriver (figuratively speaking). Especially the extra build tags appear pointless to me

so we do need a versioned suffix of some form for the case when we have concurrent maintenance of two or more versions (rust, golang, python, openjdk and so on). I think we could so a "is_singleton" flag that removes the version suffix for buildname.

PR #1125, opened 2024-05-14
Add custom_build_name property

dcermak:
Alternative proposal to #1124

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -527,6 +527,10 @@ class BaseContainerImage(abc.ABC):
     #: present
     logo_url: str = ""
 
+    #: Optional custom build name that will override the default buildname, if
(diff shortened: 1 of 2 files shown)


src/bci_build/package/__init__.py
```diff
@@ -527,6 +527,10 @@ class BaseContainerImage(abc.ABC):
     #: present
     logo_url: str = ""
 
+    #: Optional custom build name that will override the default buildname, if
+    #: set
+    custom_build_name: str = ""
```
dirkmueller, line 532: I'm just afraid that it will run out of sync, and then it will be confusing or bite us. I think a boolean would be better.

dirkmueller: implemented the boolean variant in #1124

PR #1129, opened 2024-05-14
Drop Dotnet 7 container images as they're EOL (Fixes #983)

### LINKED ISSUE #983: Remove dotnet 7
Dotnet 7 is going end of life in may. We should remove it then

dirkmueller:


dcermak:
LGTM, but I'd like to wait a day or two, so that we'll release 7.0.19 before deleting everything.

PR #1179, opened 2024-05-23
Hardcode MariaDB versions

dcermak:
The OBS service replace_using_package_version cannot touch the README, so we have to resort to manually defining the version per service pack. To prevent a version drift, we add a sanity check that will fail the build if the versions do not match

### DIFF
--- a/src/bci_build/package/appcontainers.py
+++ b/src/bci_build/package/appcontainers.py
@@ -147,13 +147,18 @@ setpriv --reuid=$u --regid=$u --clear-groups -- /bin/bash "$@"
 MARIADB_CONTAINERS = []
 MARIADB_CLIENT_CONTAINERS = []
 
-for os_version in set(ALL_NONBASE_OS_VERSIONS) | {OsVersion.BASALT}:
+for mariadb_version, os_version in zip(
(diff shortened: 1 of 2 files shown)


src/bci_build/package/appcontainers.py
```diff
@@ -147,36 +147,28 @@
...
+
     MARIADB_CONTAINERS.append(
         ApplicationStackContainer(
             package_name=f"{prefix}mariadb-image",
             additional_names=additional_names,
             os_version=os_version,
             is_latest=os_version in CAN_BE_LATEST_OS_VERSION,
             name=f"{prefix}mariadb",
-            version="%%mariadb_version%%",
+            version=mariadb_version,
```
alexandrevicenzi, line 170: Possibly all containers use a variation of `%%{name}_version%%`, and I don't see this hardcoding scaling well.

It would be easier to remove `%%{name}_version%%` references from our READMES and just use `registry.suse.com/suse/{image}:latest`.

The README is just an example, it does not need to show the correct version IMHO, all containers should have a "latest version".

alexandrevicenzi (approved):

PR #1183, opened 2024-05-24
Fix typos in the licensing and eula template

alexandrevicenzi:


src/bci_build/package/templates/licensing_and_eula.j2
```diff
@@ -1,26 +1,29 @@
...
-This image is based on [SLE BCI](https://opensource.suse.com/bci/), a stable and redistributable foundation for software innovation. SLE BCI is enterprise-ready, and it comes with an option for support.
+{% if image.eula|string == 'sle-bci' %}
+This image is based on [SLE BCI](https://opensource.suse.com/bci/), a stable and redistributable foundation for software innovation.
+SLE BCI is enterprise-ready, and it comes with an option for support.
 
 See the [SLE BCI EULA](https://www.suse.com/licensing/eula/#bci) for further information.
-{% elif image.eula|string == 'sle' -%}
-This image is based on [SUSE Linux Enterprise Server](https://www.suse.com/products/server/), a reliable,
-secure, and scalable server operating system built to power mission-critical workloads in physical and virtual environments.
+{% elif image.eula|string == 'sle-eula' %}
```
dirkmueller, line 25: so this is the essential part of the change, while everything else seems wrapping lines and unwrapping them again? what is the story behind removing the white space stripping changes? I am not able to guess it from the commit message.

alexandrevicenzi: There were some typos in the text, which were fixed.

The EULA was wrong, which was fixed.

The space strip is unnecessary in some cases, I tried to fix the formating so the final output looks better. It should not affect anything AFAIK.

In theory, I can split everything into multiple commits.

dirkmueller: @alexandrevicenzi thanks for the explanation. for the future myself (remember I'm old and forget easily) I'd appreciate the rationale as part of the commit message because I can hardly guess what the reason was now, and I won't be able to guess it in a week from now when I look at this again. 

the space stripping was that way because I wanted it to be in one paragraph. your changes made it multiple paragraphs which changes the way it is being rendered. Not particularly caring either way, I wondered what the rationale for those changes were because it seemed inconsistent the way it is done now and I can't guess what logic you tried to set from the git commit message alone, that's why I'm asking back. 

I'd be fine it being in a single commit as long as the commit message explains why it has been done, but it obviously can be also multiple commits. 

I'm totally on board with the sle->sle-eula typo fix, and I'd +1 that right away, the rest seems more difficult to digest as a change (I am also fine with reordering the "this recipe and documentation" order, don't care either way).

alexandrevicenzi: I removed most spacing changes, take a look.
I also split the changes into two commits, with better commit messages.

dirkmueller (approved):
Thank you very much. :)

PR #1184, opened 2024-05-24
Fix Nginx image

alexandrevicenzi:


alexandrevicenzi:
This https://build.opensuse.org/request/show/1174158 is pending, it would reduce the size of the image as we don't need all binaries in Nginx.

The issue is that it may take a while to get the SR accepted, if ever, and currently, the Nginx is broken if you attempt to use templates.

dcermak:
@alexandrevicenzi Please rebase and let's hope the CI works :crossed_fingers:

dcermak (approved):

PR #1195, opened 2024-05-28
Add Python image README

alexandrevicenzi:
This is related to #993.

### DIFF
--- a/src/bci_build/package/python/README.md.j2
+++ b/src/bci_build/package/python/README.md.j2
@@ -1,0 +1,45 @@
+# {{ image.pretty_name }} container image
+
+{% include 'badges.j2' %}
+
+## Description
+
+[Python](https://www.python.org/) is an interpreted, interactive, object-oriented, open-source programming language. It incorporates modules, exceptions, dynamic typing, high-level dynamic data types, and classes. It provides interfaces to many system calls, libraries, and various window systems, and it is extensible in C or C++. It is also usable as an extension language for applications that require programmable interfaces.
+
+## Usage
+
+To deploy an application, install dependencies, copy the sources, and configure the application's main script:
+
+```Dockerfile
+FROM {{ image.pretty_reference }}
+
+WORKDIR /app
+
+COPY requirements.txt ./
+RUN pip install --no-cache-dir -r requirements.txt
+
+COPY . .
+
+CMD [ "python3", "./main-script.py" ]
+```
+
+Build and run the container image:
+
+```ShellSession
+$ podman build -t my-python-app .
+$ podman run -it --rm my-python-app
+```
+
+To run a single script inside a container, use the following command:
+
+```ShellSession
+$ podman run --rm -v "$PWD":/app:Z -w /app {{ image.pretty_reference }} python3 script.py
+```
+
+## Additional tools
+
+{% set version = image.version|replace(".", "") -%}
+
+The Python container image includes [pip](https://pip.pypa.io/){% if image.has_pipx %}, [pipx](https://pipx.pypa.io/){% endif %}{% if image.has_wheel %}, [wheel](https://wheel.readthedocs.io/){% endif %}, Python Development Headers, and Git.
+
+{% include 'licensing_and_eula.j2' %}


src/bci_build/package/python/README.md.j2
```diff
@@ -0,0 +1,63 @@
...
+To run a single script inside a container, use the following command:
+
+```ShellSession
+$ podman run --rm -v "$PWD":/app:Z -w /app {{ image.pretty_reference }} python3 script.py
+```
+
+To run the tests with [pytest](https://pytest.org) inside a container, use the following command:
+
+```ShellSession
+$ podman run --rm -v "$PWD":/app:Z -w /app {{ image.pretty_reference }} pytest -vv
```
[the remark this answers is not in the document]
alexandrevicenzi: yes, true, would be better if I replace it with unittest or add a note that pytest must be installed? or I can also drop this line entirely.

src/bci_build/package/python/README.md.j2
```diff
@@ -0,0 +1,63 @@
+# {{ image.pretty_name }} Container Image
+
+{% include 'badges.j2' %}
+
+## Description
+
+[Python](https://www.python.org/) is an interpreted, interactive, object-oriented, open-source programming language. It incorporates modules, exceptions, dynamic typing, very high-level dynamic data types, and classes. Python combines remarkable power with clear syntax. It has interfaces to many system calls and libraries, as well as to various window systems, and is extensible in C or C++. It is also usable as an extension language for applications that need a programmable interface.
```
dmpop, line 7:
```suggestion
[Python](https://www.python.org/) is an interpreted, interactive, object-oriented, open-source programming language. It incorporates modules, exceptions, dynamic typing, high-level dynamic data types, and classes. It provides interfaces to many system calls, libraries, and various window systems, and it is extensible in C or C++. It is also usable as an extension language for applications that require programmable interfaces.
```

src/bci_build/package/python/README.md.j2
```diff
@@ -0,0 +1,63 @@
...
+
+## Usage
+
+To deploy an application, install dependencies, copy the sources, and configure the application main script:
```
dmpop, line 11:
```suggestion
To deploy an application, install dependencies, copy the sources, and configure the application's main script:
```

src/bci_build/package/python/README.md.j2
```diff
@@ -0,0 +1,63 @@
...
+
+To run the tests with [pytest](https://pytest.org) inside a container, use the following command:
+
+```ShellSession
+$ podman run --rm -v "$PWD":/app:Z -w /app {{ image.pretty_reference }} pytest -vv
+```
+
+## Additional tools
+
+The following additional tools are included in the image:
```
dmpop, line 47:
```suggestion
The following tools are included in the image:
```

src/bci_build/package/python/README.md.j2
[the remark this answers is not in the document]
alexandrevicenzi: I added a note that pytest should be installed.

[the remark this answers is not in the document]
dcermak: I've had another thought about this and I think we should remove this direct pytest invocation. Imho this will only work reliably in the following scenarios:
1. you build a new container into which you `pip install pytest` and use the new container instead of bci/python
2. you run the usual `python -m venv .env3 && source .env3/bin/activate && pip install pytest` **but** with the project root bind mounted into the container so that the virtualenv on your machine is created using the container's python.

I find 1. to be a bad advertisement, 2. just far too brittle in practice and overall I have my doubts whether this is a common use case or whether most users wouldn't just run this as part of the container build process.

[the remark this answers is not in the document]
alexandrevicenzi: We can remove it, no problem, there's an example of how to call a single script, most people would know that can invoke tox, pytest, unitest, or else similarly.

src/bci_build/package/python/README.md.j2
```diff
@@ -0,0 +1,57 @@
...
+
+```ShellSession
+$ podman run --rm -v "$PWD":/app:Z -w /app {{ image.pretty_reference }} python3 script.py
+```
+
+## Additional tools
+
+The following tools are included in the image:
+
+{% set has_pipx = image.is_opensuse or (image.os_version.is_sle15 and image.version == '3.11') -%}
```
dirkmueller, line 43: TBH I don't prefer adding a set "has_pipx" if the condition is only used in one place later), but I understand this is a personal preference topic. 

However, I wonder if we can move the "has_pipx" into an attribute on the "image" somehow and set it from python.py so that we have the logic only in one place and not duplicated in the README? we're going to forget to update it in two places for sure.

src/bci_build/package/python/README.md.j2
```diff
@@ -0,0 +1,57 @@
...
+
+{% include 'badges.j2' %}
+
+## Description
+
+[Python](https://www.python.org/) is an interpreted, interactive, object-oriented, open-source programming language. It incorporates modules, exceptions, dynamic typing, high-level dynamic data types, and classes. It provides interfaces to many system calls, libraries, and various window systems, and it is extensible in C or C++. It is also usable as an extension language for applications that require programmable interfaces.
+
+## Usage
+
+To deploy an application, install dependencies, copy the sources, and configure the application's main script:
```
dirkmueller, line 11: to me "In order to deploy.." reads a bit nicer.

src/bci_build/package/python/README.md.j2
```diff
@@ -0,0 +1,57 @@
...
+
+```ShellSession
+$ podman build -t my-python-app .
+$ podman run -it --rm my-python-app
+```
+
+To run a single script inside a container, use the following command:
+
+```ShellSession
+$ podman run --rm -v "$PWD":/app:Z -w /app {{ image.pretty_reference }} python3 script.py
```
dirkmueller, line 36: should this be "python" rather than "python3" ? we're using python in line 23 and should be consistent on which one to pick (don't really care either way).

src/bci_build/package/python/README.md.j2
alexandrevicenzi: This is a long check, it looks better if we use `%set` IMHO.

To be honest, I had the same question, there's a piece in the code that does the same check, and now the same piece in the README, which can be forgotten one day.

The container is `DevelopmentContainer` and it does not make sense to add a Python-exclusive variable here, we could create a `PythonDevelopmentContainer` and set it there.

I could replace this part with the attribute `package_list`, but this might bring packages we don't want, but at least you change the package list in a single place (python code), which is possibly a better idea. We could create a `pretty_package_list` to be used in the README and perhaps clean any unwanted packages if needed.

@dcermak @dirkmueller any better ideas?

src/bci_build/package/python/README.md.j2
alexandrevicenzi: @dmpop WDYT?

src/bci_build/package/python/README.md.j2
alexandrevicenzi: Yes, we should have consistency.

src/bci_build/package/python/README.md.j2
dirkmueller: maybe

```
Deploying an application by installing dependencies, copying the sources and configuring the entrypoint:
```

would be another way to go imho.

src/bci_build/package/python/README.md.j2
dirkmueller: pretty_package_list sounds like a good idea (splitting package_list into those that would be listed in the readme by specifying them as pretty_package_list and any extra stuff as package_list)

alternatively simply store it as an extra kwargs attribute that we can query. that also works for this one case. I don't think we should invent too many specialized classes but that works also to store an attribute. it's pretty much the same idea.

dcermak: I'd first try if you can perform a `setattr(CONTAINER_INSTANCE, has_pipx, True/False)` or an equivalent of that (it might be that the `dataclass` will block you from doing that). The attribute will then be available in the template.

I'm not a huge fan of either another subclass nor of `pretty_package_list` as I don't see too many use cases for that atm.

src/bci_build/package/python/README.md.j2
dmpop: _To X, do Y_ is a standard neutral style. It's not about making things sound nicer or piling gerunds. It's making the text simple to read and instructions easy to follow.

Also, keep in mind that we have a style checker that ensures that we stick to the style defined in our style guide. It will flag your suggestions as an anomaly.

dirkmueller: I see, thanks for the explanation. Indeed, "vale" is flagging it as too wordy. alright. The suggestion that vale comes up with reads well for me as well: 

```
Deploying a python application involves installing dependencies, copying the source files, and configuring the entry point:
```

src/bci_build/package/python/README.md.j2
alexandrevicenzi: It is not possible to set attributes because the code where all the logic lives in `_get_python_kwargs`.
I can't add unknown `kwargs` in this function, and I don't have access to the `DevelopmentContainer` instance here. To add it to `DevelopmentContainer` I would need to duplicate code for each container instance, or create yet another function such as `_get_has_pipx`, which looks uglier than other options IMHO.

`pretty_package_list` does not work also, because `package_list` includes the `python311` package as well, which I have no way of removing because there's no variable to match.

No solution seems to work good enough.

For `pretty_package_list` I can remove any packages from `lifecycle_data_pkg` or any packages that are not `PackageType.IMAGE`. To remove the main packages such as `python311` or `python311-devel` I would need to have these in an attribute somewhere else that is not `package_list`.

An option is to create an attribute named `additional_package_list`, which is used by the README, but could also be used by the generator. For example, `package_list` can take the required packages, while `additional_package_list` takes the packages that add extra functionality, such as: make, git, pipx, or else.

The golang and tomcat images also have extra packages in the README. The PHP image has also extra packages that are not mandatory, such as composer. In theory, we could make use of `pretty_package_list` or `additional_package_list` in other images.

src/bci_build/package/python/README.md.j2
alexandrevicenzi: It seems there's no symlink for `python`, the proper way is `python3`.

src/bci_build/package/python/README.md.j2
alexandrevicenzi: @dirkmueller @dcermak WDYT? All options have drawbacks.

dcermak: If we want to start listing packages in the READMEs, then `pretty_package_list` is probably the best approach. Do you want to create a PR for that @alexandrevicenzi?

src/bci_build/package/python/README.md.j2
dirkmueller: then line 23 should be "python3" as well?

src/bci_build/package/python/README.md.j2
alexandrevicenzi: @dcermak `pretty_package_list` does not work unless we add an `additional_package_list` or change the `Package` class, this seems a massive change to how packages are defined. Or we list all packages, but the rule so far has been to list extra packages only, which are often useful to add a bit more functionality to a container. The user does not need to know all packages in the image, but it is useful to know that make, git, or pipx is available.

I'm more inclined to `PythonDevelopmentContainer`, which is not disruptive to other containers, and given that we have a `python.py` file where we define the Python container image it would not be that weird to extend `DevelopmentContainer`.

dirkmueller: I think the suggestion is to go with "pretty_package_list" and "additional_package_list" . this way the readme's can tag them accordingly. we just need to change a few places to include both package lists.

alexandrevicenzi: I have added #1267 and #1268 as a potential solution, we should pick one. Both proposals should have little side effects, one proposal is a lot simpler than the other.

dirkmueller (changes requested):
Please adjust after #1267  being merged.

src/bci_build/package/python.py
```diff
@@ -48,7 +48,7 @@ def _get_python_kwargs(py3_ver: _PYTHON_VERSIONS, os_version: OsVersion):
 
     kwargs = {
         "name": "python",
-        "pretty_name": f"Python {py3_ver} development",
+        "pretty_name": f"Python {py3_ver} Development",
```
dirkmueller, line 51: no we use sentence style capitalization here. see https://documentation.suse.com/style/current/single-html/docu_styleguide/#sec-capitalization-sentence

dirkmueller: we're a bit inconsistent, filed https://github.com/SUSE/BCI-dockerfile-generator/pull/1272 as a followup

alexandrevicenzi: Yes, but many of the READMEs are "Name Container Image".

We should change all READMEs in that case to be lowercase.
Otherwise, without this change, the container title will be "Python development Container Image", which is odd.

alexandrevicenzi: Fixed to be all sentence style.

src/bci_build/package/python/README.md.j2
alexandrevicenzi: Since #1267 was merged, this has been fixed and rebased.

dirkmueller (approved):

dcermak (changes requested):
Sorry, the package version is incorrect :-(

src/bci_build/package/python/README.md.j2
```diff
@@ -0,0 +1,56 @@
...
+
+{% set version = image.version|replace(".", "") -%}
+
+- python{{ version }}-devel
+- python{{ version }}-pip
+{% if image.has_pipx -%}
+- python{{ version }}-pipx
+{% endif -%}
+{% if image.has_wheel -%}
+- python{{ version }}-wheel
```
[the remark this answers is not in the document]
alexandrevicenzi: is it worth listing the `python3-devel` package here? otherwise I need to add a check `if is_python_system` because it will be `python3-devel` for the system version, not for 3.11 on SLE.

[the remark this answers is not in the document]
dcermak: It is not worth listing `python3-devel` imho, but the package names are incorrect for pip & wheel too. It's `python3-pip` for python 3.6 and not `python36-pip`

src/bci_build/package/python/README.md.j2
```diff
@@ -0,0 +1,56 @@
...
+{% endif -%}
+- curl
```
dirkmueller, line 53: curl is part of the base image already.

lvicoun (approved):
LGTM!

src/bci_build/package/python/README.md.j2
```diff
@@ -0,0 +1,56 @@
...
+
+To run a single script inside a container, use the following command:
+
+```ShellSession
+$ podman run --rm -v "$PWD":/app:Z -w /app {{ image.pretty_reference }} python3 script.py
+```
+
+## Additional tools
+
+The following tools are included in the image:
```
dirkmueller, line 41: maybe this shouldn't say "following tools" and then a list of package names. I think we could simply have two paragraphs that explain what "pip" is (if pip is available, which is always the case) and what pipx is (if pipx is available in the image).

alexandrevicenzi: Instead of package name, I was considering having the tool name (e.g not python3-pipx, just pipx) and adding a small description.

- pipx — Install and Run Python Applications in Isolated Environments

@dirkmueller @dcermak WDYT? this at least fix the problem with including the package version

alexandrevicenzi: one thing we discussed is the usefulness of this section, because of this section this PR is stuck. Perhaps removing this completely and replacing it with a # pipx and # else section is better.

dcermak: I think we can just mention that pip (& pipx) is/are installed, or just dropping it entirely

src/bci_build/package/python/README.md.j2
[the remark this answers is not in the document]
alexandrevicenzi: I reworked this section, no version number in names anymore.

src/bci_build/package/python/README.md.j2
alexandrevicenzi: I kept this section, but I changed it to include tool names, it seems more useful.

PR #1203, opened 2024-05-30
Update registry example

frelon:
The config file used by the registry container is `/etc/registry/config.yml`, as shown here:

```
$ docker inspect --format='{{.Config.Cmd}}' registry.suse.com/suse/registry:latest
[serve /etc/registry/config.yml]
```


dcermak (approved):
Thanks for noticing & fixing this!

PR #1210, opened 2024-06-03
Add module to fetch package versions from OBS

dcermak:
This allows us to have more or less automatically updated versions in READMEs


src/bci_build/package/versions.py
```diff
@@ -0,0 +1,151 @@
...
+The package versions are available via the function :py:func:`get_pkg_version`.
+
+The json file is updated via the function :py:func:`run_version_update`. It can
+also be called directly via :command:`poetry run update-versions` and is also
+run in the CI via :file:`.github/workflows/update-versions.yml`. Note that
+:py:func:`run_version_update` reads your OBS credentials from the environment
+variables ``OSC_USER`` and ``OSC_PASSWORD``. The function fails if one of them
+is not present.
+
+To add a new package to the versions json, add it's package name and all code
```
alexandrevicenzi, line 32: Could the script create this automatically? The code knows all containers and all OS versions.

pyproject.toml
```diff
@@ -10,17 +10,19 @@ packages = [
     { include = "dotnet", from = "src" },
 ]
 [tool.poetry.dependencies]
-python = ">=3.10,<4.0"
+python = "^3.11"
```
dirkmueller, line 13: is it intentional that we drop 3.10 here?

source/adding_containers.rst
```diff
@@ -123,6 +123,15 @@ seen in the above code block, where we also set the
 attribute via this mechanism.
 
 
+.. Caution::
+
+   The current setup **cannot** replace versions in READMEs. If you rely on
```
dirkmueller, line 128: we only have this limitation for tumbleweed/opensuse containers? the others should be working fine?

it feels like yet another cumbersome thing to maintain going forward.

pyproject.toml
dcermak: yes, `py_obs` depends on `python = "^3.11"`.

source/adding_containers.rst
dcermak: > we only have this limitation for tumbleweed/opensuse containers? the others should be working fine?

This is an issue for **every** container. The replace_using_package_version service replaces the variables in the README **in the buildroot**. But the README that we "ship" is not the one from inside the container, but the one in the source package. That one is not modified by replace_using_package_version (and also cannot be).

> it feels like yet another cumbersome thing to maintain going forward. did anyone report this as an issue so far?

I talked about this last week with darix and there's really no easy solution that we came up with.

src/bci_build/package/versions.py
alexandrevicenzi: can't you just loop over `ALL_CONTAINER_IMAGE_NAMES` and print the name/version with an empty version?

dcermak: That creates a circular import

alexandrevicenzi: ok, in that case, it makes sense to manually add just what you need.

alexandrevicenzi (approved):

src/bci_build/package/spack.py
```diff
@@ -19,26 +21,15 @@
         os_version=os_version,
         is_latest=os_version in CAN_BE_LATEST_OS_VERSION,
         logo_url="https://spack.io/assets/images/spack-logo-white.svg",
-        version="%%spack_version%%",
+        version=(spack_pkg_version := get_pkg_version("spack", os_version)),
         version_in_uid=False,
         additional_versions=["%%spack_minor%%"],
         replacements_via_service=[
-            Replacement(
-                regex_in_build_description="%%spack_version%%",
```
dirkmueller, line 27: why are we removing that here? this is unrelated to the README stuff ?

dcermak: The replacement isn't used anywhere else, so I dropped it to not run the service needlessly

dirkmueller: but're still having https://github.com/SUSE/BCI-dockerfile-generator/commit/488575b4d70e43991df3f7e033aa4547ec1c157d#diff-e769e1c8493dbb658288f74a0b9aa816607c37dc429a6f7d5651ea02f56cbb6aR19 ? so %%spack_minor%% needs to be changed as well?

src/bci_build/package/helpers.py
```diff
@@ -0,0 +1,18 @@
...
+    package with the name ``pkg_name`` is not at the provided version
+    ``pkg_version``. The optional parameter ``parse_version`` allows you to
+    restrict the version check to only match the major, major+minor, or
+    major+minor+patch version.
+
+    """
+    cut_count: dict[PARSE_VERSION_T, int] = {"major": 1, "minor": 2, "patch": 3}
+
+    return f"""# sanity check that the version from the tag is equal to the version of {pkg_name} that we expect
+{DOCKERFILE_RUN} [[ $(rpm -q --qf "%{{version}}" {pkg_name} | cut -d "." -f -{cut_count[parse_version]}) = "{pkg_version}" ]]"""
```
dirkmueller, line 18: avoid unnecessary bashism:

```suggestion
{DOCKERFILE_RUN} [ "$(rpm -q --qf "%{{version}}" {pkg_name} | cut -d "." -f -{cut_count[parse_version]})" = "{pkg_version}" ]"""
```

src/bci_build/package/spack.py
dirkmueller: ok I see this doesn't have to be replaced strictly speaking.

dirkmueller (changes requested):
```diff

--- a/src/bci_build/package/spack/README.md.j2
+++ b/src/bci_build/package/spack/README.md.j2
@@ -1,4 +1,4 @@
-# Spack %%spack_version%% Container Image
+# Spack {{ image.version }} Container Image
 {% include 'badges.j2' %}
 
 ## Description
``` 

is missing

dirkmueller (approved):

PR #1220, opened 2024-06-05
Generate the bci-base container via dockerfile generator as well

dirkmueller:


dirkmueller:
(this needs a bit more work)

src/bci_build/package/basecontainers.py
```diff
@@ -35,6 +35,145 @@ def _get_eula_package_names(os_version: OsVersion) -> tuple[str, ...]:
     return ("skelcd-EULA-bci",)
 
 
+def _get_base_config_sh_script(os_version: OsVersion) -> str:
+    return textwrap.dedent(
+        r"""
```
dcermak, line 40: The dedent does nothing
```suggestion
    return r"""
```

src/bci_build/package/basecontainers.py
```diff
@@ -35,6 +35,145 @@ def _get_eula_package_names(os_version: OsVersion) -> tuple[str, ...]:
...
+rm -f /var/lib/zypp/AnonymousUniqueId
+
+# Remove various log files. Although possible to just rm -rf /var/log/*, that
+# would also remove some package owned directories (not %ghost) and some files
+# are actually wanted, like lastlog in the !docker case.
+# For those wondering about YaST2 here: Kiwi writes /etc/hosts, so the version
+# from the netcfg package ends up as /etc/hosts.rpmnew, which zypper writes a
+# letter about to /var/log/YaST2/config_diff_2022_03_06.log. Kiwi fixes this,
+# but the log file remains.
+rm -rf /var/log/{zypper.log,zypp/history,YaST2}
```
dcermak, line 98: This could be replaced with our `LOG_CLEAN` constant

src/bci_build/package/__init__.py
```diff
@@ -930,6 +933,11 @@ def pkg_filter_func(p: str | Package) -> bool:
 
             return pkg_filter_func
 
+        def pkg_listing_func(pkg: Package) -> str:
+            if pkg.replaces:
+                return f'<package name="{pkg}" replaces="rpm"/>'
```
dcermak, line 938: The `replaces=` attribute has been deleted in schemaversion 6.7:
```
    Add XSL stylesheet to convert from v6.6 to v6.7
    
    Delete the following obsolete and unhandled attributes and sections
    
    * <image ... kiwirevision="..."/>
    * <type ... checkprebuilt="..."/>
    * <type ... fsnocheck="..."/>
    * <repository ... status="..."/>
    * <repository ... prefer-license="..."/>
    * <package replaces="..."/>
    * <defaultprebuilt>
    * <defaultdestination>
    * <defaultroot>
    * <partitioner>
    * <rpm-force>
```

src/bci_build/package/basecontainers.py
dirkmueller: I would like to first produce a 100% identical result with the handcrafted config.sh. we can clean this up then in a followup step.

src/bci_build/package/__init__.py
dirkmueller: This has been fixed now.

src/bci_build/package/basecontainers.py
dirkmueller: replaced with jinja.template rendering.

Vogtinator:
Looking at the diff, some issues:

- [ ] Duplicate `suseImportBuildKey`
- [ ] Duplicate `/var/log` cleaning. The manual one has better comments.
- [ ] Duplicate `Remove locale files`
- [ ] `<label name="com.suse.image-type" value="sle-bci"/>` got lost

dirkmueller:
> Looking at the diff, some issues:
> * [ ]  `<label name="com.suse.image-type" value="sle-bci"/>` got lost

This is intentional, we have deprecated that label and removed it from all BCI containers.

Vogtinator:
In the case that we might get rid of findutils in the container again, we'd have to switch back to the globstar based locale deletion

dirkmueller:
>  Duplicate /var/log cleaning. The manual one has better comments.

I can add the better comments to the sources in dockerfile generator?

> Duplicate Remove locale files

removed.

dirkmueller:
> In the case that we might get rid of findutils in the container again, we'd have to switch back to the globstar based locale deletion

removed find(1) usage again.

src/bci_build/package/__init__.py
```diff
@@ -31,7 +31,11 @@
 
 #: Remove various log files. While it is possible to just ``rm -rf /var/log/*``,
 #: that would also remove some package owned directories (not %ghost)
-LOG_CLEAN: str = "rm -rf /var/log/{lastlog,tallylog,zypper.log,zypp/history,YaST2}"
+LOG_CLEAN: str = """
+# Remove various log files. While it's possible to just rm -rf /var/log/*
+# would also remove some package owned directories (not %ghost)
+# and some files are actually wanted
+rm -rf /var/log/{lastlog,tallylog,zypper.log,zypp/history,YaST2}"""
```
dcermak, line 38: You cannot do this, `LOG_CLEAN` must be a single line instruction, because it is used in all `Dockerfile` templates inside of `RUN` instructions.

dirkmueller: restored

src/bci_build/package/base/README.md.j2
```diff
@@ -0,0 +1,30 @@
...
+standards are applied consistently and continuously improves security-related
+capabilities.
+
+SLE BCI is lightweight and easy to adopt, with the ability to run with any
+Linux OS. Avoid lock-in imposed by other vendors and get exactly what you need,
+fast. SLE BCI delivers a flexible developer experience that accounts for,
+integrates with, and supports language-native tools and workflows.
+
+## Usage
+
```
dcermak, line 29: Leftover?
```suggestion
```

src/bci_build/templates.py
```diff
@@ -91,6 +91,8 @@
 <!-- {{ INFOHEADER }}-->
 <!-- OBS-AddTag: {% for tag in image.build_tags -%} {{ tag }} {% endfor -%}-->
 <!-- OBS-Imagerepo: obsrepositories:/ -->
+{%- if image.os_version.is_sle15 and image.name == "base" %}
+<!-- OBS-IgnorePackage: rpm -->{% endif %}
```
dcermak, line 95: Do we need this? I don't see it in [sles15-image.kiwi](https://build.suse.de/projects/SUSE:SLE-15-SP6:Update/packages/sles15-image/files/sles15-image.kiwi?expand=1) but there's this instead:
```xml
<!-- OBS-ExcludeArch: i586 -->
```

Vogtinator: Previous kiwi schemaversions allow `<package ... replaces="foobar"/>` but that's no longer part of the schema, so it has to be done a different way

Vogtinator: The `<!-- OBS-ExcludeArch: i586 -->` is only needed if this is part of the SLE project itself which provides the i586 arch as well.

src/bci_build/package/base/README.md.j2
```diff
@@ -0,0 +1,30 @@
...
+{% include 'badges.j2' %}
+
+## Description
+
+SUSE Linux Enterprise Base Container Images (SLE BCI) provide truly open,
+flexible, and secure container images and application development tools. The
+images consist of container environments based on SUSE Linux Enterprise and
+designed to be a secure base for any containerized workload.
+
+SLE BCI is freely available, re-distributable, and supported across many
```
alexandrevicenzi, line 11: nit: you could remove _many_ from _across many different environments_

src/bci_build/package/base/README.md.j2
```diff
@@ -0,0 +1,30 @@
...
+different environments. These templates and tools address modern, containerized
```
alexandrevicenzi, line 12: which templates and tools? it is not clear here.

src/bci_build/package/base/README.md.j2
```diff
@@ -0,0 +1,30 @@
...
+application development and CI/CD application containerization.  They can be
```
alexandrevicenzi, line 13: what can be used? BCI? the templates and tools from the previous sentence? which is also not clear.

src/bci_build/package/base/README.md.j2
```diff
@@ -0,0 +1,30 @@
...
+used immediately by developers and integrators without the “lock-in” imposed by
+other offerings.
+
+SLE BCI inherits industry-leading security and compliance from SUSE Linux
```
alexandrevicenzi, line 17: nit: it would be nice to mention what security and compliance are being supported here, as we know what we support.

src/bci_build/package/base/README.md.j2
```diff
@@ -0,0 +1,30 @@
...
+within your container build process. The container images are designed to be a
+secure base for any application workload. SUSE ensures that compliance
+standards are applied consistently and continuously improves security-related
+capabilities.
+
+SLE BCI is lightweight and easy to adopt, with the ability to run with any
+Linux OS. Avoid lock-in imposed by other vendors and get exactly what you need,
```
alexandrevicenzi, line 24: nit: it would be nice to make it clear that Linux OS here is the host OS of the container runtime, otherwise it gives an impression that you can run BCI with another OS inside.

src/bci_build/templates.py
dcermak: > Previous kiwi schemaversions allow `<package ... replaces="foobar"/>` but that's no longer part of the schema, so it has to be done a different way

I see, thanks!

> The `<!-- OBS-ExcludeArch: i586 -->` is only needed if this is part of the SLE project itself which provides the i586 arch as well.

Don't we want to publish the container from the SLE project though? Then we'd need this additional comment

src/bci_build/package/base/README.md.j2
dirkmueller: I'd like to address this in a followup. this one is intended to as closely as possible resemble the existing sle15-image. we can clean it up once this PR has been merged. it is large enough as is.

src/bci_build/package/base/README.md.j2
dirkmueller: I'd like to address this in a followup. this one is intended to as closely as possible resemble the existing sle15-image. we can clean it up once this PR has been merged. it is large enough as is.

src/bci_build/package/base/README.md.j2
dirkmueller: I'd like to address this in a followup. this one is intended to as closely as possible resemble the existing sle15-image. we can clean it up once this PR has been merged. it is large enough as is.

src/bci_build/package/base/README.md.j2
dirkmueller: I'd like to address this in a followup. this one is intended to as closely as possible resemble the existing sle15-image. we can clean it up once this PR has been merged. it is large enough as is.

alexandrevicenzi: ok, fair.

alexandrevicenzi (dismissed):

Vogtinator:
I'm looking at the resulting diff of sles15-image and I found the following:

```diff
 #======================================
-# Exclude docs intallation
+# Exclude docs installation
 #--------------------------------------
-sed -i 's/.*rpm.install.excludedocs.*/# rpm.install.excludedocs = yes/g' /etc/zypp/zypp.conf
+sed -i 's/.*rpm.install.excludedocs.*/rpm.install.excludedocs = yes/g' /etc/zypp/zypp.conf
```

i.e. previously excludedocs was not actually set to yes, ouch. IMO ok to fix, but might need discussion and documentation. This hits derived containers only.

Vogtinator:
```diff
-    <version>15.0.0</version>
+    <version>15.5.0</version>
```

will reset the cicount, meaning tags would start from 15.5.1.0 again.

Vogtinator:
```
+<!-- OBS-AddTag: suse/sle15:%OS_VERSION_ID_SP% suse/sle15:%OS_VERSION_ID_SP%.%RELEASE% bci/bci-base:%OS_VERSION_ID_SP% bci/bci-base:%OS_VERSION_ID_SP%.%RELEASE% -->
```

but also

```
+      <containerconfig name="suse/sle15" tag="%OS_VERSION_ID_SP%" maintainer="SUSE LLC (https://www.suse.com/)" additionaltags="%OS_VERSION_ID_SP%,%OS_VERSION_ID_SP%.%RELEASE%">
```

i.e. the `suse/sle15` tags are set twice.

Vogtinator:
`<package name="patterns-base-minimal_base"/>` is new in bootstrap packages. Intentional?

dirkmueller:
> i.e. the `suse/sle15` tags are set twice.

yes, luckily OBS de-duplicates them, so it's not a real issue.

dirkmueller:
> `<package name="patterns-base-minimal_base"/>` is new in bootstrap packages. Intentional?

not new, there since

```
------------------------------------------------------------------- 
Mon Oct 23 13:34:14 UTC 2023 - Dirk Müller <dmueller@suse.com>      
                                                                    
- add patterns-base-minimal_base to have the SLE branding        
```

Vogtinator:
> > i.e. the `suse/sle15` tags are set twice.
> 
> yes, luckily OBS de-duplicates them, so it's not a real issue.

Can we avoid the duplication? Could be as simple as removing `additionaltags` from the template. FWICT there's a bug anyway:

```xml
      <containerconfig
          name="{{ image.build_tags[0].split(':')[0] }}"
          tag="{{ image.build_tags[0].split(':')[1] }}"
          maintainer="{{ image.maintainer }}"{% if image.kiwi_additional_tags %}
{%- if image.entrypoint_user  %}
          user="{{ image.entrypoint_user }}"
{%- endif %}
          additionaltags="{{ image.kiwi_additional_tags }}"{% endif %}>
        <labels>
```

If `image.kiwi_additional_tags` is empty, the user is not set either.

dirkmueller:
> ```diff
> -    <version>15.0.0</version>
> +    <version>15.5.0</version>
> ```
> 
> will reset the cicount, meaning tags would start from 15.5.1.0 again.

yes, we can set that via buildrelease though. for now I've restricted it to sp6 and set a buildrelease matching that.

Vogtinator:
> > `<package name="patterns-base-minimal_base"/>` is new in bootstrap packages. Intentional?
> 
> not new, there since
> 
> ```
> ------------------------------------------------------------------- 
> Mon Oct 23 13:34:14 UTC 2023 - Dirk Müller <dmueller@suse.com>      
>                                                                     
> - add patterns-base-minimal_base to have the SLE branding        
> ```

Not there in SUSE:SLE-15-SP5:Update/sles15-image yet

dirkmueller:
> If `image.kiwi_additional_tags` is empty, the user is not set either.

good catch, moved to https://github.com/SUSE/BCI-dockerfile-generator/pull/1419 

in regards to reducing the duplication, I'll take a look later.

src/bci_build/templates.py
dcermak, line 95: Can we please not have hacks like this in the template? Image specific settings belong into the image and not into the general purpose template imho

src/bci_build/package/__init__.py
```diff
@@ -198,14 +198,14 @@ def is_sle15(self) -> bool:
...
-        return self.value == OsVersion.TUMBLEWEED.value
-
     @property
     def is_ltss(self) -> bool:
         return self in ALL_OS_LTSS_VERSIONS
 
+    @property
+    def is_tumbleweed(self) -> bool:
+        return self.value == OsVersion.TUMBLEWEED.value
+
```
dcermak, line 208: Rebase artifact

dirkmueller:
> > > `<package name="patterns-base-minimal_base"/>` is new in bootstrap packages. Intentional?
> Not there in SUSE:SLE-15-SP5:Update/sles15-image yet

correct, but we'll only do the SP6 image for now. but I've added a condition for it it on SP6

src/bci_build/templates.py
dirkmueller: added a separate kiwi_ignore_packages flag for this.

src/bci_build/package/__init__.py
dirkmueller: yeah.. removed.

dirkmueller:
> Can we avoid the duplication? Could be as simple as removing `additionaltags` from the template.

I'd rather go the other way and remove OBS specific hacks so that builds outside obs work better. moved that to #1423 (which is not working atm because of some bug somewhere)

for now, this is a cosmetical  problem, I would ignore it in order to not block this change from going in further.

Vogtinator:
There's duplication even in the kiwi tag:

```
+          tag="%OS_VERSION_ID_SP%"
+          additionaltags="%OS_VERSION_ID_SP%,%OS_VERSION_ID_SP%.%RELEASE%"
```

i.e. `suse/sle15:%OS_VERSION_ID_SP%` is defined three times now :-/

Vogtinator:
Looks good otherwise.

> I'd rather go the other way and remove OBS specific hacks so that builds outside obs work better. moved that to https://github.com/SUSE/BCI-dockerfile-generator/pull/1423 (which is not working atm because of some bug somewhere)

Where: OBS, kiwi, dockerfile-generator?

src/bci_build/package/base/README.md.j2
Vogtinator: Looks like it's back?

Having an empty section looks really ugly

dirkmueller:
> There's duplication even in the kiwi tag:

> i.e. `suse/sle15:%OS_VERSION_ID_SP%` is defined three times now :-/

Yes, but it doesn't matter. Yet, I fixed it now.

alexandrevicenzi (approved):

dcermak (approved):
LGTM, but I'd like @Vogtinator to give his ACK here

dirkmueller:
> > I'd rather go the other way and remove OBS specific hacks so that builds outside obs work better. moved that to #1423 (which is not working atm because of some bug somewhere)
> Where: OBS, kiwi, dockerfile-generator?

I believe a bug in kiwi, but I haven't checked it closely (could be elsewhere). I can see the skopeo call to add the additional names, but they are still not there on the build result. maybe also OBS interaction. needs to be debugged tehre.

Vogtinator (approved):
Looking at the diff between SUSE:SLE-15-SP6:Update and this I don't see anything obvious wrong anymore, just mostly cosmetic issues.

PR #1223, opened 2024-06-05
Fips base fixes

dirkmueller:


src/bci_build/package/templates/badges.j2
```diff
@@ -1,6 +1,12 @@
...
 {%- endif %}
-{% if not image.is_opensuse and image.support_level | string != 'l3' -%}
+{%- if not image.is_opensuse and image.support_level | string != 'l3' -%}
 ![Support Level](https://img.shields.io/badge/Support_Level-{{ image.support_level }}-blue)
+{%- endif %}
+{%- if not image.is_opensuse and image.build_tags[0].find('/ltss/') != -1 -%}
+![Access Protected](https://img.shields.io/badge/Requires_login_for_access-orange)
+{%- endif %}
+{%- if not image.is_opensuse and image.build_tags[0].find('/ltss/') != -1 -%}
+![Long Term Service Pack Support](https://img.shields.io/badge/LTSS-Yes-orange)
```
dcermak, line 11: Duplicate if?
```suggestion
{%- if not image.is_opensuse and image.build_tags[0].find('/ltss/') != -1 -%}
![Access Protected](https://img.shields.io/badge/Requires_login_for_access-orange)
![Long Term Service Pack Support](https://img.shields.io/badge/LTSS-Yes-orange)
```

Also, this looks to me like it could be worth adding a property `is_ltss` to either the image or the `OsVersion`. WDYT?

dirkmueller: I intentionally duplicated the if because eventually we'll have to have a different if condition for "access protected" and "ltss" since we'll introduce more access protected images going forward.

src/bci_build/package/templates/badges.j2
```diff
@@ -1,6 +1,10 @@
 {% if image.eula == 'sle-bci' -%}
 ![Redistributable](https://img.shields.io/badge/Redistributable-Yes-green)
 {%- endif %}
-{% if not image.is_opensuse and image.support_level | string != 'l3' -%}
+{%- if not image.is_opensuse and image.support_level | string != 'l3' -%}
```
dcermak, line 4: I know that this is unrelated, but don't we want to highlight l3 support on these images?

PR #1241, opened 2024-06-07
Add NodeJS image README

alexandrevicenzi:
This is related to #993.

### DIFF
--- a/src/bci_build/package/nodejs/README.md.j2
+++ b/src/bci_build/package/nodejs/README.md.j2
@@ -1,0 +1,48 @@
+# {{ image.pretty_name }} container image
+
+{% include 'badges.j2' %}
+
+## Description
+
+[Node.js](https://nodejs.org/) is a free, open-source, cross-platform JavaScript run-time environment that lets developers write server-side applications and tools outside of a browser.
+
+## Usage
+
+To deploy an application, install dependencies, copy the sources, and configure the application's main script:
+
+```Dockerfile
+FROM {{ image.pretty_reference }}
+
+WORKDIR /app
+
+COPY package.json package-lock.json ./
+RUN npm install
+
+COPY . .
+
+EXPOSE 3000
+
+CMD [ "node", "./server.js" ]
+```
+
+Build and run the container image:
+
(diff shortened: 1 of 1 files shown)


src/bci_build/package/nodejs/README.md.j2
```diff
@@ -0,0 +1,54 @@
...
+```ShellSession
+$ podman run --rm -v "$PWD":/app:Z -w /app {{ image.pretty_reference }} node script.js
+```
+
+## Additional tools
+
+The following additional tools are included in the image:
+
+- git-core
+
```
dcermak, line 53: I find this section is of questionable usefulness
```suggestion
```

dcermak (changes requested):
Can you also please add a section about how to obtain `yarn`? (you have to install it via `npm install -g yarn`)

alexandrevicenzi:
> Can you also please add a section about how to obtain `yarn`? (you have to install it via `npm install -g yarn`)

I left it out on purpose.

Why yarn, and why not also pnpm or bun?
If I add one, I feel the need to add everything else.
Indeed, yarn is more popular than pnpm or bun, but npm is the default.

In the Python image we don't mention how to use/install Poetry or Conda.

src/bci_build/package/nodejs/README.md.j2
alexandrevicenzi: Yes, I had a similar feeling.

dirkmueller: imho we should only document it if we have a test that enforces git-core being available. 

I think generally it is a good idea to list the packages that are core and auxiliary installed but that requires a bit more plumbing in the dockerfile generator to not having this listed manually in the readmes.

alexandrevicenzi: I removed this section.

dcermak (approved):
> Why yarn, and why not also pnpm or bun?

Fair enough, the dockerhub image also doesn't mention it, so let's skip it.

dirkmueller (approved):

PR #1249, opened 2024-06-11
Add lifecycle data for the dotnet containers

dirkmueller:


src/dotnet/updater.py
```diff
@@ -176,6 +186,13 @@ def __post_init__(self):
 
         ver = version.parse(str(self.version))
 
+        # Set the lifecycle information taken from
+        # https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core
+        self.supported_until = {
+            "6.0": datetime.date(2024, 11, 12),
+            "8.0": datetime.date(2026, 11, 10),
+        }.get(str(self.version))
```
dcermak, line 194: WDYT about making this blow up if we add a new version without an EoL?
```suggestion
        self.supported_until = {
            "6.0": datetime.date(2024, 11, 12),
            "8.0": datetime.date(2026, 11, 10),
        }[str(self.version)]
```

dirkmueller: I'm not sure it's worth it. it's just a minor thing. don't really care either way.

dirkmueller: I've added an assert.

PR #1251, opened 2024-06-11
Add a SLSA (v0.1) badge

dirkmueller:


src/bci_build/package/templates/badges.j2
```diff
@@ -8,3 +8,7 @@
 ![Access Protected](https://img.shields.io/badge/Requires_login_for_access-orange)
 ![Long Term Service Pack Support](https://img.shields.io/badge/LTSS-Yes-orange)
 {%- endif %}
+{%- if not image.is_opensuse %}
+[<img src="https://img.shields.io/badge/SLSA_(v0.9)-Level_4-Green">](https://documentation.suse.com/sbp/server-linux/html/SBP-SLSA4/)
+[<img src="https://img.shields.io/badge/Provenance-Available-Green">](https://documentation.suse.com/container/all/html/Container-guide/index.html#container-verify)
```
dcermak, line 13: This looks odd, is this a HTML tag or a markdown directive?

dirkmueller: It's a link as a markdown directive in the format [linktext](link) with the linktext being a html tag called '' which is used to render an image (a "badge")

the alternative is to do something like

'[imageref](destlink)' 

but according to trusted sources from the internet, that syntax is not working in github anymore since 2022 and is not supported by most markdown renderers. so I opted for the safe option.

dirkmueller: we can try the other option but I have no way to verify whether the fancy rubygem that the registry webui has supports that functionality.

dcermak: If it's known to be problematic or broken, then don't add it. I was merely asking.

dirkmueller:
okay, so it does render in github. I guess that's good enough then.

PR #1267, opened 2024-06-12
Add PythonDevelopmentContainer class

alexandrevicenzi:
The PythonDevelopmentContainer class extends the DevelopmentContainer class to add extra attributes that only make sense in a Python context.

These attributes can later be used in the README.

### DIFF
--- a/src/bci_build/package/python.py
+++ b/src/bci_build/package/python.py
@@ -1,6 +1,7 @@
 """Python language related BCI containers"""
 
 import datetime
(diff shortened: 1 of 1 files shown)


dcermak (approved):
This is so unintrusive, that it can go in imho. Love the simplicity :+1:

dirkmueller (approved):

dirkmueller:
LGTM

PR #1268, opened 2024-06-12
Package category

alexandrevicenzi:
This is a solution to the additional package problem in #1195.

This is one of my proposals, it adds extra variables and functions, but in theory, it does not affect anything that does not use the new attributes.

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -120,6 +120,17 @@ class PackageType(enum.Enum):
 
 
 @enum.unique
+class PackageCategory(enum.Enum):
+    """The category of a package in the image."""
+
+    CORE = "core"
+    EXTRA = "extra"
+
+    def __str__(self) -> str:
+        return self.value
+
+
+@enum.unique
 class OsVersion(enum.Enum):
     """Enumeration of the base operating system versions for BCI."""
 
@@ -267,6 +278,9 @@ class Package:
     #: The package type. This parameter is only applicable for kiwi builds and
     #: defines into which ``<packages>`` element this package is inserted.
     pkg_type: PackageType = PackageType.IMAGE
+
+    #: The package category. This parameter is intendend for documentation.
+    pkg_category: PackageCategory = PackageCategory.CORE
 
     def __str__(self) -> str:
         return self.name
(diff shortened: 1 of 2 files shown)


src/bci_build/package/python.py
```diff
@@ -56,7 +57,13 @@ def _get_python_kwargs(py3_ver: _PYTHON_VERSIONS, os_version: OsVersion):
...
             "PIP_VERSION": pip3_replacement,
         },
-        "package_list": [f"{py3}-devel", py3, pip3, "curl", "git-core"]
+        "package_list": [
+            Package(py3),
+            Package(pip3),
+            f"{py3}-devel",
+            "curl",
+            "git-core",
+        ]
```
dcermak, line 66: Wrong type, `package_list` is `list[str]|list[Package]` not `list[str|Package]`. We can change the type, but that will require further code adjustments.

src/bci_build/package/__init__.py
```diff
@@ -846,6 +860,34 @@ def packages(self) -> str:
...
+                name = pkg.name
+            else:
+                name = pkg
+
+            if name in self.os_version.lifecycle_data_pkg:
+                continue
+
+            package_list.append(name)
+
+        return sorted(package_list)
```
dcermak, line 889: I think that the extra packages should be an opt-in feature and not opt-out. With the current behavior, `pretty_extra_packages` is equal to `package_list` for all containers built with Docker/Podman. And for those using kiwi + bootstrap packages, it'll be empty. I'd rather try this:
```suggestion
    @property
    def pretty_extra_packages(self) -> list[str]:
        """Returns the non-core packages in this image. It is intended
        to be used in the image documentation to create an additional
        packages section.

        """
        return sorted([pkg.name for pkg in self.package_list if isinstance(pkg, Package) and pkg.pkg_category == PackageCategory.CORE])
```

dcermak (commented):
I like the general idea of this, although I think that we should think about the details a bit more. E.g. whether this package list should be opt-in or opt-out.

And I also had the idea whether it would be worth adding an optional `description` element to the `Package` class, that could get printed into the documentation.

src/bci_build/package/python.py
alexandrevicenzi: Yes, true, I missed that.

My idea was to only allow `list[Package]`, but this would affect all images.
It has its benefits on the long run if we go this way.

src/bci_build/package/__init__.py
alexandrevicenzi: Yes, that is something I was considering. I believe if we make exclusive `PackageCategory.EXTRA` it is a lot clearer what this function will output to a README.

You wrote `PackageCategory.CORE`, I assume you intended to write `PackageCategory.EXTRA` based on your comment.

alexandrevicenzi:
Closing this as #1267 was merged.

PR #1272, opened 2024-06-12
Use sentence-style capitalization

dirkmueller:


alexandrevicenzi:
The `pretty_name` variable is used in the READMEs, we need to also change the README, otherwise, it will look like "Some package Container Image" and still be inconsistent.

alexandrevicenzi (commented):
The pretty_name variable is used in the READMEs, we need to also change the README, otherwise, it will look like "Some package Container Image" and still be inconsistent.

dirkmueller:
> The pretty_name variable is used in the READMEs, we need to also change the README, otherwise, it will look like "Some package Container Image" and still be inconsistent.

good catch. better now?

alexandrevicenzi (approved):

dcermak (approved):

dirkmueller:
(checked with documentation team outside the PR, they're approving the change)

PR #1275, opened 2024-06-13
Use the version specific entrypoint/healthcheck scripts (Fixes #1207)

### LINKED ISSUE #1207: mariadb entrypoint.sh / healthcheck.sh
@dcermak,

note per https://github.com/MariaDB/mariadb-docker/commit/44ed2e231207787c08d56acc94b79d5f06efe006

%%MARIADB_VERSION_BASIC%% needs to be populated with the version of the release.
https://build.opensuse.org/package/rdiff/openSUSE:Factory/mariadb-image?linkrev=base&rev=13 wasn't sufficient.

note also healthcheck.sh substitutions

https://github.com/MariaDB/mariadb-docker/blob/44ed2e231207787c08d56acc94b79d5f06efe006/update.sh#L103
https://github.com/MariaDB/mariadb-docker/blob/44ed2e231207787c08d56acc94b79d5f06efe006/update.sh#L114

with 10.4 having its last release, yes I need to cleanup the update.sh script more.

If complete Docker Official Images compatibility is a goal running
https://github.com/MariaDB/mariadb-docker/blob/master/.test/run.sh (with the other files in .test cloned) with the container image as the argument will run tests.

Some components of https://github.com/MariaDB/mariadb-docker/blob/master/docker.cnf might be required for compatibility along with `jemalloc` installation.

dirkmueller:
This is needed to pick up the right entrypoint script has there are numerous modifications between each version. upstream nicely prepares them as discussed in #1207


update-files.sh
```diff
@@ -2,17 +2,20 @@
...
+curl -sf -o src/bci_build/package/nginx/LICENSE https://raw.githubusercontent.com/nginxinc/docker-nginx/master/LICENSE
+curl -sf -o src/bci_build/package/nginx/20-envsubst-on-templates.sh https://raw.githubusercontent.com/nginxinc/docker-nginx/master/entrypoint/20-envsubst-on-templates.sh
+curl -sf -o src/bci_build/package/nginx/30-tune-worker-processes.sh https://raw.githubusercontent.com/nginxinc/docker-nginx/master/entrypoint/30-tune-worker-processes.sh
 
-curl -o src/bci_build/package/grafana/run.sh https://raw.githubusercontent.com/grafana/grafana/main/packaging/docker/run.sh
+curl -sf -o src/bci_build/package/grafana/run.sh https://raw.githubusercontent.com/grafana/grafana/main/packaging/docker/run.sh
 
-curl -o src/bci_build/package/mariadb/entrypoint.sh https://raw.githubusercontent.com/MariaDB/mariadb-docker/master/docker-entrypoint.sh
-curl -o src/bci_build/package/mariadb/healthcheck.sh https://raw.githubusercontent.com/MariaDB/mariadb-docker/master/healthcheck.sh
+for v in 10.6 10.11 11.2; do
```
alexandrevicenzi, line 15: Maybe you want to consider picking the versions from `package_versions.json`, it would require `jq` or perhaps `python` to extract the version.

The versions are not hardcoded anywhere else, one needs to remember to change the json file and this bash file.

update-files.sh
```diff
@@ -2,17 +2,20 @@
...
+    mkdir -p src/bci_build/package/mariadb/$v
```
alexandrevicenzi, line 16: If a new version is added, it will add a new folder, but it won't remove a folder of an older version. Maybe consider removing all version folders prior to download.

alexandrevicenzi (approved):

update-files.sh
dirkmueller: done

update-files.sh
dirkmueller: I can do that, but the github workflow is not git adding or removing things so it'll just break differently. either way it will need manual adjustment.

alexandrevicenzi: yes, true, but at least there are no hardcoded versions.

alexandrevicenzi (approved):

PR #1293, opened 2024-06-15
Reuse the original comment from the gh action bot and don't create new comments

dcermak:
Since we now stopped creating a new branch per new build, there's really absolutely no need to create new comments. It just needlessly spams the PR sending out notifications to everyone.

### DIFF
--- a/.github/workflows/obs_build.yml
+++ b/.github/workflows/obs_build.yml
@@ -100,6 +100,8 @@ jobs:
         uses: peter-evans/create-or-update-comment@v4
         id: create_comment
         with:
+          comment-id: ${{ steps.find_comment.outputs.comment-id || '' }}
+          edit-mode: replace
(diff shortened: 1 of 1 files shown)


src/bci_build/package/appcontainers.py
```diff
@@ -41,7 +41,7 @@ def _envsubst_pkg_name(os_version: OsVersion) -> str:
 PCP_CONTAINERS = [
     ApplicationStackContainer(
         name="pcp",
-        pretty_name="Performance Co-Pilot (pcp)",
+        pretty_name="Performance Co-Pilot",
```
alexandrevicenzi, line 44: > WIP: dummy change

is this commit supposed to be here?

dcermak: Nope, I added that to get an actual change. Let me drop it.

alexandrevicenzi (approved):

PR #1298, opened 2024-06-17
Add minor versions to Prometheus, Blackbox & Alertmanager containers

dcermak:
This will no longer require to hardcode patch versions in https://github.com/SUSE/BCI-tests/pull/503 and allow us to resort to minor or major versions

### DIFF
--- a/src/bci_build/package/appcontainers.py
+++ b/src/bci_build/package/appcontainers.py
@@ -203,7 +203,7 @@ HEALTHCHECK --interval=10s --start-period=10s --timeout=5s \
     + [(pg_ver, OsVersion.TUMBLEWEED) for pg_ver in (14, 13, 12)]
 ]
 
-PROMETHEUS_PACKAGE_NAME = "golang-github-prometheus-prometheus"
+_PROMETHEUS_PACKAGE_NAME = "golang-github-prometheus-prometheus"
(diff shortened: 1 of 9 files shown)


witekest:
Thanks, I'll change Grafana definition in the same way.

dcermak:
> Thanks, I'll change Grafana definition in the same way.

I can do that as part of this pr as well

src/bci_build/package/appcontainers.py
```diff
@@ -203,87 +203,96 @@ def _envsubst_pkg_name(os_version: OsVersion) -> str:
...
         replacements_via_service=[
             Replacement(
-                regex_in_build_description="%%prometheus_version%%",
-                package_name=PROMETHEUS_PACKAGE_NAME,
-                parse_version="patch",
+                regex_in_build_description=f"%%prometheus_{level}_version%%",
+                package_name=_PROMETHEUS_PACKAGE_NAME,
+                parse_version=level,
             )
+            for level in ("major", "minor", "patch")
```
alexandrevicenzi, line 228: maybe consider creating an Enum for levels since it is used in multiple places.

dcermak: Right, that's actually better

alexandrevicenzi (approved):

PR #1306, opened 2024-06-18
Add HEALTHCHECK to the Prometheus family of containers

dcermak:


src/bci_build/package/appcontainers.py
```diff
@@ -204,15 +204,23 @@ def _envsubst_pkg_name(os_version: OsVersion) -> str:
     + [(pg_ver, OsVersion.TUMBLEWEED) for pg_ver in (14, 13, 12)]
 ]
 
+
+def _generate_prometheus_family_healthcheck(port: int) -> str:
+    return f"""HEALTHCHECK --interval=5s --timeout=5s --retries=5 \
+    CMD ["/bin/sh", "-c", "[ \"$(/usr/bin/curl http://localhost:{port}/-/healthy)" = \"Healthy\" ]"]
```
witekest, line 210: Unfortunately the API specification is not consistent here. Every component returns a different [response text](https://github.com/SUSE/BCI-tests/pull/503/files#diff-5f219bf2012d07759723f500675e1769613674c398febbb2792a508257126ca1R39). Let's add another parameter to the function.

witekest (changes requested):

witekest (approved):
Looks good to me. Thank you.

src/bci_build/package/appcontainers.py
```diff
@@ -204,15 +204,23 @@ def _envsubst_pkg_name(os_version: OsVersion) -> str:
     + [(pg_ver, OsVersion.TUMBLEWEED) for pg_ver in (14, 13, 12)]
 ]
 
+
+def _generate_prometheus_family_healthcheck(port: int, expected_response: str) -> str:
+    return rf"""HEALTHCHECK --interval=5s --timeout=5s --retries=5 \
+    CMD ["/usr/bin/test", "\"$(/usr/bin/curl http://localhost:{port}/-/healthy)\" = \"{expected_response}\""]
```
dirkmueller, line 210:
```suggestion
    CMD ["/usr/bin/test", "\"$(/usr/bin/curl -sf http://localhost:{port}/-/healthy)\" = \"{expected_response}\""]
```

src/bci_build/package/appcontainers.py
```diff
@@ -229,20 +237,24 @@ def _envsubst_pkg_name(os_version: OsVersion) -> str:
             for level in (ParseVersion.MAJOR, ParseVersion.MINOR, ParseVersion.PATCH)
         ],
         volumes=["/var/lib/prometheus"],
-        exposes_tcp=[9090],
+        exposes_tcp=[_PROMETHEUS_PORT],
+        custom_end=_generate_prometheus_family_healthcheck(
+            _PROMETHEUS_PORT, "Prometheus Server is Healthy."
```
dirkmueller, line 242: Do we really need to check for this specific string to be returned? from what I've seen in kubernetes manifests, it simply sets this as a livenessProbe, aka as long as the http return code is 200-399 it's good.

src/bci_build/package/appcontainers.py
dirkmueller: imho we should only check for http code not for strings, those could change.

src/bci_build/package/appcontainers.py
dcermak: @witekest what do you think about that suggestion? I'd prefer it, as it is simpler, but I'm not sure how the Prometheus stack behaves exactly

witekest: Agree. The string is always the same and does not provide additional information.

dirkmueller (approved):

witekest (approved):

PR #1314, opened 2024-06-19
Mark SLE15 SP6 as latest, remove SP5 overlap containers

dirkmueller:
This removes creation of SP5 containers for which a SP6 variant exists.

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -241,7 +241,6 @@ RELEASED_OS_VERSIONS: list[OsVersion] = [
 
 # For which versions to create Application and Language Containers?
 ALL_NONBASE_OS_VERSIONS: list[OsVersion] = [
-    OsVersion.SP5,
     OsVersion.SP6,
     OsVersion.TUMBLEWEED,
 ]
@@ -263,7 +262,7 @@ ALL_OS_VERSIONS: set[OsVersion] = {
 }
 
 CAN_BE_LATEST_OS_VERSION: list[OsVersion] = [
-    OsVersion.SP5,
+    OsVersion.SP6,
     OsVersion.TUMBLEWEED,
     OsVersion.BASALT,
 ]
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -190,6 +190,10 @@ class OsVersion(enum.Enum):
         )
 
     @property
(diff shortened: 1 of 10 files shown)


dirkmueller:
> Java is not touched by this, I guess that still needs to be fixed?

I think the agreement was that java 11,17 die with SP5 and 21 is anyway SP6 only. still need to confirm that. but it shoudl be good as is.

dcermak:

>> Java is not touched by this, I guess that still needs to be fixed?
>
> I think the agreement was that java 11,17 die with SP5 and 21 is anyway SP6 only. still need to confirm that. but it shoudl be good as is. 

SCC lists java 11 and 17 as part of SP6:
https://scc.suse.com/packages?name=SUSE+Linux+Enterprise+Server&version=15.6&arch=x86_64&query=openjdk&module=

(openjdk 11 is part of the legacy module, but openjdk17 is in the basesystem)

dirkmueller:
> Dirk Mueller ***@***.***> writes:

> > Java is not touched by this, I guess that still needs to be fixed? I think the agreement was that java 11,17 die with SP5 and 21 is anyway SP6 only. still need to confirm that. but it shoudl be good as is.
> SCC lists java 11 and 17 as part of SP6: https://scc.suse.com/packages?name=SUSE+Linux+Enterprise+Server&version=15.6&arch=x86_64&query=openjdk&module= (openjdk 11 is part of the legacy module, but openjdk17 is in the basesystem)

yes, that is the "still need to confirm that". legacy module is excluded from SLE BCI so we don't need to discuss it. and java17 is considered deprecated in the release notes: 

https://www.suse.com/releasenotes/x86_64/SUSE-SLES/15-SP6/index.html#java-version

as that is contradicting, I summarized that as "still need to confirm that". there's a separate discussion about it. when it's confirmed, I'll update the PR with the rationale and mark it ready for review. right now it's just a draft.

dcermak (changes requested):
This needs to touch openjdk in one way or the other. Now we don't ship a latest openjdk anymore.

PVlasin (approved):
OK

dirkmueller:
> This needs to touch openjdk in one way or the other. Now we don't ship a latest openjdk anymore.

the latest tag is 21 now on SP6 and 22 on tumbleweed.

dcermak (approved):

PR #1326, opened 2024-06-21
Merge update actions

dcermak:
#1325 showed a flaw in our auto-update logic: if the mariadb version is bumped, then everything burns down in flames, as the necessary files are no longer present. To fix that, we **must** fetch the new files in the same PR to have a seamless update UX


alexandrevicenzi (dismissed):

.github/workflows/update-versions-and-files.yml
```diff
@@ -28,17 +28,23 @@ jobs:
           OSC_PASSWORD: ${{ secrets.OSC_PASSWORD }}
           OSC_USER: "defolos"
 
-      - name: create a pull request with the updated versions
+      # update the files **after** we updated the versions to get the new
+      # mariadb files
```
dirkmueller, line 32: this comment is hard to parse. how about

```suggestion
# update versions first so that updating the files  can fetch based on the updated versions
```

.github/workflows/update-versions-and-files.yml
```diff
@@ -28,17 +28,23 @@ jobs:
...
+          title: "🤖: Update package versions and files"
           body: |
-            This is an automated update via poetry run update-versions
+            This is an automated update via poetry run update-versions and ./update-files.sh
           base: main
-          branch: create-pull-request/update-versions
+          branch: create-pull-request/update-versions-and-files
           add-paths: |
-            src/bci_build/package/package_versions.json
+            src/bci_build/package/*
```
dirkmueller, line 50: eh, does this work recursively?

dcermak: Yes, see the deleted `.github/workflows/update-scripts.yml`

dirkmueller (approved):

PR #1349, opened 2024-06-25
Unite on standard tools in all language development images

dirkmueller:
Recently users have reported failures due to procps disappearing from the python language containers. Standardize on the same set of packages across all the containers and include procps.

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -183,7 +183,11 @@ class OsVersion(enum.Enum):
     def common_devel_packages(self) -> list[str]:
         """Returns a list of common development packages that are needed for
         all development containers"""
-        return ["findutils", "gawk", "git-core", "curl", "procps"]
+        r = set(("findutils", "gawk", "git-core", "curl", "procps"))
+        if self.is_tumbleweed:
+            r.add("util-linux")
+
+        return sorted(list(r))
 
     @property
     def is_sle15(self) -> bool:


src/bci_build/package/__init__.py
```diff
@@ -179,6 +179,12 @@ def lifecycle_data_pkg(self) -> list[str]:
             return ["lifecycle-data-sle-module-development-tools"]
         return []
 
+    @property
+    def common_devel_packages(self) -> list[str]:
+        """Returns a list of common development packages that are needed for
+        all development containers"""
+        return ["findutils", "gawk", "git-core", "curl", "procps"]
```
dcermak, line 186: I think this should be a constant and not a property of `OsVersion` as it is completely version agnostic

dirkmueller: it right now is, but it might change, package names could be different between major versions, and we have already other similar functions here on os_version. Are you saying it should be a constant until it has a version dependency even though all other similar functions are in os_version ?

src/bci_build/package/__init__.py
```diff
@@ -179,6 +179,16 @@ def lifecycle_data_pkg(self) -> list[str]:
             return ["lifecycle-data-sle-module-development-tools"]
         return []
 
+    @property
+    def common_devel_packages(self) -> list[str]:
+        """Returns a list of common development packages that are needed for
+        all development containers"""
+        return (
+            ["findutils", "gawk", "git-core", "curl", "procps"] + ["util-linux"]
+            if self.value in (OsVersion.TUMBLEWEED.value,)
```
dcermak, line 188: This looks nicer imho

```suggestion
            if self.is_tumbleweed
```

src/bci_build/package/__init__.py
```diff
@@ -179,6 +179,16 @@ def lifecycle_data_pkg(self) -> list[str]:
...
+            else []
```
dcermak, line 189: Let's please use braces around the last part, otherwise it's a bit hard to spot what is actually happening here.
```suggestion
            ["findutils", "gawk", "git-core", "curl", "procps"] + (["util-linux"]
            if self.value in (OsVersion.TUMBLEWEED.value,)
            else [])
```

dirkmueller: I don't see a difference. I'll change this to be more traditional and more explicit.

src/bci_build/package/__init__.py
dirkmueller: done

alexandrevicenzi (approved):

PR #1359, opened 2024-07-01
Limit Spack to sle15 (bsc#1224345)

dirkmueller:


src/bci_build/package/spack.py
```diff
@@ -62,6 +62,9 @@
         },
         # HPC module only exists for those two arches (bsc#1224130)
         exclusive_arch=[Arch.AARCH64, Arch.X86_64],
+        extra_labels={
+            "usage": "This container is only enabled and supported to be used on a SLE15+ host."
```
alexandrevicenzi, line 66: Perhaps:

> This container is enabled and supported only on an SLE15+ host.

alexandrevicenzi (approved):

src/bci_build/package/spack.py
dirkmueller: I like that, changed.

alexandrevicenzi (approved):

PR #1375, opened 2024-07-05
Improve the golang container

dirkmueller:
use a workdir like the upstream image, and set the toolchain mode to be always local.

### DIFF
--- a/src/bci_build/package/golang.py
+++ b/src/bci_build/package/golang.py
@@ -6,6 +6,7 @@ from typing import Literal
 
 from bci_build.package import CAN_BE_LATEST_OS_VERSION
 from bci_build.package import DOCKERFILE_RUN
+from bci_build.package import LOG_CLEAN
 from bci_build.package import DevelopmentContainer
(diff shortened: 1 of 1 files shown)


src/bci_build/package/golang.py
```diff
@@ -67,7 +69,11 @@ def _get_golang_kwargs(
             f"""
             # only available on go's tsan_arch architectures
             #!ArchExclusiveLine: x86_64 aarch64 s390x ppc64le
-            {DOCKERFILE_RUN} if zypper -n install {go}-race; then zypper -n clean; rm -rf /var/log/*; fi
+            {DOCKERFILE_RUN} if zypper -n install {go}-race; then zypper clean; done
+            {DOCKERFILE_RUN} install -m 755 -d /go/bin /go/src
+            {LOG_CLEAN}
```
dcermak, line 74: ftbfs otherwise
```suggestion
            {DOCKERFILE_RUN} {LOG_CLEAN}
```

dcermak (approved):

PR #1380, opened 2024-07-08
Build tomcat container per OpenJDK version

dirkmueller:
This allows to provide pre-built containers that are using specific older JDK versions in the future

### DIFF
--- a/src/bci_build/package/apache_tomcat.py
+++ b/src/bci_build/package/apache_tomcat.py
@@ -15,19 +15,23 @@ assert _TOMCAT_VERSIONS == sorted(_TOMCAT_VERSIONS)
 TOMCAT_CONTAINERS = [
     ApplicationCollectionContainer(
         name="apache-tomcat",
-        pretty_name=f"Apache Tomcat {tomcat_major}",
-        package_name=f"apache-tomcat-{tomcat_major}-image"
+        pretty_name="Apache Tomcat",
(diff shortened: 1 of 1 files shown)


src/bci_build/package/appcontainers.py
```diff
@@ -556,27 +555,26 @@ def _get_nginx_kwargs(os_version: OsVersion):
...
             (os_version in CAN_BE_LATEST_OS_VERSION)
             and tomcat_major == _TOMCAT_VERSIONS[-1]
         ),
-        version=tomcat_major,
-        additional_versions=["%%tomcat_version%%", "%%tomcat_minor%%"],
+        version=f"{tomcat_major}-openjdk{jre_version}",
+        additional_versions=[
+            f"%%tomcat_version%%-openjdk{jre_version}",
+            f"%%tomcat_minor%%-openjdk{jre_version}",
+        ],
```
dcermak, line 569: This leads to tags of the form: `9-openjdk21` or `10-openjdk22`, which are nothing like we provide atm and don't really work too well with semver. I'd much rather have `tomcat-openjdk22:10` instead of `tomcat:10-openjdk22`

dirkmueller: why is "semver" a constraint?

dirkmueller:
as discussed..

PR #1385, opened 2024-07-09
Drop the super ugly SLE15 hack and use a custom class instead

dcermak:
Proposal so that we don't have to use a super ugly hack in the code


src/bci_build/package/__init__.py
```diff
@@ -1261,11 +1261,11 @@ def kiwi_additional_tags(self) -> str | None:
         the image itself.
 
         """
-        extra_tags = []
+        extra_tags = set()
```
dirkmueller, line 1264: this breaks sorting predictability

src/bci_build/package/base.py
```diff
@@ -93,6 +94,18 @@ def _get_base_config_sh_script(os_version: OsVersion) -> str:
...
 
 
+@dataclass
+class Sles15Image(OsContainer):
+    @property
+    def build_tags(self) -> list[str]:
+        return list(
+            set(
+                super().build_tags
+                + ["suse/sle15:%OS_VERSION_ID_SP%", f"suse/sle15:{self.version_label}"]
```
dirkmueller, line 104: this changes the first tag to be `tag="%OS_VERSION_ID_SP%.%RELEASE%"` rather than `tag="%OS_VERSION_ID_SP%"`. 

I'm not sure how important that is to preserve, but it is a behavior change (I think it will lead to a different buildversion, which might have funky side effects.

dirkmueller:
I've merged the idea of a subclass into #1220 , thanks!

PR #1396, opened 2024-07-11
Move tomcat into ApplicationCollection

dirkmueller:
* Adjust for new registry publishing location
* Adjust name to "apache-tomcat" based on the Rancher naming schema


src/bci_build/package/appcollection.py
```diff
@@ -0,0 +1,122 @@
...
+        super().__post_init__()
+        if not self.is_opensuse and not self.exclusive_arch:
+            self.exclusive_arch = [Arch.AARCH64, Arch.X86_64]
+
+
+_TOMCAT_VERSIONS: list[int] = [9, 10]
+assert _TOMCAT_VERSIONS == sorted(_TOMCAT_VERSIONS)
+
+TOMCAT_CONTAINERS = [
+    ApplicationCollectionContainer(
```
alexandrevicenzi, line 59: In my understanding, we should publish to the SUSE registry and as a Rancher application. With this change, it seems only Rancher is supported.

Am I missing something? Won't this affect current images?

dirkmueller: we have never published the current images, this was just dead code before.

src/bci_build/package/appcollection.py
```diff
@@ -0,0 +1,46 @@
...
+
+from dataclasses import dataclass
+
+from bci_build.package import ApplicationStackContainer
+from bci_build.package import Arch
+
+
+@dataclass
+class ApplicationCollectionContainer(ApplicationStackContainer):
+    """Containers for the Rancher Application Collection Distribution Platform."""
```
dcermak, line 11: Will we not publish language stack containers to the application collection?

src/bci_build/package/apache_tomcat.py
```diff
@@ -0,0 +1,81 @@
...
+from bci_build.package import OsVersion
+from bci_build.package import ParseVersion
+from bci_build.package import Replacement
+
+from .appcollection import ApplicationCollectionContainer
+
+_TOMCAT_VERSIONS: list[int] = [9, 10]
+assert _TOMCAT_VERSIONS == sorted(_TOMCAT_VERSIONS)
+
+TOMCAT_CONTAINERS = [
```
dcermak, line 15: Can we **please** stop moving containers around between files? It makes blame & log needlessly annoying.

src/bci_build/package/appcollection.py
dirkmueller: what are you suggesting? calling it "ApplicationCollectionApplicationContainer" so that we have a "ApplicationCollectionDevelopmentContainer" when the need arises? so far no development container has been requested.

src/bci_build/package/apache_tomcat.py
dirkmueller: it no longer makes sense in appcontainers because those are the "now legacy" application containers. and imho each container should be in its own file. as we had to move it anyway, splitting it into its separate file ensures that annotation/blaming will continue going forward. 

btw, there is `--follow-renames`, fjyi

alexandrevicenzi: I agree that each container should be its own file, it is easier to manage in the long term. As I see, we are entering a new chapter of the project which justifies such changes.

src/bci_build/package/appcollection.py
alexandrevicenzi: I missed that. Not an issue anymore.

src/bci_build/package/apache_tomcat.py
dcermak: Brilliant, can you please then move the code in a separate commit than the changes? It at least makes reviews simpler.

src/bci_build/package/appcollection.py
```diff
@@ -0,0 +1,46 @@
...
+        if self.is_opensuse:
+            return super().title
+        return self.pretty_name
+
+    @property
+    def eula(self) -> str:
+        if self.is_opensuse:
+            return "sle-bci"
+        return "sle-eula"
+
```
dcermak, line 42: This is exactly the same code as in `ApplicationStackContainer`
```suggestion
```

src/bci_build/package/appcollection.py
```diff
@@ -0,0 +1,46 @@
...
+    def url(self) -> str:
+        if self.is_opensuse:
+            return super().url
+        return f"https://apps.rancher.io/applications/{self.name}"
+
+    @property
+    def title(self) -> str:
+        if self.is_opensuse:
+            return super().title
+        return self.pretty_name
```
dcermak, line 35: Why do we need this change?

src/bci_build/package/appcollection.py
dirkmueller: @dcermak  the patch wasn't merged yet when I wrote this code, hence the duplication.

src/bci_build/package/appcollection.py
dcermak: If we will be publishing everything to the application catalog, then it would make sense to have a flag in the container instead of creating subclasses imho

src/bci_build/package/apache_tomcat.py
dirkmueller: splitted into two commits

src/bci_build/package/appcollection.py
dirkmueller: yeah, I was thinking about creating a RegistryMixIn for this but I'm slowly losing patience and energy to do bigger architectural changes on the codebase. 

I think we are not intending to move everything over in one go, and some of the content might remain in two places. but as the descriptions and other labels are not the same, we need to generate them twice.

src/bci_build/package/appcollection.py
dirkmueller: Because Rancher App collection doesn't use the full title for title but simply the application name.

alexandrevicenzi (approved):

src/bci_build/package/apache_tomcat.py
```diff
@@ -0,0 +1,82 @@
...
+
+WORKDIR $CATALINA_HOME
+""",
+        entrypoint_user="tomcat",
+        logo_url="https://tomcat.apache.org/res/images/tomcat.png",
+    )
+    for tomcat_major, os_version in (
+        (10, OsVersion.TUMBLEWEED),
+        (9, OsVersion.TUMBLEWEED),
+        (10, OsVersion.SP6),
```
dcermak, line 80: Is it intentional that you don't build tomcat 9 from SP6?

src/bci_build/package/apache_tomcat.py
```diff
@@ -0,0 +1,82 @@
...
+        name="apache-tomcat",
+        pretty_name=f"Apache Tomcat {tomcat_major}",
+        package_name=f"apache-tomcat-{tomcat_major}-image"
+        if os_version.is_tumbleweed
+        else f"sac-apache-tomcat-{tomcat_major}-image",
+        os_version=os_version,
+        is_latest=(
+            (os_version in CAN_BE_LATEST_OS_VERSION)
+            and tomcat_major == _TOMCAT_VERSIONS[-1]
+            and os_version.is_tumbleweed
```
dcermak, line 26: This condition drops the latest tag from SP6, is that intentional?

dirkmueller: yes, we talked about that already, appcol doesn't have `:latest tags

src/bci_build/package/apache_tomcat.py
dirkmueller: yes, the intention is to start with tomcat 10 integration.

src/bci_build/package/apache_tomcat.py
dcermak: :scream:

PR #1404, opened 2024-07-15
Allow kpartx and disk compression with kiwi

dirkmueller:


src/bci_build/package/kiwi.py
```diff
@@ -28,6 +28,7 @@
         package_list=[
             "btrfsprogs",
             "checkmedia",
+            "dosfstools",
```
dcermak, line 31:
```suggestion
```

src/bci_build/package/kiwi.py
```diff
@@ -36,16 +37,19 @@
...
             "kiwi-systemdeps-core",
+            "kpartx",
             "libxml2-devel",
             "lvm2",
             "make",
             "netcat-openbsd",
             "python3-devel",
             "python3-kiwi",
             "python3-pip",
+            "squashfs",
```
dcermak, line 48:
```suggestion
            "kiwi-systemds-disk-images",
            "libxml2-devel",
            "lvm2",
            "make",
            "netcat-openbsd",
            "python3-devel",
            "python3-kiwi",
            "python3-pip",
```

dcermak (commented):
We can rely on kiwi's metapackages to cut the package list a bit

src/bci_build/package/kiwi.py
dirkmueller: that doesn't exist here. we can use kiwi-systemdeps-filesystems though hopefully

PR #1406, opened 2024-07-15
fix kiwi package name/version messup

dirkmueller:


dcermak (commented):
I would suggest to only use the major+minor version, as we'll otherwise have to submit changes to this container every week

dcermak:
And please also run `poetry run update-versions`

src/bci_build/package/kiwi.py
```diff
@@ -47,6 +48,7 @@
             "timezone",
             "xfsprogs",
         ],
+        custom_end=f"{generate_package_version_check('python3-kiwi', kiwi_ver, 'minor')}",
```
dcermak, line 51: this and a reformat:
```suggestion
        custom_end=f"{generate_package_version_check('python3-kiwi', to_major_minor_version(kiwi_ver), 'minor')}",
```

PR #1407, opened 2024-07-15
Drop the SLE 15 Base hack in the kiwi.xml template

dcermak:
Introduce an attribute for OBS magic comments instead


src/bci_build/package/base.py
```diff
@@ -123,6 +123,8 @@ def _get_base_kwargs(os_version: OsVersion) -> dict:
         "build_recipe_type": BuildType.KIWI,
         "from_image": None,
         "os_version": os_version,
+        "kiwi_obs_magic_comments": """<!-- OBS-IgnorePackage: rpm -->
```
dirkmueller, line 126: we only want those in the case if `os_version.is_sle15`

src/bci_build/package/__init__.py
```diff
@@ -620,6 +620,10 @@ class BaseContainerImage(abc.ABC):
     #: present
     logo_url: str = ""
 
+    #: Additional xml comments that are pasted at the top of the kiwi built
+    #: recipe that are processed by OBS
+    kiwi_obs_magic_comments: str = ""
```
dirkmueller, line 625: I'm not convinced this is the right abstraction or will be used for other images going forward. for the right level of abstraction, it should have some semantic other than "magic", which means two fields are needed, one for excludearch and one for ignorepackages. but we don't actually want to encourage it being used elsewhere.

dcermak: > will be used for other images going forward

Probably not, but that's not the point of this. The point is to keep the all the information of one container in a single place and not have it scattered across multiple files. The first version had multiple crucial implementation details scattered in three different places, which might be easy for you to keep in your mind, but for the average dev like me, that's not possible. Keeping everything in one place reduces the mental load tremendously and makes debugging down the line much simpler.

dirkmueller: > Probably not, but that's not the point of this. The point is to keep the all the information of one container in a single place and not have it scattered across multiple files. 

Fair point. the change was incorporated in a similar fashion on the other PR that got merged.

PR #1416, opened 2024-07-16
Install python 3.11 wheel for PEP517 wheel builds (bsc#1202140)

dirkmueller:


src/bci_build/package/python.py
```diff
@@ -45,14 +45,20 @@ def _get_python_kwargs(py3_ver: _PYTHON_VERSIONS, os_version: OsVersion):
...
+    has_wheel: bool = is_system_py
+
+    if os_version.is_tumbleweed:
+        # Tumbleweed rocks
         has_pipx = True
+        has_wheel = True
+    elif os_version.is_sle15:
+        if py3_ver != "3.6":
+            # Enabled only for Python 3.11+ on SLE15 (jsc#PED-5573)
+            has_pipx = True
```
dcermak, line 58:
```suggestion
        # Enabled only for Python 3.11+ on SLE15 (jsc#PED-5573)
        has_pipx = is_system_py
```

src/bci_build/package/python.py
```diff
@@ -45,14 +45,20 @@ def _get_python_kwargs(py3_ver: _PYTHON_VERSIONS, os_version: OsVersion):
...
     pip3_replacement = "%%pip_ver%%"
-    has_pipx = False
-    has_wheel = True if is_system_py or os_version == OsVersion.TUMBLEWEED else False
-    # Tumbleweed rocks
-    if os_version == OsVersion.TUMBLEWEED:
-        has_pipx = True
-    # Enabled only for Python 3.11 on SLE15 (jsc#PED-5573)
-    if os_version not in (OsVersion.BASALT, OsVersion.TUMBLEWEED) and py3_ver == "3.11":
+    has_pipx: bool = False
+    has_wheel: bool = is_system_py
```
dcermak, line 49: Do we have to set these? We cover all cases in the `if - elif` below

src/bci_build/package/python.py
```diff
@@ -45,14 +45,20 @@ def _get_python_kwargs(py3_ver: _PYTHON_VERSIONS, os_version: OsVersion):
...
+        # Tumbleweed rocks
         has_pipx = True
+        has_wheel = True
+    elif os_version.is_sle15:
+        if py3_ver != "3.6":
+            # Enabled only for Python 3.11+ on SLE15 (jsc#PED-5573)
+            has_pipx = True
+        # py3.12 pending discussion
+        if py3_ver not in ("3.6", "3.12"):
+            has_wheel = True
```
dcermak, line 61:
```suggestion
        # py3.12 pending discussion
        has_wheel = py3_ver == "3.11"
```

src/bci_build/package/python.py
dirkmueller: no, that would be the inverse of the condition..

src/bci_build/package/python.py
dirkmueller: not right now, but for basalt/slcc/slac we would. I can remove them for now.

src/bci_build/package/python.py
dirkmueller: yeah, I picked this style so that when we add a new version, it pulls in wheel, rather than not pulling in wheel (which happened now and was missed that way). hence I wanted it opt out, so that it fails on the next version addition rather than silently missing a feature.

src/bci_build/package/python.py
dirkmueller: actually I can not remove them because these are local variables not the dataclass attributes.

PR #1421, opened 2024-07-16
Move build_release setting out of BaseContainerImage class

dcermak:
This is a per container image setting (data) and does not belong into the implementation

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -517,6 +517,12 @@ class BaseContainerImage(abc.ABC):
     #: present
     logo_url: str = ""
 
+    #: Optional build_release setting that will be inserted as the
+    #: ``#!BuildRelease`` magic comment for the build service to ensure that
+    #: the release suffix of a container tag is sequentially increasing (for
+    #: webui sorting)
+    _min_release_counter: int | None = None
+
     def __post_init__(self) -> None:
         self.pretty_name = self.pretty_name.strip()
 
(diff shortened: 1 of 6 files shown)


src/bci_build/package/__init__.py
```diff
@@ -589,22 +594,8 @@ def build_version(self) -> str | None:
...
-            return "41"
-        if "php" in self.build_name:
-            return "30"
-        if "node" in self.build_name:
-            return "30"
-        if isinstance(self, ApplicationStackContainer):
-            return "35"
-        return None
+
+        return str(self._build_release) if self._build_release is not None else None
```
dirkmueller, line 598: well, the values are actually SP specific, so they would have to be different/higher for SP7. that's why I opted for the less clean but more central approach to map those numbers rather than having them as part of the attributes of the containers.

src/bci_build/package/__init__.py
```diff
@@ -517,6 +517,11 @@ class BaseContainerImage(abc.ABC):
     #: present
     logo_url: str = ""
 
+    #: Optional build_release setting that will be inserted as the
+    #: ``#!BuildRelease`` magic comment for the build service to ensure that
+    #: containers sort higher across service packs
```
dirkmueller, line 522:
```suggestion
    #: the release suffix of a container tag is sequentially increasing (for webui sorting)
```

src/bci_build/package/__init__.py
```diff
@@ -517,6 +517,11 @@ class BaseContainerImage(abc.ABC):
...
+    _build_release: int | None = None
```
dirkmueller, line 523: I like this being moved here, but maybe a minor nit to call it `_min_release_counter` or something like that to be more descriptive of its purpose?

alexandrevicenzi (approved):

PR #1426, opened 2024-07-17
Always deinstall compat-usrmerge-tools

dirkmueller:
we don't need them in any base container. Saves a few hundred kbs.


dcermak (commented):
I don't see `compat-usrmerge-tools` in the base container. For which container is this necessary?

dirkmueller:
It's in all the bci-minimal/init/micro containers of tumbleweed or SLFO

dirkmueller:
ah, actually this is due to a local change of mine. nevermind.

PR #1427, opened 2024-07-17
Add SLE16_0 (for 16.0/SLFO) to OsVersion handling

dirkmueller:
Also shorten TUMBLEWEED to TW, which is a bit easier to read imho.


src/bci_build/package/__init__.py
```diff
@@ -214,16 +221,14 @@ def os_version(self) -> str:
         """
         if self.is_sle15:
             return f"15.{str(self.value)}"
-        # FIXME
-        # if self.is_slcc:
-        #     return "16.0"
-
+        if self.value == OsVersion.SLE16_0:
```
dcermak, line 224:
```suggestion
        if self.value == OsVersion.SLE16_0.value:
```

src/bci_build/package/registry/README.md.j2
```diff
@@ -6,8 +6,7 @@ This container image allows to run a local OCI registry.
```
dcermak, line 7: Should the changes in this file go in? Looks like an accidental/unrelated change

dirkmueller: no, this is unrelated. weird.

src/bci_build/package/__init__.py
```diff
@@ -132,10 +132,10 @@ class OsVersion(enum.Enum):
     SP4 = 4
     #: SLE 15 Service Pack 3
     SP3 = 3
+    #: SUSE Linux Framework One
+    SLE16_0 = 16
```
alexandrevicenzi, line 136: can't you use "16.0" here? this way you can use `enum.value` similar to Tumbleweed.

alexandrevicenzi (dismissed):

src/bci_build/package/__init__.py
dirkmueller: we can. I don't know exactly how the basalt/sle sp/component architecture model story is turning out, so I'm a bit conservative with assumptions on numbers. I can change it, I don't think it matters right now one way or the other.

dcermak: We already return `16.0` from `pretty_print` and `OsContainer.os_version`, so imho it would be a bit more consistent to use 16.0 here.

src/bci_build/package/__init__.py
```diff
@@ -356,7 +361,7 @@ def __post_init__(self) -> None:
 def _build_tag_prefix(os_version: OsVersion) -> str:
     if os_version == OsVersion.TUMBLEWEED:
         return "opensuse/bci"
-    if os_version == OsVersion.BASALT:
+    if os_version == OsVersion.SLE16_0:
         return "alp/bci"
```
dcermak, line 365:
```suggestion
        return "bci"
```
or this?
```suggestion
        return "slcc/stable"
```

dirkmueller: my best bet is on "bci" for now, so that's how I changed it.

dcermak (approved):

PR #1437, opened 2024-07-18
Add a supported end date for the SAC Apache Tomcat containers

dirkmueller:
This is the minimum of the JRE and the tomcat support end dates. Don't set anything for openSUSE.

### DIFF
--- a/src/bci_build/package/apache_tomcat.py
+++ b/src/bci_build/package/apache_tomcat.py
@@ -1,8 +1,9 @@
 """Provide an Apache Tomcat container."""
 
+import datetime
+
 from bci_build.package import CAN_BE_LATEST_OS_VERSION
 from bci_build.package import DOCKERFILE_RUN
-from bci_build.package import _SUPPORTED_UNTIL_SLE
 from bci_build.package import OsVersion
 from bci_build.package import ParseVersion
 from bci_build.package import Replacement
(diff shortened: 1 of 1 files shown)


dirkmueller:
(waiting for confirmation on the correctness of the dates)

src/bci_build/package/apache_tomcat.py
```diff
@@ -12,6 +13,36 @@
...
+    tomcat_end_support_dates: dict[int, datetime.date] = {
+        9: datetime.date(2024 + 3, 2, 1),
+        10: datetime.date(2024 + 3, 7, 1),
+    }
+    end_date: datetime.date = min(
+        jre_end_support_dates.get(jre_major, datetime.date.max),
+        tomcat_end_support_dates.get(tomcat_major, datetime.date.max),
+    )
+    if end_date < datetime.date.max:
+        return end_date
```
dcermak, line 42: datetime.date.max is the year 9999, why check against that?

dirkmueller: we are checking here that either tomcat or openjdk had a time set (because of the default attribute in get() call above. if we did then we return the minimum (aka earliest date) of tomcat+openjdk combination.

src/bci_build/package/apache_tomcat.py
```diff
@@ -12,6 +13,36 @@
...
+        return None
+
+    # Taken from https://www.suse.com/releasenotes/x86_64/SUSE-SLES/15-SP6/index.html#java-version
+    jre_end_support_dates: dict[int, datetime.date] = {
+        21: datetime.date(2031, 6, 30),
+        17: datetime.date(2027, 12, 31),
+        11: datetime.date(2026, 12, 31),
+    }
+
+    # We do not have a documented policy, suggestion is to do 3 years from initial
```
alexandrevicenzi, line 31: where this 3y suggestion comes from?

src/bci_build/package/apache_tomcat.py
dcermak: okay, now I understand this. Can you please add a comment explaining this briefly? Or maybe refactor the code a bit so that this is clearer, e.g.:
```Python
# we cannot infer a eol date either from tomcat, nor from openjdk
if jre_major not in jre_end_support_dates and tomcat_major not in tomcat_end_support_dates:
    return None

# pick the lower eol of tomcat & openjdk
end_date = min(
        jre_end_support_dates.get(jre_major, datetime.date.max),
        tomcat_end_support_dates.get(tomcat_major, datetime.date.max),
    )
assert end_date < datetime.date.max
return end_date
```

dirkmueller: I did have it briefly documented in the function/method docstring, but it is expanded now to be more verbose.

PR #1438, opened 2024-07-18
Remove AppCollection class and insert the logic into Registry helper

dcermak:
This allows us to have not only application but also language & os containers in the application catalog. But I don't find the implementation satisfactory yet, especially the decorator for the title isn't exactly what I was wishing for.


src/bci_build/package/__init__.py
```diff
@@ -1428,15 +1502,15 @@ def __post_init__(self) -> None:
...
-        return "suse"
+        assert self.registry_type
+        return self.registry_type.application_prefix
 
     @property
     def image_type(self) -> ImageType:
         return ImageType.APPLICATION
 
     @property
+    @Registry.app_coll_title
```
dirkmueller, line 1513: I don't think such a mutating decorator makes understanding the code logic any easier. 

what about simply delegating that to a method in a RegistryClass (see the `def url(self, container)`) example in #1443  instead?

src/bci_build/package/__init__.py
```diff
@@ -554,6 +605,35 @@ def __post_init__(self) -> None:
...
+                f"Invalid registry for SLE 15 based container: {self.registry_type}"
+            )
+
+        if self.registry_type == Registry.APP_COLL:
+            # Limit Appcollection stuff to aarch64 and x86_64
+            if not self.exclusive_arch:
+                self.exclusive_arch = [Arch.AARCH64, Arch.X86_64]
+            if self.is_latest:
+                raise ValueError(
+                    "Application collection does not support the latest tag"
```
dirkmueller, line 634: this  check would have made a ton of sense in the `ApplicationCollectionContainer.__post_init__` rather than polluting our beautiful base class with some clumsy specialization specific sanity check IMHO.

src/bci_build/package/__init__.py
```diff
@@ -118,6 +118,51 @@ def __str__(self) -> str:
         return self.value
 
 
+@enum.unique
+class Registry(enum.Enum):
+    """Supported registries for container image distribution"""
+
+    SUSE_COM = "registry.suse.com"
+    APP_COLL = "dp.apps.rancher.io"
+    OPENSUSE_ORG = "registry.opensuse.org"
```
dirkmueller, line 127: why  do we need the  registry as enumerators?  are you enumerating (aka going over each one of them) anywhere? seems not the ideal abstraction to me.

src/bci_build/package/__init__.py
```diff
@@ -118,6 +118,51 @@ def __str__(self) -> str:
...
+
+    def __str__(self) -> str:
+        return self.value
+
+    @property
+    def application_prefix(self) -> str:
+        """Container image prefix for application containers."""
+        match self.value:
+            case Registry.SUSE_COM.value:
```
dirkmueller, line 136: this is what I personally dislike. tons of if/match statements that specialize behavior based on some magic attributes. we should delegate to a special object that has the right context rather than repeating in every property some match statement wars that doesn't make the code look elegant or easy to read.

src/bci_build/package/__init__.py
alexandrevicenzi, line 127: It would look better if it were:

`Registry.SUSE`, `Registry.RANCHER` (or `Registry.RANCHER_PRIME`) and `Registry.OPENSUSE`.

APP_COLL is truncated, and not understandable.
AFAIK, SUSE and openSUSE have only one public registry.
Rancher might have more, but I doubt Rancher Prime has any other.

src/bci_build/package/__init__.py
```diff
@@ -118,6 +118,51 @@ def __str__(self) -> str:
...
+                return "bci"
+            case Registry.APP_COLL.value:
+                return "containers"
+            case Registry.OPENSUSE_ORG.value:
+                return "opensuse/bci"
+
+    @staticmethod
+    def app_coll_title(title_property):
+        """Decorator that modifies the"""
```
alexandrevicenzi, line 156: nit1: app_coll_title? what does this mean? app_collection_title or registry_title is more meaningful.
nit2: modifies what?

src/bci_build/package/__init__.py
```diff
@@ -633,11 +713,11 @@ def lifecycle_url(self) -> str:
     @property
     def release_stage(self) -> ReleaseStage:
         """This container images' release stage.
-
-        It is :py:attr:`~ReleaseStage.RELEASED` if the container images'
-        operating system version is in the list
-        :py:const:`~bci_build.package.RELEASED_OS_VERSIONS`. Otherwise it
-        is :py:attr:`~ReleaseStage.BETA`.
+        n
```
alexandrevicenzi, line 716: nit: n

src/bci_build/package/__init__.py
```diff
@@ -694,8 +769,9 @@ def dockerfile_custom_end(self) -> str:
         return ""
 
     @property
+    @abc.abstractmethod
     def _registry_prefix(self) -> str:
-        return _build_tag_prefix(self.os_version)
+        """"""
```
alexandrevicenzi, line 774: nit:  `raise NotImplementedError`?

src/bci_build/package/__init__.py
alexandrevicenzi: This IMHO would look a bit better:

```python
from collections import namedtuple

RegistryInfo = namedtuple("RegistryInfo", ["title", "prefix", "devel_prefix"])

class Registry:
    SUSE = RegistryInfo(title="SUSE Registry", prefix="suse", devel_prefix="bci")
    RANCHER = RegistryInfo(title="Rancher Prime Application Collection", prefix="containers", devel_prefix="containers")
    OPENSUSE = RegistryInfo(title="openSUSE Registry", prefix="opensuse", devel_prefix="opensuse/bci")

print(f"{Registry.SUSE.title} - {Registry.SUSE.prefix}")
print(f"{Registry.RANCHER.title} - {Registry.RANCHER.prefix}")
print(f"{Registry.OPENSUSE.title} - {Registry.OPENSUSE.prefix}")
```

PR #1443, opened 2024-07-19
Add a Registry abstract base class with the three implementations

dirkmueller:
This removes a bit of if() condition wars in the BaseContainerImage and delegates it into its own class hierarchy.


dirkmueller:
not a finished/well tested implementation, just an alternative to the conversation the other PR.

src/bci_build/package/apache_tomcat.py
```diff
@@ -3,19 +3,22 @@
...
 assert _TOMCAT_VERSIONS == sorted(_TOMCAT_VERSIONS)
 
 TOMCAT_CONTAINERS = [
-    ApplicationCollectionContainer(
+    ApplicationStackContainer(
         name="apache-tomcat",
         pretty_name="Apache Tomcat",
+        _publish_registry=ApplicationCollectionRegistry()
+        if not os_version.is_tumbleweed
+        else None,
```
dirkmueller, line 21: this part can probably be done by a `get_appcollection_or_opensuse_registry()` helper somewhere to avoid repetitive ifs.

dcermak:
I believe this has now been merged. Can we close this @dirkmueller?

dirkmueller:
yes

PR #1446, opened 2024-07-19
Restore changes file from SP4 to make autobuild happy

dirkmueller:


dcermak (approved):
Completely unrelated image, but changes files never made sense…

PR #1454, opened 2024-07-25
Build the Tumbleweed base container also from dockerfile generator

dirkmueller:


dirkmueller:
(README fixes and testing pending)

PR #1473, opened 2024-07-30
Add Rust 1.80

dirkmueller:


dcermak (approved):
LGTM, note though, that it's broken on TW

dirkmueller:
> LGTM, note though, that it's broken on TW

the update is stuck in staging as it breaks stuff. I think we can ignore that for now?

PR #1482, opened 2024-08-01
Remove osc dependency from update-versions

dirkmueller:
This requires setting OSC_USER and OSC_PASSWORD which is just annoying and unnecessary. we can do this with a single interconnect api call just as well.


dcermak (changes requested):
Please don't. Instead use `osc = Osc(public=True)` instead so that we can rely on the error handling code inside py_obs

dirkmueller:
> Please don't. Instead use `osc = Osc(public=True)` instead so that we can rely on the error handling code inside py_obs

why do we need this error handling for this function? it just tries to inject authentication which is not necessary here, and then retries with fallback on error code 500. I don't find that sensible behavior for updating the versions. If there's error 5xx there is an overload and just blindly retrying is not useful imho for this functionality. 

also btw the current eventloop "loop/retrying" logic is failing in CI so the package_verisons are outdated, which is why I started to look into it.

dirkmueller:
obsolete, was changed via #1500

PR #1483, opened 2024-08-01
Allow to customize the oci image version label for AppCollection

dirkmueller:
OCI requires the full application verison in the oci.version label,
however when we have variants we need to suffix the primary tag
version with a variant suffix. split version_label into two
properties, tag_version and oci_version to refer to each.

tag_version then gains support for variants by adding an optional
verion_variant attribute to the container definion.

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -565,25 +565,23 @@ class BaseContainerImage(abc.ABC):
     @property
     @abc.abstractmethod
     def uid(self) -> str:
-        """unique identifier of this image, either its name or ``$name-$version``."""
+        """unique identifier of this image, either its name or ``$name-$tag_version``."""
         pass
 
     @property
     @abc.abstractmethod
-    def version_label(self) -> str:
+    def oci_version(self) -> str:
         """The "main" version label of this image.
 
         It is added as the ``org.opencontainers.image.version`` label to the
-        container image and also added to the
-        :py:attr:`~BaseContainerImage.build_tags`.
-
+        container image.
         """
         pass
 
     @property
     def build_name(self) -> str | None:
         if self.build_tags:
-            build_name = self.build_tags[0]
+            build_name: str = self.build_tags[0]
             if self.is_singleton_image:
                 build_name = build_name.partition(":")[0]
             return build_name.replace("/", ":").replace(":", "-")
@@ -1306,8 +1304,13 @@ exit 0
 
 @dataclass
 class DevelopmentContainer(BaseContainerImage):
-    #: the primary version of the language or application inside this container
+    #: the main version of the language or application inside this container
+    #: used for `org.opencontainers.image.version`
     version: str | int = ""
+
+    #: the version-$variant to use in the first build_tag. defaults to version
+    #: if not set.
+    tag_version: str | None = None
 
     # a rolling stability tag like 'stable' or 'oldstable' that will be added first
     stability_tag: str | None = None
@@ -1322,6 +1325,8 @@ class DevelopmentContainer(BaseContainerImage):
         super().__post_init__()
         if not self.version:
             raise ValueError("A language stack container requires a version")
+        if not self.tag_version:
+            self.tag_version = self.version
 
     @property
     def _registry_prefix(self) -> str:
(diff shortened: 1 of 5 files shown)


src/bci_build/package/apache_tomcat.py
```diff
@@ -52,52 +54,47 @@ def _get_sac_supported_until(
...
             (os_version in CAN_BE_LATEST_OS_VERSION)
             and tomcat_major == _TOMCAT_VERSIONS[-1]
             and jre_version == 22
             and os_version.is_tumbleweed
         ),
+        _oci_image_version=format_version(
+            tomcat_pkg_version := get_pkg_version(tomcat_pkg, os_version),
+            ParseVersion.PATCH,
+        ),
         version=f"{tomcat_major}-jre{jre_version}",
```
dcermak, line 76: Or we don't add the new property and instead implement this change, which should have the same consequence:
```suggestion
        version=tomcat_pkg_version,
```

dirkmueller: That doesn't work, as we need to add the jre suffix to make the build tag/buildversion unique across many flavors.

Also, it would change the buildversion on every minor version bump, which resets the release counter and means qa has to update their syncing and testing scripts. That seems very high overhead to me.

What we can do instead is add a version_flavor that is taken as a suffix for the tag and buildversion  if set. What do you think about that instead?

dirkmueller: I pushed a change that uses now a version_flavor variant but this is getting royally messy as "version" is used all over the place as a identifier and now I had to overload `version_in_uid` with a `str` to deduplicate the tomcat containers.

dirkmueller: actually the whole `version_flavor` mess was too complicated. now  `version` is the oci version and `tag_version` is what version used to be before. I hope this is a better approach, as it is fairly simple.

src/bci_build/package/__init__.py
```diff
@@ -1498,15 +1501,15 @@ def build_tags(self) -> list[str]:
         for name in [self.name] + self.additional_names:
             tags += [
                 f"{self._registry_prefix}/bci-{name}:%OS_VERSION_ID_SP%",
-                f"{self._registry_prefix}/bci-{name}:{self.version_label}",
+                f"{self._registry_prefix}/bci-{name}:{self.image_ref_name}",
```
dcermak, line 1504: Any particular reason why we shouldn't or don't want to use `oci_version` here?

src/bci_build/package/apache_tomcat.py
```diff
@@ -52,52 +54,44 @@ def _get_sac_supported_until(
...
+        + [f"java-{jre_version}-openjdk", f"java-{jre_version}-openjdk-headless"],
         os_version=os_version,
         is_latest=(
             (os_version in CAN_BE_LATEST_OS_VERSION)
             and tomcat_major == _TOMCAT_VERSIONS[-1]
             and jre_version == 22
             and os_version.is_tumbleweed
         ),
-        version=f"{tomcat_major}-jre{jre_version}",
+        version=(tomcat_pkg_version := get_pkg_version(tomcat_pkg, os_version)),
```
dcermak, line 72: I'm not opposed to using our version fetching mechanics, but it causes more busy work on our side instead of the buildservice handling this automatically. Do we have to prefetch the versions or can we rely on the replace_using_package_version service instead?

dcermak (commented):
Lgtm as it is, added just two questions

src/bci_build/package/__init__.py
dirkmueller: it's unlikely that `oci_version` gets reimplemented in a derived class. however `image_ref_name` might, because it is "fixed" tag that this image should be identified with. hence I think it makes sense to use that one as the primary tag, which has an important meaning for the build service (and the buildname/buildversion logic)

src/bci_build/package/apache_tomcat.py
dirkmueller: True, there was a reason why I converted that to the package version fetching logic but it no longer applies. reverted that part.

dcermak (approved):

PR #1484, opened 2024-08-01
Adding Kiwi stackbuild plugin for container image based builds

rdoxenham:
In this PR I'm seeking the addition of the python3-kiwi_stackbuild_plugin, but I'm not sure if it's available in the standard repositories.


dcermak (changes requested):
The `python3-kiwi_stackbuild_plugin` package is not available in any distribution. Where did you get the package name from?

alexandrevicenzi:
The package is not available in Factory, just in a devel project, see https://build.opensuse.org/package/show/Virtualization:Appliances:Builder/python-kiwi_stackbuild_plugin.

PR #1485, opened 2024-08-02
Use oci.image.authors instead of deprecated "MAINTAINER" setting

dirkmueller:


src/bci_build/templates.py
```diff
@@ -115,9 +114,10 @@
 {%- if image.entrypoint_user  %}
           user="{{ image.entrypoint_user }}"
 {%- endif %}
-          maintainer="{{ image.maintainer }}">
+{%- if True %}>{%- endif %}
```
dcermak, line 117: Am I missing something here?
```suggestion
```

dirkmueller: @dcermak we want to have ">" to close the element after the last attribute that is being added, without spaces or newlines. given that all attributes before are conditional, this is the only way I found it doing the thing properly in all the cases. 

if you have a better suggestion, I'm all ears.

dirkmueller: see that I extended the unit testing to have a image with an entrypoint user configured.

dcermak: OMG, I completely missed the `>` in between the tags :facepalm: 

Could you maybe add a jinja comment in there just shortly explaining it to prevent idiots like myself from wasting two hours on this?

dcermak (approved):

src/bci_build/templates.py
dirkmueller: I really don't think that highlighting minor nuances makes the code any easier to read or understand. imho the essential parts should be easy to grasp. 

but anyhow, in order to move forward I added a monstreous comment.

PR #1486, opened 2024-08-02
Revert "Bump aiohttp from 3.9.5 to 3.10.0"

dcermak:
Reverts SUSE/BCI-dockerfile-generator#1478

This update breaks py-obs which causes the update-versions action to fail. This is a temporary workaround until this is fixed upstream


dirkmueller:
Should we have a pin on `pyproject.toml` to avoid a next lock update to undo this?

dcermak:
In theory we don't have to, it should be fixed via https://github.com/dcermak/py-obs/pull/155

dirkmueller:
> In theory we don't have to, it should be fixed via [dcermak/py-obs#155](https://github.com/dcermak/py-obs/pull/155)

Ah, nice debugging. it would also not be necessary with #1482 because that serializes access rather than using massively parallel async io ;)

dcermak:
Closing in favor of #1500

PR #1502, opened 2024-08-06
Remove the oci.revision again

dirkmueller:
appcol no longer requires it, and we can not set it to a meaningful value.


dcermak (dismissed):
fine by me, just out of curiosity, what's this `%REVISION%` parameter?

dirkmueller:
> fine by me, just out of curiosity, what's this `%REVISION%` parameter?

It's the name that appcol uses for what we have as `%RELEASE%`. I've also converted this into  a draft as the discussion with Fabian led that there might be more issues lurking here.

dcermak (approved):

PR #1506, opened 2024-08-07
update logging framework in postfix SP6 image

Priyankasaggu11929:
For addressing failing SP6 tests, as per – https://github.com/SUSE/BCI-tests/pull/487#issuecomment-2270049197

PR updates logging framework from `syslog` to `rsyslog` in case of SP6

### DIFF
--- a/src/bci_build/package/postfix.py
+++ b/src/bci_build/package/postfix.py
@@ -71,7 +71,7 @@ POSTFIX_CONTAINERS = [
                     "spamass-milter",
                 )
                 if os_version == OsVersion.TUMBLEWEED
-                else ()
+                else ("rsyslog", "gawk")
             )
         ],
         entrypoint=["/entrypoint/entrypoint.sh"],
--- a/src/bci_build/package/postfix/entrypoint/entrypoint.sles.sh
+++ b/src/bci_build/package/postfix/entrypoint/entrypoint.sles.sh
@@ -358,7 +358,7 @@ configure_postfix() {
 
 terminate() {
     base=$(basename "$1")
-    pid=$(/bin/pidof "$base")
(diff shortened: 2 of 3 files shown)


Priyankasaggu11929:
will check again in the morning. 

SP6 build failing on rsyslog installation 

```
[   47s] Problem: 1: nothing provides '/bin/gawk' needed by the to be installed sysconfig-netconfig-0.85.9-150200.12.1.x86_64
[   47s]  Solution 1: do not install rsyslog-8.2306.0-150600.10.6.x86_64
[   47s]  Solution 2: break sysconfig-netconfig-0.85.9-150200.12.1.x86_64 by ignoring some of its dependencies
```

Priyankasaggu11929:
@dirkmueller, I've tried building the sp6 image without installing `rsyslog` as a Dockerfile step, but later manually in the built image (`registry.opensuse.org/home/defolos/bci/staging/sle-15-sp6/6-1506/containerfile/containers/postfix:3.8`) and installation works.

Could you please point out what is different in the OBS build tooling that causing the build fail for SP6 variant:
https://build.opensuse.org/package/live_build_log/home:defolos:BCI:Staging:SLE-15-SP6:6-1506/sac-postfix-image/containerfile/x86_64

dirkmueller:
> @dirkmueller,> Could you please point out what is different in the OBS build tooling that's causing the build fail for SP6 variant. Thanks!

Thats a missing Fileprovides on gawk. I've filed an issue to get that fixed, but for now we can fix this by manually installing gawk. updated your patch accordingly.

dirkmueller (approved):

Priyankasaggu11929:
> > @dirkmueller,> Could you please point out what is different in the OBS build tooling that's causing the build fail for SP6 variant. Thanks!
> 
> Thats a missing Fileprovides on gawk. I've filed an issue to get that fixed, but for now we can fix this by manually installing gawk. updated your patch accordingly.

Thank you!

PR #1507, opened 2024-08-07
MariaDB gosu - use exec and quote user

grooverdan:
Without exec, the mariadbd doesn't end up with PID 1 and this results in errors managing the container.

bash was also unneeded.

Closes https://bugzilla.opensuse.org/show_bug.cgi?id=1213154


dirkmueller (approved):
Oh wow, we screwed this up the 3rd time in a row! :-( thanks for the fix!

PR #1519, opened 2024-08-08
Make package_name attribute optional

dirkmueller:
If not overridden, simply use $name-image as default.


dcermak (commented):
This causes package renames of tomcat, java & postgres. I don't think we want that?

dirkmueller:
> This causes package renames of tomcat, java & postgres. I don't think we want that?

This is just the bot doing nonsense I think. I'll verify in a new PR

dirkmueller:
it works in the new PR

PR #1523, opened 2024-08-09
Fix fetching of the mariadb scripts

dirkmueller:
3264da304a5ed7fe3ff03936277d0e02d00d7b04 broke update-files becaues it now tried to fetch the files called "minor" from GitHub which do not exist. This would have been prevented by having versions in a separate object tree.


PVlasin (approved):
OK

PR #1526, opened 2024-08-09
Make package_name attribute optional

dirkmueller:
If not overridden, simply use $name-image as default.


dirkmueller:
continuation of #1519 hopefully without the bot doing nonsense.

PR #1541, opened 2024-08-14
[postfix] update entrypoint.sh to remove syslogd dependency

Priyankasaggu11929:
PR updates the entrypoint for postfix container image with (upstream) changes in https://github.com/thkukuk/containers-mailserver/commit/11d144f9fa673fada91786f5d457f17189bdacc8, which fixes:

```
 /entrypoint/entrypoint.sh: line 411: /usr/sbin/syslogd: No such file or directory
 ``` 

Following above change, another error is introduced:

```
/usr/lib/postfix/bin//postfix-script: line 400: cmp: command not found
```

which will be addressed by SR (postfix package): https://build.opensuse.org/request/show/1192813

### DIFF
--- a/src/bci_build/package/postfix/entrypoint/entrypoint.sles.sh
+++ b/src/bci_build/package/postfix/entrypoint/entrypoint.sles.sh
@@ -166,7 +166,6 @@ setup_submission() {
 	echo "Enable submission port"
 
 	echo "submission inet n       -       n       -       -       smtpd" >> /etc/postfix/master.cf
-	echo " -o syslog_name=postfix/submission" >> /etc/postfix/master.cf
 
 	if [ "${SMTPD_USE_TLS}" -eq "1" ]; then
 	    echo " -o smtpd_tls_security_level=encrypt" >> /etc/postfix/master.cf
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/bci_build/package/postfix/entrypoint/entrypoint.sles.sh [hunk 1]
```diff
@@ -427,11 +418,5 @@ configure_postfix
...
-if [ "$1" = 'postfix' ]; then
-    start_daemons "$@"
-    echo "postfix running and ready"
-    echo "[info] refer to postfix manual pages at https://www.postfix.org/postfix-manuals.html"
-    sleep infinity & wait $!
-else
-    exec "$@"
-fi
+exec "$@"
+echo "[info] refer to postfix manual pages at https://www.postfix.org/postfix-manuals.html"
```
thkukuk, line 422: This line will never be executed due to the exec before.

thkukuk (changes requested):

src/bci_build/package/postfix/entrypoint/entrypoint.sles.sh
dirkmueller: I swapped the two statements.

dirkmueller (approved):

PR #1590, opened 2024-09-02
Fix version labels to list the actual language version

dirkmueller:
the org.opencontainer.image.version label should list the actual full version number not our simplified categorization version. set that one as tag_version instead.

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -1448,17 +1448,18 @@ class DevelopmentContainer(BaseContainerImage):
     def build_version(self) -> str | None:
         build_ver = super().build_version
         if build_ver:
-            # if self.version is a numeric version and not a macro, then
+            container_version: str = self.tag_version
+            # if container_version is a numeric version and not a macro, then
             # version.parse() returns a `Version` object => then we concatenate
             # it with the existing build_version
             # for non PEP440 versions, we'll get an exception and just return
             # the parent's classes build_version
             try:
-                version.parse(str(self.version))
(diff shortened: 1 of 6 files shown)


dcermak (changes requested):
- Breaks buildversion
- Breaks the golang readme (the go-race package now contains a regex that will not be replaced and the -openssl readme got truncated).

dirkmueller:
> * Breaks buildversion
> * Breaks the golang readme (the go-race package now contains a regex that will not be replaced and the -openssl readme got truncated).
should be all fixed

dcermak (commented):
Still causes a regression in kiwi's `BuildVersion` https://github.com/SUSE/BCI-dockerfile-generator/commit/e03a35aa9505de4b8d11881c867314095c04ee69#diff-b37d2147bb42fa8e2dffe851034eea1eca592a3914440e6d62c45957b1098e17

dirkmueller:
> Still causes a regression in kiwi's `BuildVersion` [e03a35a#diff-b37d2147bb42fa8e2dffe851034eea1eca592a3914440e6d62c45957b1098e17](https://github.com/SUSE/BCI-dockerfile-generator/commit/e03a35aa9505de4b8d11881c867314095c04ee69#diff-b37d2147bb42fa8e2dffe851034eea1eca592a3914440e6d62c45957b1098e17)

it's a change, not a regression. yes, it's a downgrade but I don't think it matters since we haven't released it for SP5 yet, and the buildversion is only used to deconflict SPs. 

but I can add yet another workaround on top of the workarounds for this

PR #1603, opened 2024-09-05
set the full mariadb version in labels/metadata

dirkmueller:


alexandrevicenzi (commented):
There's a build failure.

%%mariadb_version%% appears to be missing substitution

PR #1625, opened 2024-09-09
Set the full versions on the dotnet containers

dirkmueller:


src/bci_build/package/dotnet-sdk/README.md.j2
```diff
@@ -55,7 +55,7 @@ There are situations, where you don't want to run an application inside a contai
 To compile the application without running it inside a container, use the following command:
 
 ```ShellSession
-podman run --rm -v "$PWD":/app:Z -w /app {{ image.pretty_reference }} dotnet run
+podman run --rm -v "$PWD":/app:Z -w /app {{     pretty_reference }} dotnet run
```
dcermak, line 58: This looks like a mistake

dirkmueller: reverted

dirkmueller:
this is not working. needs a different approach.

src/dotnet/updater.py
```diff
@@ -162,26 +164,51 @@ class DotNetBCI(DevelopmentContainer):
     def __post_init__(self):
         if OsVersion.TUMBLEWEED == self.os_version:
             raise ValueError(".Net BCIs are not supported for openSUSE Tumbleweed")
+
+        # import pdb; pdb.set_trace()
```
dcermak, line 168:
```suggestion
```

src/dotnet/updater.py
```diff
@@ -317,25 +340,11 @@ def _guess_version_from_pkglist(self, pkg_list: list[RpmPackage]) -> str | None:
...
-                conf=DotNetBCI._base.conf,
-                baseurl=(MS_REPO_BASEURL,),
-            )
-            DotNetBCI._base.fill_sack()
-
-        pkgs = self._fetch_packages()
-
-        if new_version := self._guess_version_from_pkglist(pkgs):
-            assert not self.additional_versions, f"additional_versions property must be unset, but got {self.additional_versions}"
-            self.additional_versions = [new_version]
```
dcermak, line 333: This is a very costly operation, moving it to `__post_init__()` will cause it to get executed on every import. This will also slow down the bot considerably (just calling `scratch_build_bot --help` takes five times longer on this branch). I'd greatly appreciate if we could keep this costly part here and not move it into the constructor.

dirkmueller: It's not cleanly possible. DevelopmentContainer's __post_init__ barfs when self.version is None, so it has to be determined at init time. 

The alternative is that we set self.version to some random garbage value and fix it up later. Which I didn't go for because I think that'll just be causing hard to trace subtle bugs later on. 

The 3rd alternative is that we make update-versions determine the dotnet package versions and store it in package_version.json instead.

dirkmueller: which of the options do you prefer? don't have time to implement all of them.

alexandrevicenzi: I would go with `package_version.json` if it is easy to implement.

dirkmueller: I wasn't able to get this done easily, the package version logic is too tied up with OBS. so I added a `prepare_generation` hook that can be used to generate expensive things outside initalization and before generating the templates.

src/bci_build/package/__init__.py
```diff
@@ -1343,10 +1351,21 @@ class DevelopmentContainer(BaseContainerImage):
...
+
         if not self.version:
-            raise ValueError("A language stack container requires a version")
+            raise ValueError("A development container requires a version")
+
         if not self.tag_version:
-            self.tag_version = self.version
+            raise ValueError("A development container requires a tag_version")
+
+        pass
```
dcermak, line 1368:
```suggestion
```

src/bci_build/package/__init__.py
```diff
@@ -1343,10 +1351,21 @@ class DevelopmentContainer(BaseContainerImage):
...
+        """Hook to do delayed expensive work prior template rendering"""
+        super().prepare_generation()
+
         if not self.version:
-            raise ValueError("A language stack container requires a version")
+            raise ValueError("A development container requires a version")
+
         if not self.tag_version:
-            self.tag_version = self.version
+            raise ValueError("A development container requires a tag_version")
```
dcermak, line 1366: Imho this can stay in `__post_init__`. This check is super cheap and should fail as early as possible

src/bci_build/package/__init__.py
```diff
@@ -586,6 +586,12 @@ def __post_init__(self) -> None:
         ):
             self.support_level = SupportLevel.TECHPREVIEW
 
+    @abc.abstractmethod
+    def prepare_generation(self) -> None:
```
alexandrevicenzi, line 590: nit: prepare generation of what? this name feels incomplete, but I'm not sure how to improve also. Perhaps `prepare_template_generation` or similar.

dirkmueller: renamed to prepare_template

dirkmueller:
> I like the new way, unfortunately it completely messes up the .Net dockerfiles: [148a304#diff-7e4e897e93d17a79d1b7a44f88ee11684a9dcad7ce1aefe5f2395aeeb475e929](https://github.com/SUSE/BCI-dockerfile-generator/commit/148a304676322f3f0bdc93456373d433c16d2555#diff-7e4e897e93d17a79d1b7a44f88ee11684a9dcad7ce1aefe5f2395aeeb475e929)

ah, always the last untested change that messes it up. fixed.

PR #1632, opened 2024-09-10
set full version on the PHP containers

dirkmueller:


src/bci_build/package/templates/php_common.j2
```diff
@@ -38,13 +38,14 @@ RUN set -euo pipefail; \
...
 
 The following scripts ship with the image to keep it compatible with the
 DockerHub image: `docker-php-source`, `docker-php-ext-configure`,
-`docker-php-ext-enable`, and `docker-php-ext-install`. Note that only
-`docker-php-ext-install` performs an actual job. None of the other scripts
-require to be executed in the image.
+`docker-php-ext-enable`, and `docker-php-ext-install`.
+The script `docker-php-ext-install` performs an actual job, all others are
+just no-operation scripts for interoperability.
+
```
dcermak, line 51:
```suggestion
```

src/bci_build/package/php/README.md.j2
```diff
@@ -1,4 +1,4 @@
-# The PHP {{ image.version }} container image
+# The PHP {{ image.tag_version }} container image
```
dcermak, line 1: This is now a regex in the rendered version

dcermak (changes requested):
README shouldn't contain a regex, otherwise lgtm

src/bci_build/package/php/README.md.j2
dirkmueller: no, that was php-fpm/README.md.j2 missing a fix. fixed now also.

dcermak (approved):

PR #1635, opened 2024-09-10
Replace crane call with a pythonic alternative that also finds tags

dcermak:
This improves the output from the staging bot a bit as it now includes tags as well

### DIFF
--- a/src/staging/bot.py
+++ b/src/staging/bot.py
@@ -1,5 +1,6 @@
 import asyncio
 import itertools
+import json
 import os
 import random
 import string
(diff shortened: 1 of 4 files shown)


src/staging/bot.py
```diff
@@ -1089,6 +1089,56 @@ async def write_underscore_config() -> list[str]:
...
         return flattened_file_list
 
+    def fetch_container_urls(self) -> list[str]:
+        """Fetch the full urls to all published containers in the staging
+        project including their most relevant tag (either ``latest`` or the
+        first tag). Querying a staging project without published images results
+        in a ``RuntimeError``.
+
+        """
+        import json
```
dirkmueller, line 1099: maybe move to the top of the module?

src/staging/bot.py
```diff
@@ -1089,6 +1089,56 @@ async def write_underscore_config() -> list[str]:
...
+
+        from oras.client import OrasClient
```
dirkmueller, line 1101: also this?

dirkmueller (approved):

PR #1641, opened 2024-09-11
Use bci-base as builder container for SAC containers

dirkmueller:


src/bci_build/package/appcollection.py
```diff
@@ -34,6 +36,15 @@ def title(self) -> str:
             return super().title
         return self.pretty_name
 
+    @property
+    def _from_image(self) -> str | None:
+        if self.os_version.is_tumbleweed:
+            return super()._from_image
+        if self.os_version == OsVersion.SLE16_0:
+            return f"{_build_tag_prefix(self.os_version)}/bci-base:{self.os_version}"
```
dcermak, line 44: `super()._from_image` is `f"{_build_tag_prefix(self.os_version)}/bci-base:{self.os_version}"` for SLE 16 too:
```suggestion
        if self.os_version.is_tumbleweed or self.os_version == OsVersion.SLE16_0:
            return super()._from_image
```

alexandrevicenzi (approved):

dirkmueller:
this was prior approved already, only fixed the ruff error..

PR #1642, opened 2024-09-11
Make the last stage a fully qualified reference to the registry

dirkmueller:
Although "FROM target" works, it does not generate the necessary annotations as buildah/podman can not trace it back to the original layer.

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -686,6 +686,13 @@ class BaseContainerImage(abc.ABC):
         return "SUSE LLC"
 
     @property
+    def base_image_registry(self) -> str:
+        """The registry where the base image is available on."""
+        if self.os_version.is_tumbleweed:
+            return "registry.opensuse.org"
+        return "registry.suse.com"
+
+    @property
     def registry(self) -> str:
         """The registry where the image is available on."""
         if self.os_version.is_tumbleweed:
@@ -823,7 +830,7 @@ exit 0
 
         if self.from_target_image:
             return (
-                f"FROM {self.from_target_image} AS target\n"
+                f"FROM {self.base_image_registry}/{self.from_target_image} AS target\n"
                 f"FROM {self._from_image} AS builder"
             )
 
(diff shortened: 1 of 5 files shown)


dcermak:
This gives a whole bunch of unresolvable errors on SLE 15 SP7 & Tumbleweed. Since when can OBS handle full registry urls anyway?

dirkmueller:
> This gives a whole bunch of unresolvable errors on SLE 15 SP7 & Tumbleweed.

SP7: because we haven't released the SP7 base containers yet. 
Tumbleweed: because the staging setup is broken. Added fixes. 

> Since when can OBS handle full registry urls anyway?
it always could, why?

dirkmueller:
The Tumbleweed tests are failing on a have-choice error because we're having the containers twice in the path. 

basically we need this in the prjconf, but I don't have permission to change it: 

```
# Prefer the released image from the registry
Prefer: -container:micro-image -container:opensuse-tumbleweed-image-docker

```

dirkmueller:
okay, there was a bot PR that I could merge. let's see if it works now..

dcermak:
Are you sure that the resulting dockerfile should now look like this?
```
FROM bci/bci-micro:15.6 AS target
FROM bci/bci-base:15.6 AS builder
COPY --from=target / /target

RUN set -euo pipefail; zypper --installroot /target --gpg-auto-import-keys  -n in --no-recommends tomcat10 curl sed java-21-openjdk java-21-openjdk-headless; zypper -n clean; rm -rf /var/log/{lastlog,tallylog,zypper.log,zypp/history,YaST2}
FROM registry.suse.com/bci/bci-micro:15.6
COPY --from=builder /target /
```
(the last `FROM` used to be a `from target`)

Now we have two target images and we're copying stuff from the bci-micro on OBS/IBS into the micro image from registry.suse.com.

dirkmueller:
> Are you sure that the resulting dockerfile should now look like this?
> 
> ```
> FROM bci/bci-micro:15.6 AS target
> FROM bci/bci-base:15.6 AS builder
> COPY --from=target / /target
> 
> RUN set -euo pipefail; zypper --installroot /target --gpg-auto-import-keys  -n in --no-recommends tomcat10 curl sed java-21-openjdk java-21-openjdk-headless; zypper -n clean; rm -rf /var/log/{lastlog,tallylog,zypper.log,zypp/history,YaST2}
> FROM registry.suse.com/bci/bci-micro:15.6
> COPY --from=builder /target /
> ```
> 
> (the last `FROM` used to be a `from target`)
> 
> Now we have two target images and we're copying stuff from the bci-micro on OBS/IBS into the micro image from registry.suse.com.

it is "always r.s.c for target" and "always unqualified for builder image" as discussed offline now.

PR #1643, opened 2024-09-11
Set oci base image reference and digest annotations

dirkmueller:
waiting for https://build.opensuse.org/request/show/1200158

### DIFF
--- a/src/bci_build/templates.py
+++ b/src/bci_build/templates.py
@@ -74,6 +74,10 @@ LABEL com.suse.lifecycle-url="{{ image.lifecycle_url }}"
 LABEL com.suse.release-stage="{{ image.release_stage }}"
 {%- endif %}
 # endlabelprefix
+{%- if image.is_base_container_annotation_available %}
+LABEL org.opencontainers.image.base.name="%BASE_REFNAME%"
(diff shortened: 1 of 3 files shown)


src/bci_build/templates.py
```diff
@@ -72,6 +72,10 @@
 LABEL com.suse.release-stage="{{ image.release_stage }}"
 {%- endif %}
 # endlabelprefix
+{%- if image.dockerfile_from_line %}
+LABEL org.opencontainers.image.base.name="%BASE_REFNAME%"
+LABEL org.opencontainers.image.base.digest="%BASE_DIGEST%"
+{%- endif %}
```
dcermak, line 78: Any reason why you put it below the labelprefix section?

dirkmueller: the only reason is that I hate that label duplication for oci defined standard tags with passion. 

You can recursively look up the base.digests though. e.g. the pcp image which stacks on init-image (and base.digest refers to which one) and then you use the init-image's base.digest to look up the underlying base image. 

So I think reducing the number of pointless duplicated labels (given that these should be annotations to begin with) was the main driving force for me.

alexandrevicenzi (approved):

PR #1658, opened 2024-09-13
Install sles-release with coreutils

dirkmueller:
the sles-release on sle15 has a requires on coreutils or busybox-coreutils. We don't want to default to busybox so explicitly require coreutils.

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -261,8 +261,6 @@ class OsVersion(enum.Enum):
             return ("SLES-release",)
         if self.is_ltss:
             return ("sles-ltss-release",)
-        # if self.is_slcc:
-        #     return (f"{str(self.value).lower()}-release",)
 
         assert self.is_sle15
         return ("sles-release",)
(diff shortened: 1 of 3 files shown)


alexandrevicenzi:
It seems this causes a conflict in some images because both coreutils are being provided.

Also, does this change increase the size of the image considerably or just a tiny bit?

src/bci_build/package/__init__.py
```diff
@@ -261,11 +261,9 @@ def release_package_names(self) -> tuple[str, ...]:
             return ("SLES-release",)
         if self.is_ltss:
             return ("sles-ltss-release",)
-        # if self.is_slcc:
-        #     return (f"{str(self.value).lower()}-release",)
 
         assert self.is_sle15
-        return ("sles-release",)
+        return ("sles-release", "coreutils")
```
dcermak, line 266: The `coreutils` package is added to the busybox container too, where it causes conflicts

dcermak (changes requested):

src/bci_build/package/__init__.py
dirkmueller: ok, then we do this only for the fips containers, all the others anyway already had coreutils in the list

msmeissn:
FWIW we had unresolved also for the dotnet containers for this

dirkmueller:
> FWIW we had unresolved also for the dotnet containers for this

added there as well.

dcermak (approved):

PR #1669, opened 2024-09-13
remove minor versions for tomcat

dirkmueller:
there should be "branch" (which is major version) and fullversion-flavor.

### DIFF
--- a/src/bci_build/package/apache_tomcat.py
+++ b/src/bci_build/package/apache_tomcat.py
@@ -6,18 +6,18 @@ from bci_build.package import CAN_BE_LATEST_OS_VERSION
 from bci_build.package import DOCKERFILE_RUN
 from bci_build.package import OsContainer
 from bci_build.package import OsVersion
-from bci_build.package import ParseVersion
 from bci_build.package import Replacement
 from bci_build.package import _build_tag_prefix
 
 from .appcollection import ApplicationCollectionContainer
 
-_TOMCAT_VERSIONS: list[int] = [9, 10]
-assert _TOMCAT_VERSIONS == sorted(_TOMCAT_VERSIONS)
+# last version needs to be the newest
+_TOMCAT_VERSIONS: list[str] = ["9", "10.1"]
+assert _TOMCAT_VERSIONS == sorted(_TOMCAT_VERSIONS, key=float)
 
 
 def _get_sac_supported_until(
-    os_version: OsVersion, tomcat_major: int, jre_major: int
+    os_version: OsVersion, tomcat_ver: str, jre_major: int
 ) -> datetime.date:
     """Return the predicted minimum end of support date for this os/tomcat/jre combination. We pick
     the minimum time that either the given tomcat or JRE is known to be supported."""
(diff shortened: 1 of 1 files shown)


src/bci_build/package/apache_tomcat.py
```diff
@@ -91,19 +90,14 @@ def _get_sac_supported_until(
...
-            ),
         ],
         cmd=[
             f"/usr/{'libexec' if os_version in (OsVersion.TUMBLEWEED, OsVersion.SLE16_0) else 'lib'}/tomcat/server",
             "start",
         ],
         exposes_tcp=[8080],
         env={
-            "TOMCAT_MAJOR": tomcat_major,
+            "TOMCAT_MAJOR": math.floor(tomcat_ver),
```
alexandrevicenzi, line 100: nit: `int(tomcat_ver)` achieves the same thing and does not require an extra library import.

src/bci_build/package/apache_tomcat.py
```diff
@@ -121,12 +115,12 @@ def _get_sac_supported_until(
         entrypoint_user="tomcat",
         logo_url="https://tomcat.apache.org/res/images/tomcat.png",
     )
-    for tomcat_major, os_version, jre_version in (
-        (10, OsVersion.TUMBLEWEED, 22),
-        (10, OsVersion.TUMBLEWEED, 21),
-        (10, OsVersion.TUMBLEWEED, 17),
+    for tomcat_ver, os_version, jre_version in (
```
alexandrevicenzi, line 118: if you are hardcoding versions already, why not include `tomcat_ver_major` and avoid any math conversions?

dcermak (commented):
Floats are imho not the right way to handle version numbers due to potential representation & rounding issues. I'd suggest to use either strings or instances of [`packaging.Version`](https://packaging.pypa.io/en/stable/version.html)

src/bci_build/package/apache_tomcat.py
dirkmueller: deriving it from one value is more DRY than duplicating it in the tuple, imho.

src/bci_build/package/apache_tomcat.py
```diff
@@ -6,18 +6,17 @@
...
 
-_TOMCAT_VERSIONS: list[int] = [9, 10]
-assert _TOMCAT_VERSIONS == sorted(_TOMCAT_VERSIONS)
+# last version needs to be the newest
+_TOMCAT_VERSIONS: list[str] = ["9", "10.1"]
 
 
 def _get_sac_supported_until(
-    os_version: OsVersion, tomcat_major: int, jre_major: int
+    os_version: OsVersion, tomcat_major: str, jre_major: int
```
dcermak, line 19: nit: this is no longer only the major version (needs more adjustment down the line too)
```suggestion
    os_version: OsVersion, tomcat_ver: str, jre_major: int
```

dcermak (commented):
LGTM, only have one nitpick

src/bci_build/package/apache_tomcat.py
dirkmueller: all references renamed.

PR #1670, opened 2024-09-13
remove %RELEASE% suffixed tags for the additional_versions

dirkmueller:
This partially reverts cb350c0b2f9efbb7694ed00f2cbf913fa49a88ca which didn't really have a description on why it was done. Removing them is a requisite for appcol listing.

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -1444,11 +1444,14 @@ class DevelopmentContainer(BaseContainerImage):
             ver_labels: list[str] = [self.tag_version]
             if self.stability_tag:
                 ver_labels = [self.stability_tag] + ver_labels
-            for ver_label in ver_labels + self.additional_versions:
+            for ver_label in ver_labels:
                 tags += [f"{self._registry_prefix}/{name}:{ver_label}"]
                 tags += [
                     f"{self._registry_prefix}/{name}:{ver_label}-{self._release_suffix}"
                 ]
+            for ver_label in self.additional_versions:
+                tags += [f"{self._registry_prefix}/{name}:{ver_label}"]
+
             if self.is_latest:
                 tags += [f"{self._registry_prefix}/{name}:latest"]
         return tags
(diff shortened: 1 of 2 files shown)


dcermak (changes requested):
I'm not opposed to not appending `%RELEASE%`, but atm it leads to rather useless tags:
```diff
#!BuildTag: bci/kiwi:10
#!BuildTag: bci/kiwi:10-%RELEASE%
#!BuildTag: bci/kiwi:10.1
-#!BuildTag: bci/kiwi:10.1-%RELEASE%
#!BuildTag: bci/kiwi:10.1.8
-#!BuildTag: bci/kiwi:10.1.8-%RELEASE%
```
IMHO, it really only makes sense to have `bci/kiwi:10.1.8-%RELEASE%`, all the other tags have a very limited usefulness. If we have to get rid of this "shoot the fly with the cannon approach", then I'd go all in and just drop the `-%RELEASE%` suffix entirely.

dirkmueller:
> I'm not opposed to not appending `%RELEASE%`, but atm it leads to rather useless tags:
> 
> ```diff
> #!BuildTag: bci/kiwi:10
> #!BuildTag: bci/kiwi:10-%RELEASE%
> #!BuildTag: bci/kiwi:10.1
> -#!BuildTag: bci/kiwi:10.1-%RELEASE%
> #!BuildTag: bci/kiwi:10.1.8
> -#!BuildTag: bci/kiwi:10.1.8-%RELEASE%
> ```
> 
> IMHO, it really only makes sense to have `bci/kiwi:10.1.8-%RELEASE%`, all the other tags have a very limited usefulness. If we have to get rid of this "shoot the fly with the cannon approach", then I'd go all in and just drop the `-%RELEASE%` suffix entirely.

this is because "version" is not set correctly (it should be the full version). the tags will get better with the other PRs related to this getting merged.

dcermak (approved):

alexandrevicenzi (approved):

dirkmueller:
(this still needs a better determination what is the most "fixed" tag to attach a -%RELEASE suffix to)

PR #1671, opened 2024-09-13
Set the full version as version for the gcc containers

dirkmueller:


dcermak (commented):
This removes the option to pull the gcc container by the full gcc version:
```Diff
-#!BuildTag: bci/gcc:%%gcc_version%%
-#!BuildTag: bci/gcc:%%gcc_version%%-%RELEASE%
```

Do we really want that? I know we have no formal requirements about versioning, but this looks like a bit of a usability regression to me

dirkmueller:
we do need the full version-%release "immutable" tag for org.opencontainers.image.ref.name label. currently this is a bit messed up because we use tag_version for two entirely different purposes (one to group branches, and one to define this full version).

src/bci_build/package/gcc.py
```diff
@@ -76,13 +77,12 @@ def _is_main_gcc(os_version: OsVersion, gcc_version: _GCC_VERSIONS) -> bool:
...
             Replacement(
-                regex_in_build_description="%%gcc_version%%",
+                regex_in_build_description="%%gcc_minor_version%%",
                 package_name=gcc_pkg,
                 parse_version=ParseVersion.MINOR,
-            )
+            ),
         ],
-        env={"GCC_VERSION": "%%gcc_version%%"},
-        additional_versions=["%%gcc_version%%"],
```
dcermak, line 85: Can't we just resurrect this line as follows? That should solve the missing build tags
```Python
        additional_versions=["%%gcc_minor_version%%"],
```

dirkmueller: done.

PR #1673, opened 2024-09-13
Use full version in version label for openjdk containers

dirkmueller:


src/bci_build/package/openjdk.py
```diff
@@ -44,6 +46,16 @@ def _get_openjdk_kwargs(
             # prevent ftbfs on workers with a root partition with 4GB
             "_constraints": generate_disk_size_constraints(6)
         },
+        "replacements_via_service": [
+            Replacement(
+                regex_in_build_description="%%java_version%%",
+                package_name=(
+                    f"java-{java_version}-openjdk-devel"
+                    if devel
+                    else f"java-{java_version}-openjdk"
```
dcermak, line 55: Isn't `f"java-{java_version}-openjdk"` not in both the devel & non-devel image?

dcermak (changes requested):
This leads to us having a regex in a readme again:
```diff
  # Bundle the application into OpenJDK runtime image
- FROM registry.suse.com/bci/openjdk:21
+ FROM registry.suse.com/bci/openjdk:%%java_version%%
```

src/bci_build/package/openjdk.py
dirkmueller: it is, but do we want to rely on that fact?

dirkmueller:
> This leads to us having a regex in a readme again:

Fixed.

dcermak (approved):

PR #1674, opened 2024-09-16
Make image.reference optional and hide it for appcollection

dirkmueller:
It was requested to not refer to opensuse in the Rancher AppCol.

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -1028,7 +1028,7 @@ exit 0
 
     @property
     @abc.abstractmethod
-    def reference(self) -> str:
+    def reference(self) -> str | None:
         """The primary URL via which this image can be pulled. It is used to set the
         ``org.opensuse.reference`` label and defaults to
         ``{self.registry}/{self.image_ref_name}``.
@@ -1435,7 +1435,7 @@ class DevelopmentContainer(BaseContainerImage):
         return f"{self.tag_version}-{self._release_suffix}"
 
     @property
-    def reference(self) -> str:
+    def reference(self) -> str | None:
         return (
             f"{self.registry}/{self._registry_prefix}/{self.name}:{self.image_ref_name}"
         )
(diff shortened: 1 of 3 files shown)


src/bci_build/package/appcollection.py
```diff
@@ -42,6 +42,12 @@ def _from_image(self) -> str | None:
 
         return f"bci/bci-base:15.{self.os_version}"
 
+    @property
+    def reference(self) -> str | None:
+        if self.os_version.is_tumbleweed:
+            return super().reference
+        return None
```
alexandrevicenzi, line 49: tecnically, `{%- if image.reference %}` also works if you return an empty string, no need to change the return type of this function.

src/bci_build/package/__init__.py
```diff
@@ -1028,7 +1028,7 @@ def image_ref_name(self) -> str:
 
     @property
     @abc.abstractmethod
-    def reference(self) -> str:
+    def reference(self) -> str | None:
         """The primary URL via which this image can be pulled. It is used to set the
         ``org.opensuse.reference`` label and defaults to
         ``{self.registry}/{self.image_ref_name}``.
```
dcermak, line 1034:
```suggestion
    def reference(self) -> str | None:
        """The primary URL via which this image can be pulled. It is used to set the
        ``org.opensuse.reference`` label and defaults to
        ``{self.registry}/{self.image_ref_name}``.
        
        The ``org.opensuse.reference`` label is omitted, if ``None`` is returned.
```

src/bci_build/package/appcollection.py
dirkmueller: That's true. I'm not particularly emotional about it, do you prefer this to be reverted and just use empy string for omitting?

alexandrevicenzi: The only reason to use an empty string is that it would require less refactoring as you don't change the type. I leave it up to you, but an empty string might be better.

dirkmueller:
we discussed this that this is a non-issue, so no longer needed.

PR #1675, opened 2024-09-16
Rename -jre flavor to openjdk flavor

dirkmueller:


src/bci_build/package/apache_tomcat.py
```diff
@@ -66,13 +66,13 @@ def _get_sac_supported_until(
             and os_version.is_tumbleweed
         ),
         version="%%tomcat_version%%",
-        tag_version=f"{tomcat_major}-jre{jre_version}",
+        tag_version=f"{tomcat_major}-openjdk{jre_version}",
```
alexandrevicenzi, line 69: nit: rename `jre_version` to `jdk_version` since we are talking about JDK and not JRE.

alexandrevicenzi (approved):

src/bci_build/package/apache_tomcat.py
dirkmueller: well, we still only install the JRE not the JDK (but we name the tag suffix jdk which I find a bit confusing as well)..

dcermak: Why are you proposing to change the tag suffix then if it's now more confusing?

alexandrevicenzi: the package list is `f"java-{jre_version}-openjdk", f"java-{jre_version}-openjdk-headless"`, is this JRE only? I thought it was JDK

dirkmueller: no, the jdk is the `java-{jre_version}-openjdk-devel`

PR #1685, opened 2024-09-17
Only disable maintainer listing on SLE

dcermak:
For openSUSE we won't be publishing to apps.rancher.io, so we can keep this label for registry.opensuse.org


alexandrevicenzi (approved):

dirkmueller (approved):
thanks, sorry for this oversight. we do need to fix the code eventually..

PR #1686, opened 2024-09-17
Mount /target/proc for zypper --installroot /target

Vogtinator:
Fix errors like

Running post-transaction scripts [..
%posttrans(java-21-openjdk-headless-21.0.4.0-150600.3.3.1.x86_64) script output: /usr/sbin/update-ca-certificates: line 83: /dev/fd/62: No such file or directory .done]


Vogtinator:
Turns out there is already some discussion about this at https://bugzilla.suse.com/show_bug.cgi?id=1219736

Looks like it's currently unfixable :disappointed:

PR #1687, opened 2024-09-17
Add Postgres volume notes

alexandrevicenzi:
initdb requires the volume location to be empty, otherwise it fails to create a new database.

These notes are due to bsc#1230393.

### DIFF
--- a/src/bci_build/package/postgres/README.md.j2
+++ b/src/bci_build/package/postgres/README.md.j2
@@ -24,6 +24,10 @@ $ podman run -it --rm -p 5432:5432 -e POSTGRES_PASSWORD=my-password -v /path/to/
 
 PostgreSQL data directory location.
 
+**Note 1:** The directory must be empty for `initdb` to create a new database.
+
+**Note 2:** If the volume points either to a file system mount point, a remote folder that cannot be owned by the `postgres` user, or a location that already contains files (including `lost+found` and dotfiles), a new subdirectory for storing the PostgreSQL data must be created within the `PGDATA` volume.
+
 ## Environment variables
 
 The PostgreSQL image uses several environment variables to configure the database initialization.


alexandrevicenzi:
Because `/var/lib/pgsql/data` is a volume, we can't mount a volume to `/var/lib/pgsql` as it won't sync. Is there a way around this? Should we change this volume?

dirkmueller:
just wondering, there's no way around this limitation? seems like easy enough to fix in initdb

alexandrevicenzi:
> just wondering, there's no way around this limitation? seems like easy enough to fix in initdb

There's a check hardcoded that is over 10y old in the code that you can see [here](https://github.com/postgres/postgres/blob/REL_16_4/src/port/pgcheckdir.c#L33).

In theory, removing this check should fix it, but I believe it is there for a reason.

src/bci_build/package/postgres/README.md.j2
```diff
@@ -24,6 +24,10 @@ $ podman run -it --rm -p 5432:5432 -e POSTGRES_PASSWORD=my-password -v /path/to/
 
 PostgreSQL data directory location.
 
+**Note 1:** `initdb` requires this location to be empty to create a new database.
```
dmpop, line 27:
```suggestion
**Note 1:**  The directory must be empty for `initdb` to create a new database.
```

src/bci_build/package/postgres/README.md.j2
```diff
@@ -24,6 +24,10 @@ $ podman run -it --rm -p 5432:5432 -e POSTGRES_PASSWORD=my-password -v /path/to/
...
+
+**Note 2:** If the volume points to a file system mount point, a remote folder that can't be chowned to the `postgres` user, or a location that already contains files (including `lost+found` and dotfiles), a new subdirectory must be created within the select storage method to contain the PostgreSQL data.
```
dmpop, line 29:
```suggestion
**Note 2:** If the volume points either to a file system mount point, a remote folder that cannot be owned by the `postgres` user, or a location that already contains files (including `lost+found` and dotfiles), a new subdirectory for storing the PostgreSQL data must be created within the selected storage method.
```

dmpop: _within the selected storage method_ sounds strange to me. Can you really create a subdir **within** a storage method?

alexandrevicenzi: Within in this case is inside, switching to inside sounds odd to me:

> a new subdirectory for storing the PostgreSQL data must be created inside the selected storage method.

The issue in note 2 is the following:

- Inside the container all PSQL sees is a folder, PSQL is unaware that this is a volume and neither what is behind this volume
- PSQL can't use the current folder because it has some issue (e.g. not empty location)
- For PSQL, the solution is to mount something as in PGDATA
- For the user, one solution is to create a new sub folder in the volume they are trying to use

The idea here is to document that certain things can't be used as PGDATA, we don't really need to provide the solution, as the solution could vary per setup.

The upstream Docker Hub image uses a similar text.

alexandrevicenzi: @dmpop does my previous message explains within?

dirkmueller: can we simply say "within the PGDATA volume"?

alexandrevicenzi: I went with both suggestions, please take a look @dirkmueller @dmpop.

dirkmueller (approved):

PR #1692, opened 2024-09-19
explicitly install SLES-release in containers that have a product choice

dirkmueller:


src/bci_build/package/basecontainers.py
```diff
@@ -63,7 +63,7 @@
         is_latest=os_version in CAN_BE_LATEST_OS_VERSION,
         pretty_name=f"{os_version.pretty_os_version_no_dash} Init",
         custom_description="Systemd environment for containers {based_on_container}. {podman_only}",
-        package_list=["systemd", "gzip"],
+        package_list=["systemd", "gzip", *os_version.release_package_names],
```
alexandrevicenzi, line 66: Correct me if I'm wrong, but this seems to be required for all images based on `OsContainer`. If yes, it would be better to override the `packages` property.

Something like this perhaps:

```python
@property
  def packages(self) -> str:
    return super().packages() + [*os_version.release_package_names]

```

dirkmueller: no, it's not, it is only needed for those base containers that explicitly pull in a package that requires `distribution-release`

alexandrevicenzi (approved):

src/bci_build/package/basecontainers.py
alexandrevicenzi: I looked over a few files and saw that most were including this package, but if it is not required it does not make sense to override the `packages` property.

PR #1700, opened 2024-09-20
Don't use the full registry url in FROM on Tumbleweed

dcermak:
This is discouraged in TW: https://build.opensuse.org/request/show/1202139

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -834,10 +834,13 @@ exit 0
             return ""
 
         if self.from_target_image:
-            return (
-                f"FROM {self.base_image_registry}/{self.from_target_image} AS target\n"
-                f"FROM {self._from_image} AS builder"
+            # build against the released container on SLE for proper base.digest/name generation
+            target: str = (
+                self.from_target_image
+                if self.os_version.is_tumbleweed
+                else f"{self.base_image_registry}/{self.from_target_image}"
             )
+            return f"FROM {target} AS target\nFROM {self._from_image} AS builder"
 
         return f"FROM {self._from_image}"
 


src/bci_build/package/__init__.py
```diff
@@ -834,10 +834,14 @@ def dockerfile_from_line(self) -> str:
             return ""
 
         if self.from_target_image:
-            return (
-                f"FROM {self.base_image_registry}/{self.from_target_image} AS target\n"
-                f"FROM {self._from_image} AS builder"
-            )
+            # use the full registry URL in FROM on SLE,
+            # but on TW we must not do this
```
alexandrevicenzi, line 838: nit: why not?

alexandrevicenzi (approved):

src/bci_build/package/__init__.py
dcermak: Because DimStar doesn't like it and keeps declining our containers: https://build.opensuse.org/request/show/1202139

dirkmueller (changes requested):
have we tried talking to dimstar?

src/bci_build/package/__init__.py
dirkmueller: the reason is not "doesn't like it" but the reason is "we don't want to have a rebuild triggered from outside openSUSE:Factory" which seems not entirely valid because this is not a source change and Factory is on rebuild=local only. 

I keep discussing this. anyhow we should capture the actual reason in the commit message so that we can go back to it when / if this needs to be revisited.

dirkmueller (approved):

dirkmueller:
updated commit message and merging due to unresolved discussion with tw team

PR #1705, opened 2024-09-20
Install openssl-3 in the base container

dirkmueller:
15.6+ want the "openssl" cli command in the testsuite, so install it explicitly.

### DIFF
--- a/src/bci_build/package/base.py
+++ b/src/bci_build/package/base.py
@@ -148,6 +148,7 @@ def _get_base_kwargs(os_version: OsVersion) -> dict:
                     "curl",
                     "gzip",
                     "netcfg",
+                    "openssl-3",
                     "tar",
                     "timezone",
                     *os_version.eula_package_names,


alexandrevicenzi (approved):
which testsuite requires this? would be nice to include in the log message

dirkmueller:
> which testsuite requires this? would be nice to include in the log message

right, that was kinda written in hectic, and did not read well. fixed.

dcermak:
Do we really need openssl-3 installed in the container? Can't we just explicitly `zypper in openssl` in the tests?

dirkmueller:
> Do we really need openssl-3 installed in the container? Can't we just explicitly `zypper in openssl` in the tests?

we discussed this and decided for the openssl cli to be available in the base container: 

https://confluence.suse.com/display/ENGCTNRSTORY/2023-07-11+-+BCI+Sync

dcermak:
> > Do we really need openssl-3 installed in the container? Can't we just explicitly `zypper in openssl` in the tests?
> 
> we discussed this and decided for the openssl cli to be available in the base container:
> 
> https://confluence.suse.com/display/ENGCTNRSTORY/2023-07-11+-+BCI+Sync

Fair enough, totally forgot about that one. Thanks for the reminder!

PR #1714, opened 2024-09-24
Remove full ref on tumbleweed also for last layer

dirkmueller:


src/bci_build/package/__init__.py
```diff
@@ -829,18 +829,23 @@ def _from_image(self) -> str | None:
         return f"bci/bci-base:15.{self.os_version}"
 
     @property
-    def dockerfile_from_line(self) -> str:
-        if self._from_image is None:
-            return ""
-
+    def dockerfile_from_target_ref(self) -> str | None:
         if self.from_target_image:
```
alexandrevicenzi, line 833: nit: This might be better, fail fast, return an empty string, no need to change the function return type.

```suggestion
        if self.from_target_image is None:
            return ""
```

alexandrevicenzi (dismissed):

alexandrevicenzi (approved):

PR #1720, opened 2024-09-24
Add latest tags to LTSS containers

dirkmueller:


alexandrevicenzi (approved):
Don't we have any BZ or Jira ticket for this? It would be nice to capture in the commit description why we went with this change.

PR #1750, opened 2024-09-26
Remove the "SLE BCI" advertising for Appcollection

dirkmueller:


src/bci_build/package/apache_tomcat.py
```diff
@@ -58,8 +58,9 @@ def _get_sac_supported_until(
         pretty_name="Apache Tomcat",
         custom_description=(
             "Apache Tomcat is a free and open-source implementation of the Jakarta Servlet, "
-            "Jakarta Expression Language, and WebSocket technologies, {based_on_container}."
-        ),
+            "Jakarta Expression Language, and WebSocket technologies"
+        )
+        + (", {based_on_container}." if os_version.is_tumbleweed else "."),
```
alexandrevicenzi, line 63: I'm assuming this is a Jinja template, if yes, it could be done this way:

```suggestion
            f"Jakarta Expression Language, and WebSocket technologies{{% if {os_version.is_tumbleweed} %}}, {{based_on_container}}{{% endif %}}."
```

It looks better in some ways, but not perfect.

alexandrevicenzi (approved):

src/bci_build/package/apache_tomcat.py
dirkmueller: no, its not via jinja, but via str.format(). I'm open to changing that, but not in this PR.

PR #1755, opened 2024-09-26
Build the apache-tomcat containers as build_flavors

dirkmueller:
This allows to sync the -RELEASE tag between multiple flavors, which is a requirement for appcollection.

For now all the build_flavors need to be manually passed in for _multibuild and _service generation. This could be eventually replaced by more automagic later on.

### DIFF
--- a/src/bci_build/containercrate.py
+++ b/src/bci_build/containercrate.py
@@ -1,0 +1,39 @@
+"""Crate to handle multibuild containers in the generator."""
+
+
+class ContainerCrate:
+    """ContainerCrate is combining multiple container build flavors.
+
+    This provides package-central functions like generating _service and
+    _multibuild files.
+    """
+
+    def __init__(self, containers: list):
+        """Assign the crate for every container."""
+        self._all_build_flavors: dict[tuple, set] = {}
+        for container in containers:
+            if container.build_flavor:
+                self._all_build_flavors.setdefault(
+                    (container.os_version, container.package_name), set()
+                ).add(container.build_flavor)
+
+        for container in containers:
+            if container.crate is not None:
+                raise ValueError("Container is already part of a ContainerCrate")
+            container.crate = self
+
+    def all_build_flavors(self, container):
+        """Return all build flavors for this container in the crate"""
+        return sorted(
+            self._all_build_flavors.get(
+                (container.os_version, container.package_name), [""]
+            )
+        )
+
+    def multibuild(self, container):
+        """Return the _multibuild file string to write for this ContainerCrate."""
+        flavors: str = "\n".join(
+            " " * 4 + f"<package>{pkg}</package>"
+            for pkg in self.all_build_flavors(container)
+        )
+        return f"<multibuild>\n{flavors}\n</multibuild>"
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -16,6 +16,7 @@ from typing import overload
 import jinja2
 from packaging import version
 
+from bci_build.containercrate import ContainerCrate
 from bci_build.templates import DOCKERFILE_TEMPLATE
 from bci_build.templates import INFOHEADER_TEMPLATE
 from bci_build.templates import KIWI_TEMPLATE
@@ -457,6 +458,12 @@ class BaseContainerImage(abc.ABC):
     env: dict[str, str | int] | dict[str, str] | dict[str, int] = field(
         default_factory=dict
     )
+
+    #: build flavors to produce for this container variant
+    build_flavor: str | None = None
+
+    #: create that this container is part of
+    crate: ContainerCrate = None
 
     #: Add any replacements via `obs-service-replace_using_package_version
     #: <https://github.com/openSUSE/obs-service-replace_using_package_version>`_
@@ -1156,8 +1163,12 @@ exit 0
         return f"{self.os_version.distribution_base_name} BCI {self.pretty_name}"
 
     @property
+    def readme_name(self) -> str:
+        return f"README.{self.build_flavor}.md" if self.build_flavor else "README.md"
+
+    @property
     def readme_path(self) -> str:
-        return f"{self.package_name}/README.md"
+        return f"{self.package_name}/{self.readme_name}"
 
     @property
     def readme_url(self) -> str:
@@ -1168,7 +1179,7 @@ exit 0
         if self.os_version.is_tumbleweed:
             return f"https://raw.githubusercontent.com/SUSE/BCI-dockerfile-generator/{self.os_version.deployment_branch_name}/{self.readme_path}"
 
-        return "%SOURCEURL%/README.md"
+        return f"%SOURCEURL%/{self.readme_name}"
 
     @property
     def readme(self) -> str:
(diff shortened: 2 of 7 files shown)


src/bci_build/package/__init__.py
```diff
@@ -1320,9 +1344,22 @@ async def write_file_to_dest(fname: str, contents: str | bytes) -> None:
                 False
             ), f"got an unexpected build_recipe_type: '{self.build_recipe_type}'"
 
+        if self.build_flavor:
+            flavors: str = "\n".join(
+                " " * 4 + f"<package>{pkg}</package>"
```
alexandrevicenzi, line 1349: This looks really odd, just to produce what this does:

```suggestion
                f"    <package>{pkg}</package>"
```

src/bci_build/package/__init__.py
```diff
@@ -458,6 +458,12 @@ class BaseContainerImage(abc.ABC):
         default_factory=dict
     )
 
+    #: build flavors to produce for this container variant
+    build_flavor: str | None = None
+
+    #: Provide all the build flavors that exist, which is used for _multibuild/_service writing
+    all_build_flavors: list[str] | None = None
+
```
alexandrevicenzi, line 466: This could be one single list IMHO. For example, if you create a tuple, dict, or a Flavor class.

```python
build_flavors: list[Flavor] = [Flavor(name="foo", current=True), Flavor(name="bar", current=False)]
```

If you have one item, you can remove some new checks in `__post_init__`.

src/bci_build/package/apache_tomcat.py
```diff
@@ -73,6 +76,8 @@ def _get_sac_supported_until(
         supported_until=_get_sac_supported_until(
             os_version=os_version, tomcat_ver=tomcat_ver, jre_major=jre_version
         ),
+        build_flavor=f"openjdk{jre_version}",
+        all_build_flavors=[f"openjdk{ver}" for ver in all_jre_versions],
```
alexandrevicenzi, line 80: Taking my idea about one single item, I think something like this could be done here.

```python
        build_flavors=[Flavor(name=f"openjdk{ver}", current=ver==jre_version) for ver in all_jre_versions],
```

tests/test_service.py
```diff
@@ -108,3 +108,35 @@ def test_service_with_replacement_docker():
   </service>
 </services>"""
     )
+
+
+def test_service_with_multi_flavor_docker():
```
alexandrevicenzi, line 113: possibly worth adding a test to check if `_multibuild` is correct, not just for `_service`.

src/bci_build/package/__init__.py
dirkmueller: I guess this comes down to a personal preference and DRY, don't care either way.

src/bci_build/package/__init__.py
dirkmueller: As originally indicated I wanted to make the all_build_flavor generation automatic, I just didn't have the time yesterday. 

did so now in the updated version.

tests/test_service.py
dirkmueller: tests added.

src/bci_build/package/apache_tomcat.py
dirkmueller: I think this is way too complicated for a simple container definition, and would have to repeated for each one of them (lets think about the python containers becoming multibuild, for example).

alexandrevicenzi: Yes, it was just an idea, because I wasn't that happy with the solution proposed.

src/bci_build/containercrate.py
```diff
@@ -0,0 +1,40 @@
...
+class ContainerCrate:
+    """ContainerCrate is combining multiple container build flavors.
+
+    This provides package-central functions like generating _service and
+    _multibuild files.
+    """
+
+    def __init__(self, containers: list):
+        """Assign the crate for every container."""
+        self._all_build_flavors: dict[tuple, set] = {}
```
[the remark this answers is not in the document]
dirkmueller: The crate is to unify over `(os_version, container.package_name)`. if I understand correctly, you're unifying over all given containers (which could be many different versions, which have different build flavors depending on version and also have different package names, like in the tomcat case). However, we do not want to collect flavors that are not built for this version or not build for this package name. hence, we need to group them by this, which is not working as far as I can see in your proposal. Maybe I'm missing something.

src/bci_build/containercrate.py
```diff
@@ -0,0 +1,40 @@
...
+
+        for container in containers:
+            if container.crate is not None:
+                raise ValueError("Container is already part of a ContainerCrate")
+            container.crate = self
+            container.all_build_flavors = self.all_build_flavors(container)
+
+    def all_build_flavors(self, container):
+        """Return all build flavors for this container in the crate"""
+        return sorted(
```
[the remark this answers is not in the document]
dirkmueller: in `__init__` it is still a set, which is unsorted. I agree we move away from storing it as a set and switch to a sorted list. I don't think you'll be able to measure any difference but I agree it is more efficient.

src/bci_build/templates.py
```diff
@@ -194,13 +194,24 @@
     """<services>
   <service mode="buildtime" name="{{ image.build_recipe_type }}_label_helper"/>
   <service mode="buildtime" name="kiwi_metainfo_helper"/>
+{%- set all_build_flavors = [""] %}
+{%-     if image.crate and image.build_flavor %}
+{%-         set all_build_flavors = image.crate.all_build_flavors(image) %}
+{%-     endif %}
+{%- for flavor in all_build_flavors %}
```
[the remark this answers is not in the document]
dirkmueller: yeah, I was trying to avoid storing the (essentially duplicated) information in the ContainerImage because it is only needed for this one case. Polluting yet another method into that ContainerImage seems rather messy to me. 

my intention was to move the generation of `_multibuild` and `_service` into the Crate entirely, but that was a bit too much for one commit so it'll come later on.

src/bci_build/containercrate.py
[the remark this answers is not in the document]
dirkmueller: As an example, with tomcat, there are two flavors on SP6 (17, 21) and 3 on tumbleweed (17, 21, 22). the generated _multibuild file needs to be specific to the os-version. in theory it could even be specific to the package name. 

dict/set are the most efficient ways to aggregate it.

src/bci_build/package/__init__.py
```diff
@@ -1137,9 +1144,13 @@ def title(self) -> str:
         """
         return f"{self.os_version.distribution_base_name} BCI {self.pretty_name}"
 
+    @property
+    def readme_name(self) -> str:
+        return f"README.{self.build_flavor}.md" if self.build_flavor else "README.md"
+
```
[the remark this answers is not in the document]
dirkmueller: that was a leftover anyhow, it's not used anywhere.

dcermak (commented):
To be frankly honest, I don't particularly like the current abstraction2 both in naming and behavior: it adds additional behind the scenes mutation and doesn't make it very obvious that we're actually combining multiple containers in one package.

I'd like to re-design the API. Merge this as is, if it is time critical though.

dirkmueller:
> To be frankly honest, I don't particularly like the current abstraction2 both in naming and behavior: it adds additional behind the scenes mutation and doesn't make it very obvious that we're actually combining multiple containers in one package.

well, names can be changed. so "PackageContainer" then instead of Crate?

> I'd like to re-design the API. Merge this as is, if it is time critical though.

I don't see how .. yeah, it is kinda urgent for the SAC prototype, but I can adjust it if there is a concrete feedback on what to improve.

dirkmueller:
as discussed, will be solved in a followup.

PR #1757, opened 2024-09-27
Add a Python 3.9 container (jsc#PED-10823)

dirkmueller:
For AppCol..


src/bci_build/registry.py
```diff
@@ -0,0 +1,73 @@
...
+    registry: str
+    """The vendor that is put into the ``org.opencontainers.image.vendor`` label"""
+    vendor: str
+
+    @abstractmethod
+    def url(self, container) -> str:
+        pass
+
+    @abstractmethod
+    def registry_prefix(self, *, is_application) -> str:
```
dcermak, line 24: no usage of `self`:
```suggestion
    @staticmethod
    def registry_prefix(self, *, is_application: bool) -> str:
```

src/bci_build/registry.py
```diff
@@ -0,0 +1,73 @@
...
+    """Abstract Base Class for defining Registry specific content."""
+
+    _: dataclasses.KW_ONLY
+    """The base hostname for this registry instance"""
+    registry: str
+    """The vendor that is put into the ``org.opencontainers.image.vendor`` label"""
+    vendor: str
+
+    @abstractmethod
+    def url(self, container) -> str:
```
dcermak, line 20: no usage of `self`:
```suggestion
    @staticmethod
    def url(self, container: BaseContainerImage) -> str:
```

src/bci_build/registry.py
```diff
@@ -0,0 +1,73 @@
...
+        super().__init__(registry="registry.opensuse.org", vendor="openSUSE Project")
+
+    def url(self, container) -> str:
+        return "https://www.opensuse.org"
+
+    def registry_prefix(self, *, is_application) -> str:
+        return "opensuse" if is_application else "opensuse/bci"
+
+
+def get_registry(container) -> RegistryABC:
```
dcermak, line 69:
```suggestion
def get_registry(container: BaseContainerImage) -> RegistryABC:
```

src/bci_build/registry.py
```diff
@@ -0,0 +1,73 @@
...
+    """Return the appropriate registry for the container."""
+    if container.os_version.is_tumbleweed:
+        return openSUSERegistry()
+    return SUSERegistry()
```
dcermak, line 73: Or allow to specify whether this is an appcol image and pass the `os_version` as a parameter instead of the container, that would allow to remove the duplication of `None if os_version.is_tumbleweed else ApplicationCollectionRegistry()`
```suggestion
def get_registry(os_version: OsVersion, is_app_col: bool = False) -> RegistryABC:
    """Return the appropriate registry for the container."""
    if os_version.is_tumbleweed:
        return openSUSERegistry()
    return ApplicationCollectionRegistry() if is_app_col else SUSERegistry()
```

dcermak (commented):
This is fine in principle. I've added a few suggestions how to improve the Registry classes and I'm not entirely happy with the amount of pass through methods. But it's a good start in the right direction, thank you!

src/bci_build/registry.py
dirkmueller: this is obsolete with the change to os_version

src/bci_build/registry.py
dirkmueller: this was intentionally the way it is to reduce refactoring effort. using os_version means there are recursive imports, which requires moving code around which you usually object to..

dirkmueller: splitted into a new toplevel module.

PR #1764, opened 2024-09-30
Set full ref for derived containers

dirkmueller:
We need the full registry path in the FROM line so that the build step fetches it from the registry which is required for the digest annotation to be generatable. otherwise org.opencontainers.image.base.digest is empty.


alexandrevicenzi (approved):

dcermak:
@alexandrevicenzi @dirkmueller this broke the unit tests, please take a closer look when force merging

dirkmueller:
> @alexandrevicenzi @dirkmueller this broke the unit tests, please take a closer look when force merging

yeah, I noticed that a second after merging. there is a followup PR that fixes it.

PR #1767, opened 2024-09-30
Change KIWI application builds to build against released containers

dirkmueller:
This is needed so that we have the right oci annotations. Also fixup tests after last change.

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -874,10 +874,13 @@ exit 0
     def kiwi_derived_from_entry(self) -> str:
         if self._from_image is None:
             return ""
-        # Adjust for the absolutely braindamaged format that OBS expects to reference
+        # Adjust for the special format that OBS expects to reference
         # external images
         if self.is_base_container_annotation_available:
-            return f" derived_from=\"obs://SUSE:Registry/standard/{self._from_image.replace('registry.suse.com/','').replace(':', '#')}\""
+            repo: str = self._from_image.replace("registry.suse.com/", "").replace(
+                ":", "#"
+            )
+            return f' derived_from="obs://SUSE:Registry/standard/{repo}"'
         return (
             f" derived_from=\"obsrepositories:/{self._from_image.replace(':', '#')}\""
         )
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -848,8 +848,14 @@ exit 0
     @property
     def is_base_container_annotation_available(self) -> bool:
         """return True if the obs-service-kiwi_metainfo_helper can provide base.name/digest annotations."""
+        _from_image = self._from_image
         return bool(
-            self._from_image
+            _from_image
+            and self.os_version in RELEASED_OS_VERSIONS
+            and (
+                self.from_target_image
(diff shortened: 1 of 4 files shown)


src/bci_build/package/appcontainers.py
```diff
@@ -434,7 +435,11 @@ def _get_nginx_kwargs(os_version: OsVersion):
     ApplicationStackContainer(
         name="helm",
         pretty_name="Kubernetes Package Manager",
-        from_image=f"{_build_tag_prefix(os_version)}/bci-micro:{OsContainer.version_to_container_os_version(os_version)}",
+        from_image=(
```
alexandrevicenzi, line 438: nit: can't we simplify this? The only bit that changes is the prefix.

src/bci_build/package/__init__.py
```diff
@@ -868,6 +874,10 @@ def dockerfile_from_line(self) -> str:
     def kiwi_derived_from_entry(self) -> str:
         if self._from_image is None:
             return ""
+        # Adjust for the absolutely braindamaged format that OBS expects to reference
```
alexandrevicenzi, line 877: nit: let's not be that harsh.

```suggestion
        # Adjust for the weird format that OBS expects to reference
```

src/bci_build/package/__init__.py
```diff
@@ -868,6 +874,10 @@ def dockerfile_from_line(self) -> str:
...
+        # external images
+        if self.is_base_container_annotation_available:
+            return f" derived_from=\"obs://SUSE:Registry/standard/{self._from_image.replace('registry.suse.com/','').replace(':', '#')}\""
```
alexandrevicenzi, line 880: nit:  a bit more clear
```suggestion
            # cleanup image name
            image = self._from_image.replace('registry.suse.com/','').replace(':', '#')
            return f" derived_from=\"obs://SUSE:Registry/standard/{image}\""
```

alexandrevicenzi (approved):

src/bci_build/package/__init__.py
dirkmueller: good suggestion, merged.

src/bci_build/package/appcontainers.py
dirkmueller: moved into a helper function, since we need it many times.

alexandrevicenzi (approved):

src/bci_build/package/helpers.py
```diff
@@ -25,3 +28,10 @@ def generate_package_version_check(
 
     return f"""# sanity check that the version from the tag is equal to the version of {pkg_name} that we expect
 {DOCKERFILE_RUN} [ "$(rpm -q --qf '%{{version}}' {pkg_name} | cut -d '.' -f -{cut_count})" = "{pkg_version}" ]"""
+
+
+def generate_from_image(os_version, container_name) -> str:
```
dcermak, line 33: I would also suggest to rename this function to something along the lines `from_image_tag` or `generate_from_image_url` or 

```suggestion
def generate_from_image(os_version: OsVersion, container_name: str) -> str:
    """Return the build tag or url of the container with the name `container_name` for the specified `os_version`."""
```

PR #1781, opened 2024-10-01
Revert obs:// path

dirkmueller:
That fails in sourc review. Changed the project config instead


dcermak (changes requested):
LGTM, but needs adjustments in `tests/test_project_setup.py` for the unit tests to pass

dirkmueller:
> LGTM, but needs adjustments in `tests/test_project_setup.py` for the unit tests to pass

I just needed temporarily access to fix the prjconf

dcermak:

>> LGTM, but needs adjustments in `tests/test_project_setup.py` for the unit tests to pass
>
> I just needed temporarily access to fix the prjconf

You can also modify the _config file [1] (via PR or commit) and the bot
should deploy it automatically.

Footnotes:
[1]  https://github.com/SUSE/BCI-dockerfile-generator/blob/sle15-sp6/_config

-- 
Dan Čermák ***@***.***>
Software Engineer Development tools
SUSE Software Solutions Germany GmbH
Frankenstr. 146
90461 Nürnberg
Germany

www.suse.com
Geschäftsführer: Ivo Totev, Andrew McDonald, Werner Knoblich
(HRB 36809, AG Nürnberg)

dcermak (approved):

dirkmueller:
> You can also modify the _config file [1] (via PR or commit) and the bot should deploy it automatically. https://github.com/SUSE/BCI-dockerfile-generator/blob/sle15-sp6/_config

yes, but thats not how I expected it. it should deploy the _config from the PR-XXXX branch instead (which it generated there). 

anyhow, I tricked the staging bot so it should turn green.

dcermak:
> > You can also modify the _config file [1] (via PR or commit) and the bot should deploy it automatically. https://github.com/SUSE/BCI-dockerfile-generator/blob/sle15-sp6/_config
> 
> yes, but thats not how I expected it. it should deploy the _config from the PR-XXXX branch instead (which it generated there).

Ah, well we can change that for sure, but it might lead to unforeseen consequences, as your changes in the prjconf from the PR-$num branch will not automatically be propagated into the sle15-sp$N branch.

dirkmueller:
> Ah, well we can change that for sure, but it might lead to unforeseen consequences, as your changes in the prjconf from the PR-$num branch will not automatically be propagated into the sle15-sp$N branch.

it does generate it into the for-deploy-XX branches already (somehow)

PR #1794, opened 2024-10-02
Added test for test_util to ensure file absence when open for reading

ilyas829:
- I have added a pytest function to ensure file absence when open for reading.
- It basically for below purpose

1. Creates a file
2. Open the file for reading and
3. Ensures that the file is removed

- Please review my PR and merge it 
- also suggest your valuable feedback.

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -32,12 +32,13 @@ async def test_ensure_directory_absent(tmp_path: pathlib.Path):
     await ensure_absent(path)
     assert not await aiofiles.os.path.exists(path)
 
+
 @pytest.mark.asyncio
 async def test_ensure_file_absent_when_open_for_reading(tmp_path: pathlib.Path):
     path = tmp_path / "test_file"
     async with aiofiles.open(path, "w") as test_file:
         await test_file.write("foobar")
-    
+
     # Open the file for reading
     async with aiofiles.open(path, "r"):
         pass  # Simply open and close the file
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -31,3 +31,17 @@ async def test_ensure_directory_absent(tmp_path: pathlib.Path):
 
     await ensure_absent(path)
     assert not await aiofiles.os.path.exists(path)
+
+@pytest.mark.asyncio
+async def test_ensure_file_absent_when_open_for_reading(tmp_path: pathlib.Path):
+    path = tmp_path / "test_file"
+    async with aiofiles.open(path, "w") as test_file:
+        await test_file.write("foobar")
+    
+    # Open the file for reading
+    async with aiofiles.open(path, "r"):
+        pass  # Simply open and close the file
+
+    # Ensure the file is removed
+    await ensure_absent(path)
+    assert not await aiofiles.os.path.exists(path)
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -1,5 +1,5 @@
 import pathlib
-
+import os
 import aiofiles.os
 import pytest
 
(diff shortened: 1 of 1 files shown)


dirkmueller:
@ilyas829 thanks for your contribution. could you please run "poetry run ruff format" before submitting ? you could amend the commit. 

since you're only adding testcases, is the only purpose to increase test coverage or did you find a real issue somewhere?

ilyas829:
Thank you for reviewing my test coverage codes.. Yes the purpose of appending test cases were to increase test coverage but I din not found any issue as i wrote and tested these cases, but now I am moving up in the hierarchy and find errors or bugs in the test_user, test_services and so on.
After I complete the test cases, I am planning to make a test automation framework using python + selenium that will also help us to find bugs like link breakage, accessibility etc. (If you approve this I can start working on this after completion of test files).
Now I have added formatting to my previous branches too, As I have just got started I had to make several branches to avoid my confusion please approve that too.
Thanks again
Waiting for your suggestions and feedback .

dirkmueller:
Thanks. basically your new test is a duplicate of https://github.com/SUSE/BCI-dockerfile-generator/pull/1794/files#diff-0dcbadaa6c70afb6fea041dec04bb89c3cc7fe201fb7c3185a669efa49769f28R10-R16 . the fact that you can read a file after creating it is not super useful given that the purpose of this unit test is to validate that "ensure_absent" works. 

Given that you confirmed that this isn't testing for a real issue, I would actually not be in favor of adding this test. if you can find a combination of file / directory / permission creation where ensure_absent breaks and you write a test for that, that would be awesome as a contribution.

ilyas829:
Perfect, 1st of all Thank you for describing the required test case. Yes we can do that. 

- To create a scenario where ensure_absent could potentially break, we need to consider edge cases that involve:

- File permissions: Setting restrictive file or directory permissions, which could prevent deletion.
- Special file types: Testing with special files like sockets, device files, or symbolic links with specific permissions.
- One realistic edge case could involve restricted permissions on a directory or file, where ensure_absent is unable to remove it due to lack of write or execute permissions.

- Edge Case: Directory with No Write Permissions
- In this test case, we can create a directory with no write permissions and attempt to delete it using ensure_absent. This should raise a permission error, as the lack of write permissions will prevent the directory from being removed.

The code for the same is just pushed and  please review and suggest your valuable feedback

ilyas829:
Hey @dirkmueller please can you check as it this condition is working for me..

dirkmueller:
@dcermak should we fix ensure_absent to handle permission errors like this rather than raising an exception?

tests/test_util.py
```diff
@@ -31,3 +31,26 @@ async def test_ensure_directory_absent(tmp_path: pathlib.Path):
...
+):
+    # Create a directory
+    path = tmp_path / "test_dir"
+    await aiofiles.os.mkdir(path)
+
+    # Change permissions to remove write access (read and execute only)
+    os.chmod(path, 0o500)
+
+    # Try to remove the directory - should raise a PermissionError or OSError
+    with pytest.raises((PermissionError, OSError)):
```
alexandrevicenzi, line 48: This will not work if the tests are executed as root, root can do anything to a directory.

alexandrevicenzi:
> I am planning to make a test automation framework using python + selenium that will also help us to find bugs like link breakage, accessibility etc. (If you approve this I can start working on this after completion of test files).

Why? This does not make any sense. This project is not a web application, there's no reason to use Selenium to test anything.

> * To create a scenario where ensure_absent could potentially break, we need to consider edge cases that involve:
> * File permissions: Setting restrictive file or directory permissions, which could prevent deletion.
> * Special file types: Testing with special files like sockets, device files, or symbolic links with specific permissions.
> * One realistic edge case could involve restricted permissions on a directory or file, where ensure_absent is unable to remove it due to lack of write or execute permissions.
> * Edge Case: Directory with No Write Permissions

Honestly, there's no reason to test for most of these scenarios. `ensure_absent` is used in one single place in the code.

It will never attempt to remove a socket or device, and if it did that is completely wrong.
It always has permission to remove the folder, because the folder that it tries to remove is created by the running user.

alexandrevicenzi:
> @dcermak should we fix ensure_absent to handle permission errors like this rather than raising an exception?

@dirkmueller the folder that `ensure_absent` tries to remove [1] should not have permission issues because the folder is created by the running user [2].

I don't see a reason the change the current behavior of this function.

[1]: https://github.com/SUSE/BCI-dockerfile-generator/blob/main/src/staging/bot.py#L563
[2]: https://github.com/SUSE/BCI-dockerfile-generator/blob/main/src/staging/bot.py#L541

PR #1795, opened 2024-10-02
Test case added for test_util.py

ilyas829:
- To ensure that non empty directory is absent I have added a test case in the  test_Util.py which basically 

1. Create a directory
2. Create a file inside the directory
3. Try to remove the non-empty directory
4. checks whether the directory  still exist (it should) since it was not removed

- Please review my PR and merge it .
- Also provide your valuable feedback .


alexandrevicenzi:
Duplicate of #1797.

PR #1796, opened 2024-10-02
Added few more test cases in test_Util.py

ilyas829:
- Comments are added for better understanding

PR #1797, opened 2024-10-02
Negative test case

ilyas829:
Test cases added to `ensure_absent_removes_nested_empty_directory`
Comments are also added for better understandability. 👍


alexandrevicenzi:
This is basically the same as #1796.

dcermak:
Thank you for your effort, but these test scenarios are not really adding much value to the code. This tiny helper function that you're testing is only used in two places and hasn't been a source of bugs. If you should find a bug in it, then I'd certainly encourage you to submit a PR with a regression test.

If you wish to contribute to this project, then there are far uglier parts in the code that need improvements.

PR #1805, opened 2024-10-07
Explicitly install which for tomcat containers

dirkmueller:


dirkmueller:
merging to unbreak the tomcat containers..

PR #1815, opened 2024-10-08
Introduce a ObsPackage class for grouping Container Images

dcermak:
This is pretty much a re-implementation of https://github.com/SUSE/BCI-dockerfile-generator/pull/1755, with more code but an API that I personally like a bit more, because it is more explicit:
- it is now clear which containers are part of a single package
- which name the resulting package gets

This required the introduction of another base class `ObsPackageBase` from which both `ObsPackage` and `BaseContainerImage` inherit. The API is not yet great, as there's now a bit of duplication going on, if you stuff a BCI into a `ObsPackage`, but I wanted to get some feedback first.


src/bci_build/package/__init__.py
```diff
@@ -1479,7 +1484,7 @@ def generate_disk_size_constraints(size_gb: int) -> str:
 
 SORTED_CONTAINER_IMAGE_NAMES = sorted(
     ALL_CONTAINER_IMAGE_NAMES,
-    key=lambda bci: f"{ALL_CONTAINER_IMAGE_NAMES[bci].os_version}-{ALL_CONTAINER_IMAGE_NAMES[bci].name}",
+    key=lambda bci: f"{ALL_CONTAINER_IMAGE_NAMES[bci].os_version}-{ALL_CONTAINER_IMAGE_NAMES[bci].uid}",
```
dirkmueller, line 1474: that change seems unrelated?

src/bci_build/package/obs_package.py
```diff
@@ -0,0 +1,217 @@
...
+        tree = ET.ElementTree(root)
+        ET.indent(tree)
+        io = BytesIO()
+        tree.write(io, encoding="utf-8")
+        io.seek(0)
+        return io.read().decode("utf-8")
+
+    async def _write_service_file(self, dest: str) -> list[str]:
+        await write_to_file(os.path.join(dest, "_service"), self._service_file_contents)
+        return ["_service"]
```
dirkmueller, line 95: that return value seems unused? since it should end up in the files list anyway..

dirkmueller:
There's at least 3 refactorings in this PR which makes it really hard to review. the osversion -> os_version rename, the service jinja to class refactoring and the change of the containercrate to ObsPackage refactoring. 

any chance to split the first two out so that it is easier to see what the rest is about? this is a 1000 lines diff over 11 commits. 

personally I don't really like the *Base* in  the name `ObsPackageBase` . what about renaming `ObsPackageBase` to `ObsPackage` and `ObsPackage` to `MultibuildObsPackage` (or multiflavor) ? that would make it a bit more logical IMHO.

also, why is there "magic" to collate up the services, but no magic to collate up the build flavors? instead they have to be set explicitly in multiple list entries (so whenever we add or remove a os_version that has different build flavors, the whole shebang needs to be duplicated)? Its such an implementation detail, it should be hidden the same way like services imho.

src/bci_build/package/__init__.py
dcermak: It is required, because `ObsPackageBase` doesn't have a `name` property

src/bci_build/package/obs_package.py
dcermak: It is used here: https://github.com/SUSE/BCI-dockerfile-generator/blob/7e43203c16fa10c06faa7abea7eaa9e2acf676df/src/bci_build/package/obs_package.py#L187

dcermak:
> personally I don't really like the _Base_ in the name `ObsPackageBase` . what about renaming `ObsPackageBase` to `ObsPackage` and `ObsPackage` to `MultibuildObsPackage` (or multiflavor) ? that would make it a bit more logical IMHO.

I've renamed the classes as you suggested, `ObsPackageBase` is now `ObsPackage` and `ObsPackage` is now `MultiBuildObsPackage`

> also, why is there "magic" to collate up the services, but no magic to collate up the build flavors? instead they have to be set explicitly in multiple list entries (so whenever we add or remove a os_version that has different build flavors, the whole shebang needs to be duplicated)? Its such an implementation detail, it should be hidden the same way like services imho.

Hm, I don't fully understand what pain point that this should solve though?

src/bci_build/package/apache_tomcat.py
```diff
@@ -125,15 +122,29 @@ def _get_sac_supported_until(
...
+            _create_tomcat_container(os_version, tomcat_ver, jre_version)
+            for tomcat_ver, os_version, jre_version in (
+                ("10.1", OsVersion.TUMBLEWEED, 22),
+                ("10.1", OsVersion.TUMBLEWEED, 21),
+                ("10.1", OsVersion.TUMBLEWEED, 17),
+            )
+        ],
+        package_name="apache-tomcat-10-image",
+    ),
+    _create_tomcat_container(OsVersion.TUMBLEWEED, "9", 17),
```
dirkmueller, line 139: this should also be a multibuildobspackage, imho.

src/bci_build/package/obs_package.py
```diff
@@ -0,0 +1,217 @@
...
+from bci_build.service import Service
+from bci_build.util import write_to_file
+
+if TYPE_CHECKING:
+    from bci_build.package import BaseContainerImage
+
+
+@dataclass(kw_only=True)
+class ObsPackage(abc.ABC):
+    """Abstract base class of the ObsPackage and the BaseContainerImage."""
```
dirkmueller, line 27: needs updating

dirkmueller:
it seems this saga of refactorings is continued now in #2199

PR #1825, opened 2024-10-09
Move enumerations into a separate module

dcermak:
Having the enums in a separate module allows us to import it in other modules without having import cycles, split out of https://github.com/SUSE/BCI-dockerfile-generator/pull/1815

### DIFF
--- a/src/bci_build/container_attributes.py
+++ b/src/bci_build/container_attributes.py
@@ -1,0 +1,91 @@
+"""This package contains enumerations with constants like architectures or build
+types.
+
+"""
+
+import enum
+
+
+@enum.unique
+class Arch(enum.Enum):
+    """Architectures of packages on OBS"""
+
+    X86_64 = "x86_64"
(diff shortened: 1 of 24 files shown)


dirkmueller:
where is this going to be imported as well? also all the usages in bci_build.package.* are not updated to import directly rather than via bci_build.package. 

is there a better name than "constants?" e.g. os_version/osversion is also a "constant" in that sense. maybe "attributes" or "containerattributes" ?

dcermak:
> where is this going to be imported as well?

I need the attributes in the `obs_package` module in #1815.

> also all the usages in bci_build.package.* are not updated to import directly rather than via bci_build.package.

I have fixed that.

> is there a better name than "constants?" e.g. os_version/osversion is also a "constant" in that sense. maybe "attributes" or "containerattributes" ?

I've picked `container_attributes`, thanks for the suggestion!

dirkmueller (approved):

PR #1829, opened 2024-10-09
Set ForceMultiVersion for AppCollection containers

dirkmueller:
The AppCollection registry requires sequential release numbers for proper displaying. We probably should do the same for registry.suse.com at some point.


dcermak (commented):
As this is currently dependent on the registry, shouldn't we add it to the `Registry` class instead?

dirkmueller:
> As this is currently dependent on the registry, shouldn't we add it to the `Registry` class instead?

I wanted to make it configurable per image also, but YAGNI - moved it there.

PR #1828, opened 2024-10-09
Make the -RELEASE suffixed tag the first listed one

dirkmueller:
AppCol's registry integration requires the very first listed tag to be the immutable one. swap the order since BCI doesn't really care.

PR #1830, opened 2024-10-09
Introduce a branch_version concept in BCI

dirkmueller:
Now that we have build_flavors, we can add branch_versions, describing the full hierachy:

Each container can have a single `name`, one or many `branch_versions` and each of them one or many `build_flavors`.

this means our naming should become

{name}-{branch_version}-{build_flavor}

Introduce this for the AppCollection containers first.  This should eventually replace tag_version, which is a broken concept as it mixes namings, versions and flavors into and hence makes everything convoluted.

This commit mixes also a hook in Registry to use version, rather than tag_version for the build_version, which is a must-have for the AppCollection integration.

Removal of the deprecated tag_version will happen in a later followup.

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -14,7 +14,6 @@ from typing import Literal
 from typing import overload
 
 import jinja2
-from packaging import version
 
 from bci_build.container_attributes import Arch
 from bci_build.container_attributes import BuildType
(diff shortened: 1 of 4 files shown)


dirkmueller:
(ready for review but since I added tests it will require a rebase again once #1828 is merged)

alexandrevicenzi (commented):
LGTM, but there's a conflict to be resolved.

PR #1835, opened 2024-10-10
Use VERSION-RELEASE for new/future OsContainers

dirkmueller:
As decided, we're aligning the tagging scheme between OsContainers and Development Containers. To avoid friction with existing users and pipelines, we do this change only for future releases.


dirkmueller:
testsuite seems to not care about the change, which is good news

alexandrevicenzi (approved):

PR #1848, opened 2024-10-12
Redefine the meaning of tag_version to exclude build_flavor

dirkmueller:
for build_flavor containers, we had previously to include the build_flavor into the tag_version to disambiguate the tags and buildnames. Change the meaning of tag_version to be without build_flavor and instead explicitly handle build_flavor where necessary.

This commit mixes also a hook in Registry to use version, rather than tag_version for the build_version, which is a must-have for the AppCollection integration.

Add tests for ApplicationStackContainers in unit test, which was entirely untested so far.


dirkmueller:
alternative implementation to #1830  which is hopefully easier to review / less code to maintain.

src/bci_build/package/__init__.py
```diff
@@ -1235,7 +1258,13 @@ def build_tags(self) -> list[str]:
         tags = []
 
         for name in [self.name] + self.additional_names:
-            ver_labels: list[str] = [self.tag_version]
+            ver_labels: list[str] = [
+                f"{self.version}-{self.build_flavor}"
+                if self.build_flavor
+                else self.version
```
dcermak, line 1264: This would be `_version_variant`

src/bci_build/package/__init__.py
```diff
@@ -1252,7 +1281,8 @@ def build_tags(self) -> list[str]:
 
     @property
     def image_ref_name(self) -> str:
-        return f"{self.tag_version}-{self._release_suffix}"
+        variant: str = f"-{self.build_flavor}" if self.build_flavor else ""
+        return f"{self.version}{variant}-{self._release_suffix}"
```
dcermak, line 1285: And this one too:
```suggestion
        return f"{self.version}{self._version_variant}-{self._release_suffix}"
```

dcermak (commented):
This looks very good to me, thanks! I've added some minor suggestions, but I can address those later on myself

dcermak (approved):

PR #1855, opened 2024-10-14
Service abstraction

dcermak:
Split out of https://github.com/SUSE/BCI-dockerfile-generator/pull/1815


src/bci_build/service.py
```diff
@@ -0,0 +1,41 @@
...
+
+@dataclass(kw_only=True, frozen=True)
+class Service:
+    """Representation of an arbitrary source service in the Open Build Service."""
+
+    #: service mode (i.e. when the service runs)
+    mode: Literal["buildtime"] = "buildtime"
+
+    #: name of this service
+    name: str
```
dirkmueller, line 20: can we move this to the beginning, e.g. before mode? I prefer to have non-optional dataclass members first

PR #1861, opened 2024-10-18
expose ports in bci

rcmadhankumar:


src/bci_build/package/__init__.py
```diff
@@ -736,12 +746,15 @@ def _dockerfile_volume_expose(
...
         self,
-        instruction: Literal["EXPOSE", "VOLUME"],
+        instruction: Literal["EXPOSE", "EXPOSEUDP", "VOLUME"],
         entries: list[int] | list[str] | None,
     ):
         if not entries:
             return ""
 
+        if instruction == "EXPOSEUDP":
+            return "\n" + "EXPOSE " + " ".join(str(e) + "/udp" for e in entries)
```
dirkmueller, line 756: does it have to be a separate "EXPOSE" statement for udp? it feels like it could be all in one statement. so maybe the simplest thing would be to rename `exposes_tcp` into `exposes_port` and make it a list of strings instead, and then we can define udp exports as "1234/udp" and don't need special logic for handling udp ports. 

And then later on this can be refactored into a dataclass that takes the protocol as a type information.

dcermak: I agree with Dirk, I think we can go the full route now to avoid duplicate work later.

dirkmueller (approved):

PR #1869, opened 2024-10-21
Add redis container

cwh42:
We need a redis container for our Private registry project.
Might be useful for others, too.

Is this the way to contribute new bci images?

Not so inspired by [opensuse-redis-image.kiwi](https://build.opensuse.org/projects/openSUSE:Factory:ToTest/packages/opensuse-redis-image/files/opensuse-redis-image.kiwi?expand=1) any more


src/bci_build/package/appcontainers.py
```diff
@@ -484,3 +484,39 @@ def _get_nginx_kwargs(os_version: OsVersion):
     )
     for os_version in (OsVersion.TUMBLEWEED,)
 ]
+
+_REDIS_ENTRYPOINT = (Path(__file__).parent / "redis" / "entrypoint.tar.gz").read_bytes()
```
dirkmueller, line 488: please don't add binary blobs here. if this is coming from somewhere, please fetch it from https://github.com/SUSE/BCI-dockerfile-generator/blob/main/update-files.sh and add it unpacked.

cwh42: Removed the entrypoint. Blob is now gone.

src/bci_build/package/appcontainers.py
```diff
@@ -484,3 +485,41 @@ def _get_nginx_kwargs(os_version: OsVersion):
     )
     for os_version in (OsVersion.TUMBLEWEED,)
 ]
+
+REDIS_CONTAINERS = [
```
dirkmueller, line 489: generally looks good to me. there are a couple of small changes still outstanding:

* please move the REDIS_CONTAINERS definition to a src/bci_build/package/redis.py separate module instead. we're trying to slowly phase out appcontainers.py as a lumping collection of containers and instead have one module per container
* consider creating a README.md.j2 under src/bci_build/package/redis/ with some description and usage instructions (I'm working on something similar for the valkey container, there is an open unfinished PR from me on that
* potentially we have to convert this over to a multistage docker build, but that can be done as a followup, so it is only a NIT for now.

src/bci_build/package/appcontainers.py
```diff
@@ -484,3 +485,41 @@ def _get_nginx_kwargs(os_version: OsVersion):
...
+            chown redis:redis /var/lib/redis
+            """
+        ),
+        entrypoint_user="redis",
+        cmd=["redis-server", "--protected-mode no", "--dir /var/lib/redis"],
+        volumes=["/var/lib/redis"],
+        exposes_tcp=[6379],
+        build_recipe_type=BuildType.KIWI,
+    )
+    for os_version in ALL_NONBASE_OS_VERSIONS
```
dirkmueller, line 524: afaik redis  is dropped in SP7 and tumbleweed, so we should specify this only for SP6 for now.

cwh42: Ok. Then I am a bit late to the party.
Probably we should move to valkey soon

dirkmueller:
we have a valkey container now - please use that instead

PR #1881, opened 2024-10-23
Add workdir

cwh42:
Both, Dockerfile and KIWI support the WORK(ING)DIR directive.
Adding that to `BaseContainerImage`


src/bci_build/templates.py
```diff
@@ -100,6 +100,7 @@
 {%- if image.expose_dockerfile %}{{ image.expose_dockerfile }}{% endif %}
 {% if image.dockerfile_custom_end %}{{ image.dockerfile_custom_end }}{% endif %}
 {%- if image.entrypoint_user %}USER {{ image.entrypoint_user }}{% endif %}
+{%- if image.workdir %}WORKDIR {{ image.workdir }}{% endif %}
```
dirkmueller, line 103: currently we embed "WORKDIR" into "dockerfile_custom_end" which is a bit earlier in the order of Dockerfile statements than the new one you add here. 

could you please convert the existing users (or at least one of them) over to the new attribute (e.g. src/bci_build/package/php.py and others) so that we can see that it works the same way? alternatively extend tests/test_build_recipe to have workdir defined also somewhere.

dcermak:
Thanks for the PR @cwh42! Do you have a specific use case in mind, where the current approach of defining `WORKINGDIR` in the `Dockerfile` directly is insufficient?

PR #1883, opened 2024-10-23
Refactor expose ports to a list of ints or NetworkPort instances

dcermak:


tests/test_package.py
```diff
@@ -117,8 +119,15 @@ def test_expose_dockerfile(bci: BCI_FIXTURE_RET_T):
     cls, kwargs = bci
 
     assert (
-        cls(**kwargs, exposes_ports=["80", "443", "67/udp"]).expose_dockerfile
-        == "\nEXPOSE 80 443 67/udp"
+        cls(
+            **kwargs,
+            exposes_ports=[
+                NetworkPort(80),
+                NetworkPort(443),
```
dirkmueller, line 126: maybe test that an "int" still works by using plan `443` here ?

dirkmueller (approved):
nit on the testing coverage, lgtm otherwise

tests/test_package.py
dcermak: Technically you shouldn't do that (it works, but you shouldn't), as `exposes_ports` has the type `list[int] | list[NetworkPort] | None` and not `list[int | NetworkPort] | None`

dirkmueller: does anything prevent us from fixing the type annotation? or even better, stop supporting `list[int]` which seems to be now unwanted and unlogical hack over the typing aware indirection?

This seems to be an indirection just to make type checkers happy. Having `NetworkPort(443,  NetworkProtocol.TCP)` at least would make what it does explicit rather than implicit. Since you're touching all the statements anyway, lets convert them all to NetworkPort and drop the int variant.

alexandrevicenzi (approved):

tests/test_package.py
dcermak: The type `list[int] | list[NetworkPort]` was a conscious decision, I wanted to preserve the old & convenient behavior to just list TCP ports as integers (which is the most common use case and needn't be more complicated). But if the user wants UDP ports too, then they must use the `NetworkPort` class and have to resort to the more verbose syntax.

What I didn't want to do, is to force everyone to list TCP ports as `NetworkPort` instances even when no UDP ports are exposed. But I do see your point with explicit vs implicit.

@alexandrevicenzi what's your take on this?

alexandrevicenzi: Honestly, the keep the code concise, I think it makes sense to be explicit, otherwise mixing int and NetworkPort can be slightly confusing.

Perhaps you could have a shortcode `TCP(80)` and `UDP(80)` to avoid having to type `NetworkPort(80, NetworkProtocol.UDP)`.

dcermak: I like that a lot, great idea! I've adjusted the code accordingly

tests/test_package.py
```diff
@@ -96,11 +98,11 @@ def test_expose_port_kiwi(bci: BCI_FIXTURE_RET_T):
     cls, kwargs = bci
 
     assert (
-        cls(**kwargs, exposes_ports=["443", "80"]).exposes_kiwi
+        cls(**kwargs, exposes_ports=[TCP(443), TCP(80)]).exposes_kiwi
         == """
         <expose>
-          <port number="443" />
-          <port number="80" />
+          <port number="443/tcp" />
```
dirkmueller, line 104: is kiwi able to handle that syntax?

dirkmueller (approved):
LGTM assuming that kiwi can handle that syntax?

tests/test_package.py
dcermak: Yes, I checked with upstream

PR #1890, opened 2024-10-24
Add factory-maintainers group to the TW devel project

dcermak:


dirkmueller (approved):
It's a bit scary that there is no change in the test suite but that's how it is..

PR #1892, opened 2024-10-25
🤖: Update build recipes for SP6

dcermak:
This is an automated pull request to update from branch for-deploy-6


dcermak:
Disregard the check the changelogs failure, it's a false positive as due to a revert.

dirkmueller (approved):

dirkmueller:
changelog check here is wrong, there is no change to the tomcat containers.

PR #1897, opened 2024-10-25
Set the buildrelease counter on the main dockerfile

dirkmueller:
The buildrelease on flavor builds are ignored and copied from the main one.

### DIFF
--- a/src/bci_build/containercrate.py
+++ b/src/bci_build/containercrate.py
@@ -30,9 +30,13 @@ class ContainerCrate:
             )
         )
 
-    def default_dockerfile(self) -> str:
+    def default_dockerfile(self, container) -> str:
+        buildrelease: str = ""
+        if container.build_release:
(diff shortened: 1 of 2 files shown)


dcermak:
This doesn't look right: https://github.com/SUSE/BCI-dockerfile-generator/commit/c1d29dcb061c6d7f9a946384222d77ef23d2084f#diff-5430f28d74a56d4697369e4b9c0878d78e8074308609f7210dcffab7c263a1f0 
image

dirkmueller:
> This doesn't look right

what do you mean? it looks exactly right. the `BuildVersion: unknown` is a workaround for an OBS bug that will eventually be fixed.

dirkmueller:
> what do you mean? it looks exactly right. the `BuildVersion: unknown` is a workaround for an OBS bug that will eventually be fixed.

changed it to `workaround-for-an-obs-bug`  to make it clearer.

PR #1898, opened 2024-10-25
Overwrite the incorrect author from the base container in appcol

dirkmueller:
Appcollection wants no maintainer, which is however inherited from the base container.

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -348,8 +348,8 @@ class BaseContainerImage(abc.ABC):
             # Limit to aarch64 and x86_64
             if not self.exclusive_arch:
                 self.exclusive_arch = [Arch.AARCH64, Arch.X86_64]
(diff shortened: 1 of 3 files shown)


src/bci_build/package/__init__.py
```diff
@@ -348,8 +348,8 @@ def __post_init__(self) -> None:
             # Limit to aarch64 and x86_64
             if not self.exclusive_arch:
                 self.exclusive_arch = [Arch.AARCH64, Arch.X86_64]
-            # Disable maintainer listing
-            self.maintainer = None
+            # Disable maintainer listing from base container
```
dcermak, line 351: Maybe this is a bit clearer?
```suggestion
            # List maintainer as empty string to ensure that base container value is overwritten
```

dirkmueller: updated.

PR #1901, opened 2024-10-25
List contact details in org.opencontainers.image.authors

dirkmueller:
The key is supposed to contain a contact possibility. Remove it for openSUSE and default to the BCI forum instead rather than duplicating the vendor which is already under org.opencontainers.image.vendor


src/bci_build/package/__init__.py
```diff
@@ -350,6 +344,8 @@ def __post_init__(self) -> None:
                 self.exclusive_arch = [Arch.AARCH64, Arch.X86_64]
             # Disable maintainer listing
             self.maintainer = None
+        elif not self.maintainer and self.os_version.is_sle15:
+            self.maintainer = "https://github.com/SUSE/bci/discussions"
```
dcermak, line 348: This means that the maintainer is not set for SLFO, I guess that's not intentional?

dirkmueller: it was intentional. I don't see a lot of value in the field anymore and it is marked as optional in the spec. I think we should rather set it on the images that actually have a dedicated maintainer/forum instead of everywhere, including the base containers. 

but I'm not super opinionated on this. all we know is that we have the directive to not drop labels on sle15, so first where we can do the change is tumbleweed/slfo

dirkmueller: I'd be fine with dropping it only for openSUSE if that's what is being preferred.

dirkmueller: I've changed it to "not for openSUSE" to address the concern and rebased to avoid the merge conflict.

dcermak (commented):
Looks good to me, but given that we now use the `maintainer` attribute as a contact info, how about renaming it to `maintainer_contact`?

dirkmueller:
> Looks good to me, but given that we now use the `maintainer` attribute as a contact info, how about renaming it to `maintainer_contact`?

I would prefer if we would stick with the official key names rather than inventing new ones (the cognitive load of remembering that Dans "maintainer_contact" is "oci_authors" is not worth it for me). would that also be working?

dcermak:
> > Looks good to me, but given that we now use the `maintainer` attribute as a contact info, how about renaming it to `maintainer_contact`?
> 
> I would prefer if we would stick with the official key names rather than inventing new ones (the cognitive load of remembering that Dans "maintainer_contact" is "oci_authors" is not worth it for me). would that also be working?

`oci_authors` is even better :+1:

dirkmueller:
> `oci_authors` is even better 👍

done.

alexandrevicenzi (approved):

PR #1920, opened 2024-10-29
Set rmt-server to L3 support level

ngetahun:


ngetahun:
> is https://bugzilla.suse.com/show_bug.cgi?id=1230404 going to be fixed then?

I have no idea. I have been stuck trying to reproduce the issue on SLEM.

dirkmueller (approved):

PR #1936, opened 2024-10-30
Rmt server entrypoint improvements

dcermak:


src/bci_build/package/rmt.py
```diff
@@ -31,7 +31,7 @@
         ],
         version_in_uid=False,
         support_level=SupportLevel.L3,
-        package_list=["rmt-server", "catatonit"],
+        package_list=["rmt-server", "catatonit", "bash"],
```
alexandrevicenzi, line 34: Q: bash is not part of the base image already?

dcermak: It is, but this is a safety hatch, in case anyone decides to switch to a micro based build

alexandrevicenzi: I believe that such change could cause a choice conflict because busybox and bash would provide /bin/sh.

alexandrevicenzi (dismissed):

src/bci_build/package/rmt.py
dcermak: That's one of the reasons why I added it

src/bci_build/package/rmt-server/entrypoint.sh
```diff
@@ -23,30 +23,46 @@ MYSQL_USER="${MYSQL_USER:-rmt}"
...
+log_level:
+  rails: debug
+EOF
 
 if [ $# -eq 0 ]; then
 	set -- /usr/share/rmt/bin/rails server -e production
 fi
 
-if [ "$1" == "/usr/share/rmt/bin/rails" -a "$2" == "server" ]; then
+if [ "$1" == "/usr/share/rmt/bin/rails" ] && [ "$2" == "server" ]; then
```
dirkmueller, line 51:
```suggestion
if [ "$1" = "/usr/share/rmt/bin/rails" ] && [ "$2" = "server" ]; then
```

src/bci_build/package/rmt-server/entrypoint.sh
```diff
@@ -23,30 +23,46 @@ MYSQL_USER="${MYSQL_USER:-rmt}"
...
   echo "Create/migrate SUSE RMT database"
   pushd /usr/share/rmt > /dev/null
 	/usr/share/rmt/bin/rails db:create db:migrate RAILS_ENV=production
   popd > /dev/null
-  if [ ${SCC_SYNC} == "true" ]; then
+  if [ "${SCC_SYNC}" == "true" ]; then
```
dirkmueller, line 56:
```suggestion
  if [ "${SCC_SYNC}" = "true" ]; then
```

dirkmueller (approved):

alexandrevicenzi (approved):

ngetahun (approved):
LGTM!

PR #1982, opened 2024-11-11
Use the full container version for build_version

dcermak:


dirkmueller (commented):
we're expecting tag_version to be used in various places, e.g. the openqa scripts. I wonder if it wouldn't be simpler to use forcemultiversion instead and then go with buildrelease?

src/bci_build/registry.py
```diff
@@ -49,20 +49,32 @@ def registry_prefix(*, is_application: bool) -> str:
...
+                stability_suffix: str = ""
+                if container._stability_suffix:
+                    stability_suffix = "." + container._stability_suffix
+                return f"{base_version}.{container_version}{stability_suffix}"
+            except packaging.version.InvalidVersion:
+                return None
+
+        # first try to use the containers full version, if that fails, pick
+        # tag_version and if that also fails, fall back to the parent's classes
+        # version
```
dirkmueller, line 72: so that means switching from a `%%..`%%` based version to another version will change the behavior. is that really a good idea? seems very dangerous to me. 

I think the current setup of "version" being the %%placeholder and "tag_version" being a placeholder-free "stable" version is good enough and more predictable in behavior. I know we fscked it up for kiwi but that has been fixed and there is now a check that tag_version needs to be set if there is a placeholder in version. we could enforce it even more to have tag_version always set which I think would make the behavior more predicatble.

dirkmueller:
@dcermak is this still needed or can it be closed?

PR #2017, opened 2024-11-15
Add a SL Micro 6.0 GA kernel module development container

dirkmueller:


src/bci_build/package/basecontainers.py
```diff
@@ -330,6 +334,56 @@ def _get_minimal_kwargs(os_version: OsVersion):
...
+            ]
+        ),
+        exclusive_arch=[Arch.X86_64, Arch.S390X, Arch.AARCH64],
+        extra_files={"_constraints": generate_disk_size_constraints(8)},
+        custom_end=_get_asset_script(_SLM60_BASEURL, _SLM60_KERNEL_PACKAGES)
+        + textwrap.dedent(
+            f"""
+            {DOCKERFILE_RUN} \\
+                [ $(LC_ALL=C rpm --checksig -v *rpm | \\
+                    grep -c -E "^ *V3.*key ID 09d9ea69:") = {len(_SLM60_KERNEL_PACKAGES)} ] \\
```
alexandrevicenzi, line 378: nit: should this key number be a variable alongside `_SLM60_KERNEL_PACKAGES` for future updates?

alexandrevicenzi (approved):

src/bci_build/package/basecontainers.py
dirkmueller: It's the signing key, we're not going to change that.. but I don't care, I can make it a variable if that's wanted.

dcermak:
Has the kernel team agreed to this?

dirkmueller:
> Has the kernel team agreed to this?

the kernel team is not needed to agree to this because it pulls gcc-build not gcc. I guess your question is whether the toolchain team has agreed to it and that discussion is currently commencing. does that prevent us from merging this PR and the corresponding test suite improvements though?

so far we said that jira + commitment is only needed for releasing the container, which this PR is not doing.

dirkmueller:
`gcc-build` is now marked as L3 supported in SL Micro 6.0. @dcermak can we move forward with this?

src/bci_build/package/basecontainers.py
```diff
@@ -330,6 +334,57 @@ def _get_minimal_kwargs(os_version: OsVersion):
...
+]
+_SLFO_KEY_ID = "09d9ea69"
+KERNEL_MODULE_CONTAINERS.append(
+    OsContainer(
+        name="slm60-kernel-module-devel",
+        pretty_name="SUSE Linux Micro 6.0 Kernel module development",
+        logo_url="https://opensource.suse.com/bci/SLE_BCI_logomark_green.svg",
+        os_version=OsVersion.SLE16_0,
+        supported_until=_SUPPORTED_UNTIL_SLE.get(os_version),
+        is_latest=OsVersion.SLE16_0 in CAN_BE_LATEST_OS_VERSION,
```
dcermak, line 358: nit: not sure if this is a good idea as this will most likely make this container image `latest` once sle 16 is released, despite nothing changing for this container. It doesn't _really_ make a difference, but I guess there's no danger hardcoding this value?

dcermak (approved):
LGTM, left a minor nit

src/bci_build/package/basecontainers.py
dirkmueller: hardcoded to `True`

PR #2025, opened 2024-11-19
Set buildname/version for SLFO based development containers

dirkmueller:
prefix them with 16.0 to avoid downgrades not getting published.

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -154,9 +154,6 @@ class BaseContainerImage(abc.ABC):
     # ``devel:BCI:SLE-15-SP$ver`` (on  OBS) or ``SUSE:SLE-15-SP$ver:Update:BCI`` (on IBS)
     package_name: str | None = None
 
-    #: Epoch to use for handling os_version downgrades
-    os_epoch: int | None = None
-
     #: The container from which the build stage is running. On SLE15, this defaults to
     #: ``suse/sle15:15.$SP`` for Application Containers and ``bci/bci-base:15.$SP``
     #: for all other images. On openSUSE, ``opensuse/tumbleweed:latest`` is used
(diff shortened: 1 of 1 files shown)


src/bci_build/package/__init__.py
```diff
@@ -392,11 +392,13 @@ def build_name(self) -> str | None:
...
+        epoch = f"{self.os_epoch}." if self.os_epoch else ""
         if self.os_version.is_sle15:
-            epoch = ""
-            if self.os_epoch:
-                epoch = f"{self.os_epoch}."
             return f"15.{epoch}{int(self.os_version.value)}"
+        if self.os_version.is_slfo:
+            return (
+                f"16.{epoch}{str(float(self.os_version.value) % 1).partition('.')[2]}"
+            )
```
dcermak, line 401: This will lead to funny rounding errors if we ever bump `os_version` to `16.1`:
```
>>> str(16.1 % 1)
'0.10000000000000142'
```

How about omitting the modulo call then? You're throwing the major version away anyway
```suggestion
        if self.os_version.is_slfo:
            return (
                f"16.{epoch}{str(float(self.os_version.value)).partition('.')[2]}"
            )
```

dirkmueller: as you wish to remove epoch handling anyway, we don't need to make it that convoluted.

dirkmueller:
> This will need further work as we now set:
> 
> ```dockerfile
> #!BuildName: bci-bci-init-%OS_VERSION_ID_SP%
> ```

we've been doing that for SLE15's containers for 4 years. I'm not against fixing it, but why does this need to be done in one of the most urgently waiting pull requests and can't be done later?

dirkmueller:
> This will need further work as we now set:
> 
> ```dockerfile
> #!BuildName: bci-bci-init-%OS_VERSION_ID_SP%
> ```

I've moved this to https://github.com/SUSE/BCI-dockerfile-generator/pull/2055 but it will require changes to all the agregates and testing pipelines when merged.

dcermak (approved):

PR #2045, opened 2024-11-25
os_epoch is completely unused => remove it

dcermak:
The property `os_epoch` has only been used to set the `BuildVersion` and nowhere else. There's currently no container customizing it and even if there were, then this property should be rather a part of `OsVersion` anyway


dirkmueller:
this should have been a review comment on https://github.com/SUSE/BCI-dockerfile-generator/pull/2025 instead. included there now.

PR #2049, opened 2024-11-25
README for the git container

dirkmueller:


src/bci_build/package/git/README.md.j2
```diff
@@ -0,0 +1,36 @@
...
+Design goals of Git include speed, data integrity, and support for
+distributed, non-linear workflows — thousands of parallel branches running on
+different computers.
+
+
+## Usage
+
+This container provides the {{ image.publish_registry.vendor }} version of Git.
+
+Example of using git container:
```
alexandrevicenzi, line 20: nit: shout it be Git or git?

src/bci_build/package/git/README.md.j2
```diff
@@ -0,0 +1,36 @@
...
+
+```ShellSession
+$ podman run {{ image.pretty_reference }} git help
+```
+
+As Git requires a git repository for most operations to function, the container
```
alexandrevicenzi, line 26: nit: should it be Git repository, git repository, or just repository?

dirkmueller: yeah, I like "repository".

src/bci_build/package/git/README.md.j2
dirkmueller: changed it to upper case, not sure it is better..

src/bci_build/package/git/README.md.j2
```diff
@@ -0,0 +1,36 @@
...
+
+## Description
+
+Git is a distributed version control system that tracks
+versions of files. It is often used to control source code by programmers who
+are developing software collaboratively.
+
+Design goals of Git include speed, data integrity, and support for
+distributed, non-linear workflows — thousands of parallel branches running on
+different computers.
```
dmpop, line 13:
```suggestion
Git is a distributed version control system that tracks
versions of files. Git is primarily designed for controlling source code in collaborative software development.
```

Or something like that: short and to the point.

src/bci_build/package/git/README.md.j2
```diff
@@ -0,0 +1,36 @@
...
+
+```ShellSession
+$ podman run {{ image.pretty_reference }} git help
+```
+
+As Git requires a repository for most operations to function, the container
+does not explicitly set an entrypoint. This way you can launch the container in
+interactive mode to clone a repository and work on it. To have your changes persisted,
+you need to use a persistent volume mount on launch to avoid losing all your changes
+when exiting the container.
```
dmpop, line 30:
```suggestion
As Git requires a repository, the container
does not explicitly set an entrypoint. This way, you can launch the container in
interactive mode to clone a repository and work on it. To avoid losing all your changes when exiting the container, use a persistent volume mount on launch.
```

dcermak (approved):

PR #2055, opened 2024-11-25
Mark the init-image as a singleton container

dirkmueller:
This removes the "buildname" suffix that was considered an issue, but of course will break our entire CI testing and publishing pipeline.

### DIFF
--- a/src/bci_build/package/basecontainers.py
+++ b/src/bci_build/package/basecontainers.py
@@ -65,6 +65,10 @@ INIT_CONTAINERS = [
         pretty_name=f"{os_version.pretty_os_version_no_dash} Init",
         custom_description="Systemd environment for containers {based_on_container}. {podman_only}",
         package_list=["systemd", "gzip", *os_version.release_package_names],
+        is_singleton_image=(
+            # preserve backwards compatibility on already released distributions
+            os_version not in (OsVersion.SP6, OsVersion.TUMBLEWEED)
+        ),
         cmd=["/usr/lib/systemd/systemd"],
         extra_labels={
             "usage": "This container should only be used to build containers for daemons. Add your packages and enable services using systemctl."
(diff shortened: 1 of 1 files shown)


dcermak:
> This removes the "buildname" suffix that was considered an issue, but of course will break our entire CI testing and publishing pipeline.

What is the issue that this is solving? The price of breaking our entire pipeline sounds pretty high tbh. Also we now cannot distinguish the containers of different service packs via the buildname anymore.

dirkmueller:
> What is the issue that this is solving?

I don't know. It was considered a blocking reason in https://github.com/SUSE/BCI-dockerfile-generator/pull/2025#issuecomment-2497846462 . :shrug: 

> The price of breaking our entire pipeline sounds pretty high tbh.

It's not difficult to fix the breakage, and it is a minor asymmetry in the setup. 

> Also we now cannot distinguish the containers of different service packs via the buildname anymore.

We never needed that property, and no other base container provides it. so why would it be important for bci-init specifically?

(Note I see we made the same mistake for the `base-fips` container, will fix).

dirkmueller:
We can't actually fix this on released codestreams because changing the buildname resets the buildcounter. so limited to 16.0 / slfo for now. which hopefully addresses the original concern. will still break the publishing pipeline for 16.0 but nobody cares about that right now.

src/bci_build/package/basecontainers.py
```diff
@@ -65,6 +65,10 @@
         pretty_name=f"{os_version.pretty_os_version_no_dash} Init",
         custom_description="Systemd environment for containers {based_on_container}. {podman_only}",
         package_list=["systemd", "gzip", *os_version.release_package_names],
+        is_singleton_image=(
+            # preserve backwards compatibility on already released distributions
+            False if os_version in (OsVersion.SP6, OsVersion.TUMBLEWEED) else True
```
dcermak, line 70:
```suggestion
            os_version not in (OsVersion.SP6, OsVersion.TUMBLEWEED)
```

src/bci_build/package/basecontainers.py
```diff
@@ -165,6 +169,19 @@ def _get_supported_until_fips(os_version: OsVersion) -> datetime.date:
...
+            False
+            if os_version
+            in (
+                OsVersion.SP3,
+                OsVersion.SP4,
+                OsVersion.SP5,
+                OsVersion.SP6,
+                OsVersion.TUMBLEWEED,
+            )
+            else True
```
dcermak, line 183:
```suggestion
            os_version not in (
                OsVersion.SP3,
                OsVersion.SP4,
                OsVersion.SP5,
                OsVersion.SP6,
                OsVersion.TUMBLEWEED,
            )
```

dcermak (commented):
This also appears to change SP5 for init

dirkmueller:
> This also appears to change SP5 for init

no it doesn't because SP5 is EOL. there is just a false message in this PR because the PR is from times where SP5 was still a thing. init image does not exist for SP5 anymore.

alexandrevicenzi (approved):

PR #2059, opened 2024-11-26
Add PostgreSQL 17 containers

dirkmueller:


dirkmueller:
There's a failure on SP7, but that should disappear with the next SP7 product milestone, so I think we can ignore that for a bit.

PR #2062, opened 2024-11-26
Remove patterns-base-fips from SLFO base containers

dirkmueller:
This is pulling crypto-policies and we have base-fips now instead.

### DIFF
--- a/src/bci_build/package/basecontainers.py
+++ b/src/bci_build/package/basecontainers.py
@@ -115,10 +115,9 @@ def _get_fips_base_custom_end(os_version: OsVersion) -> str:
             bins = _FIPS_15_SP2_BINARIES
         case OsVersion.SP4:
             bins = _FIPS_15_SP4_BINARIES
-        case OsVersion.SP5 | OsVersion.SP6:
-            pass
-        case _:
-            raise NotImplementedError(f"Unsupported os_version: {os_version}")
+
+    if os_version not in ALL_BASE_OS_VERSIONS:
+        raise NotImplementedError(f"Unsupported os_version: {os_version}")
 
     custom_install_bins: str = textwrap.dedent(
         f"""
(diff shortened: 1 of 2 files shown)


src/bci_build/package/basecontainers.py
```diff
@@ -173,11 +173,14 @@ def _get_supported_until_fips(os_version: OsVersion) -> datetime.date:
...
-            if os_version == OsVersion.SP3
-            else ["crypto-policies-scripts"]
+        package_list=(
+            [*os_version.release_package_names, "coreutils"]
+            + (
+                ["fipscheck"]
+                if os_version == OsVersion.SP3
+                else ["crypto-policies-scripts"]
+            )
+            + (["patterns-base-fips"] if os_version.is_slfo else [])
```
dcermak, line 183: Note that this has no effect atm, as this container is not being built for SLE 16

msmeissn (commented):
okay

src/bci_build/package/basecontainers.py
```diff
@@ -144,7 +144,13 @@ def _get_fips_pretty_name(os_version: OsVersion) -> str:
...
+        case (
+            OsVersion.SP4
+            | OsVersion.SP5
+            | OsVersion.SP6
+            | OsVersion.SP7
+            | OsVersion.SLE16_0
+        ):
             return f"{os_version.pretty_os_version_no_dash} FIPS-140-3"
         case _:
             raise NotImplementedError(f"Unsupported os_version: {os_version}")
```
dcermak, line 156: Wouldn't this be a bit simpler:
```Python
def _get_fips_pretty_name(os_version: OsVersion) -> str:
    if os_version == OsVersion.SP3:
        return f"{os_version.pretty_os_version_no_dash} FIPS-140-2"
    if os_version.is_sle15 or os_version.is_slfo:
        return f"{os_version.pretty_os_version_no_dash} FIPS-140-3"

    raise NotImplementedError(f"Unsupported os_version: {os_version}")
```

dirkmueller: yes, we can do that.

dcermak (approved):

PR #2069, opened 2024-11-26
🤖: Update build recipes from templates

dirkmueller:


dirkmueller:
/vc add buildname setting for stable container aggregation

PR #2077, opened 2024-11-27
Add a pip convenience symlink also for the python 3.9 container

dirkmueller:
the python3.9-pip package does not provide it for historic reasons, and the testsuite expects it to be there (rightfully).

### DIFF
--- a/src/bci_build/package/python.py
+++ b/src/bci_build/package/python.py
@@ -97,6 +97,12 @@ def _get_python_kwargs(py3_ver: _PYTHON_VERSIONS, os_version: OsVersion):
         config_sh_script += rf"""; ln -s /usr/bin/python{py3_ver} /usr/local/bin/python3; \
     ln -s /usr/bin/pydoc{py3_ver} /usr/local/bin/pydoc"""
 
+    # broken packaging without a "pip" update-alternatives link
+    if py3_ver in ("3.9",):
+        config_sh_script += (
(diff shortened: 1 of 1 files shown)


src/bci_build/package/python.py
```diff
@@ -97,6 +97,10 @@ def _get_python_kwargs(py3_ver: _PYTHON_VERSIONS, os_version: OsVersion):
         config_sh_script += rf"""; ln -s /usr/bin/python{py3_ver} /usr/local/bin/python3; \
     ln -s /usr/bin/pydoc{py3_ver} /usr/local/bin/pydoc"""
 
+    # broken packaging without a "pip" update-alternatives link
+    if py3_ver in ("3.9",):
+        config_sh_script += rf"""; test -e /usr/bin/pip || ln -s /usr/bin/pip{py3_ver} /usr/local/bin/pip"""
```
dcermak, line 102: Do we need the `r` or the multiline string part?
```suggestion
        config_sh_script += f"; test -e /usr/bin/pip || ln -s /usr/bin/pip{py3_ver} /usr/local/bin/pip"
```

dirkmueller:
> LGTM, just one thing to consider: do we want to convert this script so that it fails once the symlink is in place? Otherwise I'm afraid it'll linger around for all eternity ™️

I don't expect the underlying bug, which is python 3.9 specific, to be fixed as there is an explicit rm on the pip link in the package and no update-alternatives hooks which all the other pip variants have.

dcermak (approved):

PR #2085, opened 2024-11-28
Move SAC Python containers to multibuild

dirkmueller:
This is needed to be able to provide multiple branches. Add 3.9 and 3.11.


src/bci_build/containercrate.py
```diff
@@ -1,34 +1,54 @@
...
 
-class ContainerCrate:
-    """ContainerCrate is combining multiple container build flavors.
+from bci_build.util import write_to_file
 
-    This provides package-central functions like generating _service and
-    _multibuild files.
-    """
+
+class ContainerCrateAssigner:
```
dcermak, line 8: Can we please stop adding more magic indirection to the code? I already voiced my disagreement with this implicit magic and it keeps getting worse.

dirkmueller: This is a draft PR. the one that needs to move ahead is https://github.com/SUSE/BCI-dockerfile-generator/pull/2096 instead.

PR #2094, opened 2024-11-29
Order attributes alphabetically

dirkmueller:
This is more stable and consistent with other tags which were already doing so.

### DIFF
--- a/src/bci_build/service.py
+++ b/src/bci_build/service.py
@@ -29,7 +29,7 @@ class Service:
         :py:class:`~xml.etree.ElementTree.Element`.
 
         """
-        root = ET.Element("service", attrib={"name": self.name, "mode": self.mode})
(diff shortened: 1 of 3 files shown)


dirkmueller:
> This is more stable and consistent with other tags which were already doing so.

This is more stable and consistent with other tags which were already doing so. There was no rationale given in https://github.com/SUSE/BCI-dockerfile-generator/pull/1815/commits/a12656f0280d794d766905d013ab6732ae641d9e either :-/

what I can see is that here: https://github.com/SUSE/BCI-dockerfile-generator/pull/1815/commits/a12656f0280d794d766905d013ab6732ae641d9e#diff-d7fb70d71c01e6c6a932249a22ad45e9e8e5dd800f8d9260922b7726ada72402L108-R109 all the names are now the first attribute, and mode is consistently the 2nd, making it easier to scan over for the human eye.

dcermak (approved):

PR #2096, opened 2024-12-01
Add SAC Python containers in addition to the normal python containers (jsc#PED-11663)

dirkmueller:
We need to set a buildcountersynctag for them to satisfy branch-listing requirements.

Add 3.9 and 3.11.

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -236,6 +236,11 @@ class BaseContainerImage(abc.ABC):
     #: contain arbitrary instructions valid for a :file:`Dockerfile`.
     build_stage_custom_end: str | None = None
 
+    #: This string defines which build counter identifier should be used for this
+    #: container. can be an arbitrary string is used to derive the checkin-counter.
+    #: Defaults to the main package for packages with a flavor.
+    buildcounter_synctag: str | None = None
+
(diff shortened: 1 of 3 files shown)


alexandrevicenzi (approved):

src/bci_build/package/__init__.py
```diff
@@ -1226,7 +1235,16 @@ def _tag_variant(self) -> str:
...
+        registry = (
+            "-sac"
+            if isinstance(self._publish_registry, ApplicationCollectionRegistry)
+            else ""
+        )
+        return (
+            f"{self.name}-{self._tag_variant}{registry}"
+            if self.version_in_uid
+            else self.name
+        )
```
dcermak, line 1247: Please don't do this, it'll break the crb yaml PR

dirkmueller: What should I be doing instead? I need to generate two python 3.11 containers, one for registry and one for SAC. I don't see a way other than doing a special version of the uid. 

Why does this break the crb generator and why can this not be fixed? the yaml generator shouldn't care about the uid (or if it generates this into the yaml key, then we need to adjust the keys accordingly)?

dirkmueller: would a prefix be helping instead of a suffix?

dcermak (approved):

PR #2097, opened 2024-12-02
Use List comprehension for defining containers

dirkmueller:
- Also avoid mkdir -p as it doesn't define the permissions
- set KEA_PIDFILE_DIR
- don't set latest tag

### DIFF
--- a/src/bci_build/package/kea.py
+++ b/src/bci_build/package/kea.py
@@ -1,5 +1,10 @@
 """KEA DHCP Server BCI container"""
 
+import textwrap
+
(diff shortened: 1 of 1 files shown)


src/bci_build/package/postgres.py
dcermak, line 1: Could you please make this a separate commit?

dirkmueller: it's already in https://github.com/SUSE/BCI-dockerfile-generator/pull/2095

dcermak (approved):

PR #2100, opened 2024-12-02
Disable protected-mode in valkey

dirkmueller:
upstream does this by recompiling valkey which is not something that we can easily do. let's see how far we'll get by patching the configuration file.


src/bci_build/package/valkey.py
```diff
@@ -39,8 +39,8 @@
             )
         ],
         license="BSD-3-Clause",
-        package_list=["valkey"],
-        entrypoint=["/usr/bin/valkey-server"],
+        package_list=["valkey", "sed"],
+        entrypoint=["/usr/bin/valkey-server", "/etc/valkey/valkey.conf"],
```
dcermak, line 43: Wouldn't it make sense to keep entrypoint as it is and add `"/etc/valkey/valkey.conf"` as `CMD`?

src/bci_build/package/valkey.py
```diff
@@ -39,8 +39,8 @@
             )
         ],
         license="BSD-3-Clause",
-        package_list=["valkey"],
-        entrypoint=["/usr/bin/valkey-server"],
+        package_list=["valkey", "sed"],
```
dcermak, line 42:
```suggestion
        package_list=["valkey"],
```

src/bci_build/package/valkey.py
```diff
@@ -49,6 +49,7 @@
         ),
         custom_end=textwrap.dedent(
             f"""
+            {DOCKERFILE_RUN} sed -e 's/^protected-mode yes/protected-mode no/' -e 's/^bind .*//' < /etc/valkey/default.conf.example > /etc/valkey/valkey.conf
```
dcermak, line 52:
```suggestion
            {DOCKERFILE_RUN} while IFS= read -r line; do \
              line="${line/#protected-mode yes/protected-mode no}"; \
              line="${line/#bind /}"; \
              printf '%s\n' "$line"; \
            done < /etc/valkey/default.conf.example > /etc/valkey/valkey.conf
```

dcermak (commented):
We can avoid adding the `sed` dependency. Also, why do we want to disable protected mode if it's the distribution default?

src/bci_build/package/valkey.py
dirkmueller: theoretically, RUN statements should be posix shell and this isn't posix. given that sed is not that big, do we need to optimize it that way? I don't expect this to be a perfect permanent fix anyway.

dcermak: What's not POSIX about this? It does the job with Tumbleweed's sh

dirkmueller: > What's not POSIX about this? 

```
line="${line/#protected-mode yes/protected-mode no}";
```

there is no shell builtin string replacement in posix sh. and Tumbleweeds "sh" is bash in bash mode, not in posix mode.

dcermak (approved):

alexandrevicenzi (approved):
LGTM.

I think sed is negligible in size and more readable.

PR #2109, opened 2024-12-03
Drop jshell as the CMD for openjdk-devel

dcermak:
`jshell` is broken on ppc64le for mysterious reasons and it's not really a useful `CMD` anyway, so let's just stop using it by default until someone complains

### DIFF
--- a/src/bci_build/package/openjdk-devel/README.md.j2
+++ b/src/bci_build/package/openjdk-devel/README.md.j2
@@ -10,14 +10,14 @@ The OpenJDK development image is intended to be used as a build environment. For
 
 ## Usage
 
-The default command for the image is the Java Shell tool (JShell).
+{% if image.os_version | string in ("5", "6") %}The default command for the image is the Java Shell tool (JShell).
 
 ```ShellSession
 $ podman run -it --rm {{ image.pretty_reference }}
 jshell> /help
 ```
 
-To compile and deploy an application, copy the sources and build the binary:
+{% endif %}To compile and deploy an application, copy the sources and build the binary:
 
 ```Dockerfile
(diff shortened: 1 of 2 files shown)


alexandrevicenzi (approved):

dirkmueller (changes requested):
We should be limiting that change to sle16 or sle15-sp7+, and not change behavior in a maintenance update without any good reason (and ppc64le is not a good reason). 

also the README needs adjustments.

src/bci_build/package/openjdk.py
```diff
@@ -61,12 +63,16 @@ def _get_openjdk_kwargs(
     }
 
     if devel:
+        # don't set CMD in SP6 onward as jshell is broken and the CMD is
```
dirkmueller, line 66: SP6 is already released, did you mean SP7?

dcermak: I meant SP6, because it's already broken on SP6

dirkmueller: not on platforms other than ppc64le. I think we should do behavior changes only with unreleased SPs, not within the maintenance period of a released SP.

src/bci_build/package/openjdk-devel/README.md.j2
```diff
@@ -10,14 +10,14 @@ The OpenJDK development image is intended to be used as a build environment. For
 
 ## Usage
 
-The default command for the image is the Java Shell tool (JShell).
+{% if image.os_version | string == "5" %}The default command for the image is the Java Shell tool (JShell).
```
dirkmueller, line 13: this needs to be adjusted as well

dirkmueller (approved):

PR #2110, opened 2024-12-03
add stunnel container image

dcermak:
https://jira.suse.com/browse/PED-11085

### DIFF
--- a/src/bci_build/package/stunnel.py
+++ b/src/bci_build/package/stunnel.py
@@ -1,0 +1,48 @@
+"""Build description for the stunnel BCI."""
+
+from pathlib import Path
+
+from bci_build.os_version import ALL_NONBASE_OS_VERSIONS
+from bci_build.os_version import CAN_BE_LATEST_OS_VERSION
+from bci_build.package import DOCKERFILE_RUN
+from bci_build.package import ApplicationStackContainer
+from bci_build.package import ParseVersion
+from bci_build.package import Replacement
+from bci_build.package.helpers import generate_from_image_tag
+from bci_build.package.helpers import generate_package_version_check
+from bci_build.registry import publish_registry
+
+STUNNEL_CONTAINERS = [
+    ApplicationStackContainer(
+        name="stunnel",
+        os_version=os_version,
+        tag_version=(tag_ver := "5"),
+        is_latest=os_version in CAN_BE_LATEST_OS_VERSION,
+        from_target_image=generate_from_image_tag(os_version, "bci-micro"),
+        version=(stunnel_version_re := "%%stunnel_re%%"),
+        pretty_name="Stunnel",
+        package_list=["stunnel"],
+        replacements_via_service=[
+            Replacement(stunnel_version_re, package_name="stunnel")
+        ],
+        extra_files={
+            "entrypoint.sh": (
+                (stunnel_dir := Path(__file__).parent / "stunnel") / "entrypoint.sh"
+            ).read_bytes(),
+            "stunnel.conf": (stunnel_dir / "stunnel.conf").read_bytes(),
+        },
+        _publish_registry=publish_registry(os_version, app_collection=True),
+        build_stage_custom_end=generate_package_version_check(
+            "stunnel", tag_ver, ParseVersion.MAJOR, use_target=True
+        ),
+        custom_end=f"""COPY entrypoint.sh /usr/local/bin/
+COPY stunnel.conf /etc/stunnel/stunnel.conf
+{DOCKERFILE_RUN} chmod 0755 /usr/local/bin/entrypoint.sh; \
+    chown --recursive stunnel /etc/stunnel
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
+CMD ["/usr/sbin/stunnel"]
+USER stunnel
+""",
+    )
+    for os_version in ALL_NONBASE_OS_VERSIONS
+]
--- a/src/bci_build/package/stunnel/README.md.j2
+++ b/src/bci_build/package/stunnel/README.md.j2
@@ -1,0 +1,80 @@
+# {{ image.pretty_name }} Container Image
+
+{% include 'badges.j2' %}
+
+
+## Description
+
+Stunnel is an open-source multi-platform application that provides a universal
+TLS/SSL tunneling service.
+
+
+## How to use this image
+
+By default, the Stunnel container image launches `stunnel` using a minimal
+configuration file that specifies the following:
+- run in foreground
+- load further configuration files from `/etc/stunnel/conf.d`
+
+Custom configuration files must be placed into the directory
+`/etc/stunnel/conf.d`.
+
+The container entry point configures TLS/SSL automatically by setting the key
+and certificate to the values of the environment variables `STUNNEL_KEY` and
+`STUNNEL_CERT`. If one of the environment variables is unset, then the
+entrypoint defaults to `/etc/stunnel/stunnel.key` for `STUNNEL_KEY` and
+`/etc/stunnel/stunnel.pem` for `STUNNEL_CERT`.
+
+The entrypoint can set up a single service via environment variables, so that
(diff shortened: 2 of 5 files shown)


src/bci_build/package/appcontainers.py
```diff
@@ -442,3 +443,36 @@ def _get_nginx_kwargs(os_version: OsVersion):
     )
     for os_version in (OsVersion.TUMBLEWEED,)
 ]
+
+STUNNEL_CONTAINERS = [
```
dirkmueller, line 447: please create a separate stunnel.py toplevel and not add further containers to appcontainers (I apologize for having started this, we definitely shouldn't continue with it)

src/bci_build/package/stunnel/README.md.j2
```diff
@@ -0,0 +1,71 @@
+# {{ image.pretty_name }} Container Image
+
+{% include 'badges.j2' %}
+
+
+## Description
+
+Stunnel is an open-source multi-platform application used to provide a universal
+TLS/SSL tunneling service.
```
dmpop, line 9:
```suggestion
Stunnel is an open-source multi-platform application that provides a universal
TLS/SSL tunneling service.
```

src/bci_build/package/stunnel/README.md.j2
```diff
@@ -0,0 +1,71 @@
...
+
+
+## How to use this Image
```
dmpop, line 12:
```suggestion
## How to use this image
```

src/bci_build/package/stunnel/README.md.j2
```diff
@@ -0,0 +1,71 @@
...
+
+The Stunnel Container Image launches `stunnel` by default using a minimal
+configuration file that only sets:
+- run in foreground
+- load further configuration files from `/etc/stunnel/conf.d`
```
dmpop, line 17:
```suggestion
By default, the Stunnel container image launches `stunnel` using a minimal
configuration file that specifies the following:
- run in foreground
- load further configuration files from `/etc/stunnel/conf.d`
```

src/bci_build/package/stunnel/README.md.j2
```diff
@@ -0,0 +1,71 @@
...
+
+Custom configuration files should be placed into the directory
+`/etc/stunnel/conf.d`.
```
dmpop, line 20:
```suggestion
Custom configuration files must be placed into the directory
`/etc/stunnel/conf.d`.
```

src/bci_build/package/stunnel/README.md.j2
```diff
@@ -0,0 +1,71 @@
...
+The container entry point configures TLS/SSL automatically by setting the key
+and certificate to the values of the environment variables `STUNNEL_KEY` and
+`STUNNEL_CERT`. If one of the environment variables is unset, then the
+entrypoint defaults to `/etc/stunnel/stunnel.key` for `STUNNEL_KEY` and
+`/etc/stunnel/stunnel.pem` for `STUNNEL_CERT`.
+
+The entrypoint can setup a single service via environment variables so that the
+user doesn't have to write and mount their own configuration file. This can be
+setup via the environment variables `STUNNEL_SERVICE_NAME`, `STUNNEL_ACCEPT` and
+`STUNNEL_CONNECT`:
```
dmpop, line 31:
```suggestion
The entrypoint can set up a single service via environment variables, so that the
user doesn't have to write and mount their own configuration file. This can be
specified via the environment variables `STUNNEL_SERVICE_NAME`, `STUNNEL_ACCEPT` and
`STUNNEL_CONNECT`:
```

src/bci_build/package/stunnel/README.md.j2
```diff
@@ -0,0 +1,71 @@
...
+
+- `STUNNEL_SERVICE_NAME`: name or otherwise unique identifier of the service
+  (used for documentation purpose only)
+
+- `STUNNEL_ACCEPT`: address on which new connections should be accepted, can be
+  either a hostname + port number or just a port number (then localhost is
+  assumed as the host)
```
dmpop, line 38:
```suggestion
- `STUNNEL_ACCEPT`: address on which new connections should be accepted. It can be
  either a hostname and a port number or just a port number (in which case, localhost is
  assumed to be the host)
```

src/bci_build/package/stunnel/README.md.j2
```diff
@@ -0,0 +1,71 @@
...
+
+- `STUNNEL_CONNECT`: address on which the un-encrypted service is listening and
+  to which stunnel connects, can be either a hostname + port number or just a
+  port number (then localhost is assumed as the host)
```
dmpop, line 42:
```suggestion
- `STUNNEL_CONNECT`: address on which the unencrypted service is listening and
  to which stunnel connects. It can be either a hostname and port number or just a
  port number (in which case, localhost is assumed to be the host)
```

src/bci_build/package/stunnel/README.md.j2
```diff
@@ -0,0 +1,71 @@
...
+
+
+For example to create an SSL endpoint for a webserver listening on port `8000`
+on localhost, execute:
```
dmpop, line 46:
```suggestion
For example, to create an SSL endpoint for a webserver listening on port `8000`
on localhost, run the following command:
```

src/bci_build/package/stunnel/README.md.j2
```diff
@@ -0,0 +1,71 @@
...
+    {{ image.pretty_reference }}
+```
+
+
+### Pitfalls
+
+The Stunnel Container Image is configured to launch `stunnel` as the `stunnel`
+user. Files mounted into a running container however belong to the `root` user
+by default. Set the file permissions of mounted files accordingly, so that
+non-owners and non-group members can read them.
```
dmpop, line 65:
```suggestion
The Stunnel container image is configured to launch `stunnel` as the `stunnel`
user. But by default, files mounted into a running container belong to the `root` user. Set the file permissions of mounted files accordingly, so that
non-owners and non-group members can read them.
```

src/bci_build/package/stunnel/README.md.j2
```diff
@@ -0,0 +1,71 @@
...
+
+Stunnel's `inetd` mode is not supported in this Container Image, as it is very
+minimal and does not ship a package manager to install any relevant services.
```
dmpop, line 68:
```suggestion
Stunnel's `inetd` mode is not supported in the container image, and it does not ship a package manager for installing any services.
```

src/bci_build/package/stunnel.py
```diff
@@ -0,0 +1,48 @@
...
+            "stunnel.conf": (stunnel_dir / "stunnel.conf").read_bytes(),
+        },
+        _publish_registry=publish_registry(os_version, app_collection=True),
+        build_stage_custom_end=generate_package_version_check(
+            "stunnel", tag_ver, ParseVersion.MAJOR, use_target=True
+        ),
+        custom_end=f"""COPY entrypoint.sh /usr/local/bin/
+COPY stunnel.conf /etc/stunnel/stunnel.conf
+{DOCKERFILE_RUN} chmod 0755 /usr/local/bin/entrypoint.sh; \
+    chown --recursive stunnel /etc/stunnel
```
dirkmueller, line 41: shouldn't this just be on the `stunnel.conf` file rather than the directory as well?

dcermak: No, I think it would be better if the `stunnel` user owns the whole directory. The entrypoint will write settings into `/etc/stunnel/conf.d/`, so it makes imho sense for the directory to be owned by the stunnel user. We could resort to just the `conf.d` subdir, but I see little benefit in that (and a lot of potential inconveniences).

dirkmueller: I don't like directories being owned by a service user as that allows that service user to modify permissions. probably not a big concern here but in general a owner=root,group=servicegroup,mod 0750 or so makes more sense to me. anyhow not a blocking issue.

PR #2127, opened 2024-12-09
Helm version check

dirkmueller:
Fix helm version handling
   
Ensure we fail the build when the version number is not the expected one. use the full version as tag and also set a tag_version.

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -1456,7 +1456,6 @@ from .apache_tomcat import TOMCAT_CONTAINERS  # noqa: E402
 from .appcontainers import ALERTMANAGER_CONTAINERS  # noqa: E402
 from .appcontainers import BLACKBOX_EXPORTER_CONTAINERS  # noqa: E402
 from .appcontainers import GRAFANA_CONTAINERS  # noqa: E402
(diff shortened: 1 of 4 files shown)


dcermak (commented):
Why the switch to the multistage dockerfile based build? It increases the image size:
```
❯ podman inspect -f "{{.VirtualSize}}" registry.opensuse.org/home/defolos/bci/staging/sle-15-sp6/6-2127/containerfile/suse/helm:latest
107281600
❯ podman inspect -f "{{.VirtualSize}}" registry.suse.com/suse/helm
102012988
```
(that's about 5%)

dirkmueller:
> Why the switch to the multistage dockerfile based build?

See the commit message:

```
This actually makes the package version check work because with
kiwi we can not run config.sh until it is too late.
```

> It increases the image size:

The image size mostly changed because we now embed a newer version of helm, which got fatter.  if you check closely, the image actually got smaller: 

```
$ podman inspect -f "{{.VirtualSize}}" registry.opensuse.org/devel/bci/sle-15-sp6/images/suse/helm:latest
108563429
$ podman inspect -f "{{.VirtualSize}}" registry.opensuse.org/home/defolos/bci/staging/sle-15-sp6/6-2127/containerfile/suse/helm:latest
107283430
```

That's 1.18% smaller.

PR #2134, opened 2024-12-10
Rename gosu

### LINKED ISSUE #2070: replace gosu "script" in mariadb containers with real gosu or at least a fully compatible wrapper
from https://github.com/SUSE/BCI-dockerfile-generator/pull/963/files#r1859152125

we should either switch to https://github.com/tianon/gosu/blob/3c1fd16340edbc0e984ba4674379da23ebb82abf/setpriv-wrapper.sh or simply give our shell script (which is similar but not the same) a different name to avoid confusion.

alexandrevicenzi:
The current gosu script is not a replacement for gosu, because of that it is better to rename to a completely new name to avoid confusion and compatibility issues.

Fixes #2070


src/bci_build/package/mariadb.py
```diff
@@ -129,8 +129,12 @@
 COPY healthcheck.sh /usr/local/bin/
 {DOCKERFILE_RUN} chmod 755 /usr/local/bin/healthcheck.sh
 
-COPY gosu /usr/local/bin/gosu
-{DOCKERFILE_RUN} chmod 755 /usr/local/bin/gosu
+COPY idexec /usr/local/bin/idexec
+{DOCKERFILE_RUN} chmod 755 /usr/local/bin/idexec
+
+# replace gosu calls with idexec
+{DOCKERFILE_RUN} sed -i 's/gosu/idexec/g' /usr/local/bin/{_ENTRYPOINT_FNAME}
```
dirkmueller, line 136: can we make this more robust by e.g. adding spaces or so around it? I'm nervous that it will replace some other occurence somewhere else and then we have a very difficult to trace bug

dirkmueller: e.g "s/ exec gosu / exec idexec /" 

would work for me.

alexandrevicenzi: There are not many calls to gosu, nor anything with gosu in the middle of a word, but I get the point.

dcermak (approved):

tianon:
Thank you! ❤️

src/bci_build/package/mariadb.py
dirkmueller: > There are not many calls to gosu, nor anything with gosu in the middle of a word, but I get the point.

not today, but there might be some in the future

PR #2137, opened 2024-12-11
Tomcat without curl & sed

dcermak:
- We don't need curl as of https://github.com/SUSE/BCI-tests/pull/699 anymore
- Also, we can remove sed with a (arguably yucky) bash builtin
- Fix the logging config path from `/etc/tomcat/logging.properties` to `/usr/share/tomcat/conf/logging.properties`


dirkmueller:
sed is still in the container because of the start scripts of tomcat requiring it. lets not go crazy.

dirkmueller (approved):

PR #2179, opened 2025-01-09
Sync changes file with production changes

dirkmueller:


dcermak (approved):
Are we making Darix happy again?

dirkmueller:
> Are we making Darix happy again?

well, lost bugreferences and so on. anyhow it is nicer that way.

PR #2199, opened 2025-01-15
Refactoring container crate into container family to make it more readable

rcmadhankumar:


src/bci_build/package/__init__.py
```diff
@@ -1435,6 +1441,97 @@ def prepare_template(self) -> None:
...
+            BaseContainerImage | DevelopmentContainer | ApplicationStackContainer
+        ],
+    ):
+        """Assign the family for every container build flavor based on os version and package name
+
+        Sample family structure:
+        {
+          (OsVersion.TumbleWeed, "test-package-name"): {"flavor1", "flavor2"}
+        }
+        """
```
dcermak, line 1463: This will be interpreted as the docstring of `__init__`, it should be just an inline comment using `#`

src/bci_build/package/__init__.py
```diff
@@ -1435,6 +1441,97 @@ def prepare_template(self) -> None:
...
+    """ContainerFamily is grouping multiple containers build flavors.
+
+    This provides package-central functions like generating _service and
+    _multibuild files, checking version_uid
+    """
+
+    def __init__(
+        self,
+        containers: Sequence[
+            BaseContainerImage | DevelopmentContainer | ApplicationStackContainer
```
dcermak, line 1454: `BaseContainerImage` is the base class of the latter two:
```suggestion
            BaseContainerImage
```

src/bci_build/package/__init__.py
```diff
@@ -1435,6 +1441,97 @@ def prepare_template(self) -> None:
...
+        for container in containers:
+            if container.family is not None:
+                raise ValueError("Container is already part of a ContainerFamily")
+            container.family = self
+
+    def get_all_build_flavors(
+        self,
+        container: BaseContainerImage
+        | DevelopmentContainer
+        | ApplicationStackContainer,
```
dcermak, line 1480:
```suggestion
        container: BaseContainerImage,
```

src/bci_build/package/__init__.py
```diff
@@ -1435,6 +1441,97 @@ def prepare_template(self) -> None:
...
+            self._container_families.get(
+                (container.os_version, container.package_name), [""]
+            )
+        )
+
+    def get_default_dockerfile_content(
+        self,
+        container: BaseContainerImage
+        | DevelopmentContainer
+        | ApplicationStackContainer,
```
dcermak, line 1493:
```suggestion
        container: BaseContainerImage,
```

src/bci_build/package/__init__.py
```diff
@@ -1435,6 +1441,97 @@ def prepare_template(self) -> None:
...
+#!ForceMultiVersion{buildrelease}
+
+# For this container we only build the Dockerfile.$flavor builds.
+"""
+
+    def get_multibuild_file_content(
+        self,
+        container: BaseContainerImage
+        | DevelopmentContainer
+        | ApplicationStackContainer,
```
dcermak, line 1509:
```suggestion
        container: BaseContainerImage,
```

src/bci_build/package/__init__.py
```diff
@@ -1435,6 +1441,97 @@ def prepare_template(self) -> None:
...
+            " " * 4 + f"<package>{pkg}</package>"
+            for pkg in self.get_all_build_flavors(container)
+        )
+        return f"<multibuild>\n{flavors}\n</multibuild>"
+
+    def check_version_in_uid(
+        self,
+        container: BaseContainerImage
+        | DevelopmentContainer
+        | ApplicationStackContainer,
```
dcermak, line 1525:
```suggestion
        container: BaseContainerImage,
```

src/bci_build/package/__init__.py
```diff
@@ -1435,6 +1441,97 @@ def prepare_template(self) -> None:
...
+    ) -> bool:
+        """.dockerignore"""
```
dcermak, line 1527: This one needs a better docstring :-D

dcermak (commented):
This is not yet what I was anticipating, currently this is just a bit more a rename of the `ContainerCrate` class.

What I'd like to see (eventually, we don't have to get there in a single PR):
- crate/family shouldn't start mutating containers passed to it
- the bot & package file writing code should switch to use ContainerFamily directly

Also, this currently is breaking multibuild generation: https://github.com/SUSE/BCI-dockerfile-generator/commit/deb6b4aa2f0dc4b5e813dcb958752d852b1f6841#diff-25c1eeff6e17b5c4eaac44d821f42827a798b8c8d36f6ef005edb1ae7771f041R84

src/bci_build/package/__init__.py
rcmadhankumar: fixed.

src/bci_build/package/__init__.py
rcmadhankumar: fixed.

src/bci_build/package/__init__.py
rcmadhankumar: fixed.

src/bci_build/package/__init__.py
rcmadhankumar: fixed.

src/bci_build/package/__init__.py
rcmadhankumar: fixed.

src/bci_build/package/__init__.py
rcmadhankumar: Docstring modified.

src/bci_build/package/__init__.py
```diff
@@ -1435,6 +1441,85 @@ def prepare_template(self) -> None:
         pass
 
 
+class ContainerFamily:
```
dirkmueller, line 1444: Hey @rcmadhankumar . I do like the new name suggestion "Family" instead of "Crate". However I was going into a slightly different route in https://github.com/SUSE/BCI-dockerfile-generator/pull/2085 - perhaps it would make sense to revive that with the updated name and rebase it instead?

dirkmueller:
> What I'd like to see (eventually, we don't have to get there in a single PR):
> * the bot & package file writing code should switch to use ContainerFamily directly

This is what #2085 is doing..

rcmadhankumar:
> This is not yet what I was anticipating, currently this is just a bit more a rename of the `ContainerCrate` class.
> 
> What I'd like to see (eventually, we don't have to get there in a single PR):
> 
>     * crate/family shouldn't start mutating containers passed to it
> 
>     * the bot & package file writing code should switch to use ContainerFamily directly
> 
> 
> Also, this currently is breaking multibuild generation: [deb6b4a#diff-25c1eeff6e17b5c4eaac44d821f42827a798b8c8d36f6ef005edb1ae7771f041R84](https://github.com/SUSE/BCI-dockerfile-generator/commit/deb6b4aa2f0dc4b5e813dcb958752d852b1f6841#diff-25c1eeff6e17b5c4eaac44d821f42827a798b8c8d36f6ef005edb1ae7771f041R84)

As per the implementation in container family, if container has more than one flavors, then we need set version_in_uid to false. Only then multibuild content will be generated. Tome cat containers doesn't have this attribute set, do we need to set for tomcat containers or any changes needed in check version in uid function?

```

def check_version_in_uid(
        self,
        container: BaseContainerImage,
    ) -> bool:
        """check if version_in_uid is set to False if a container has more than one flavour"""
        if len(self.get_all_build_flavors(container)) > 1 and getattr(
            container, "version_in_uid", False
        ):
            return False
        return True
        
        
        ```

src/bci_build/package/__init__.py
rcmadhankumar: yes, there were some recursive import issues, i will try moving it out.

PR #2202, opened 2025-01-15
Add version check for git container as well

dirkmueller:
Without that we're publishing containers with the wrong version tags.

### DIFF
--- a/src/bci_build/package/git.py
+++ b/src/bci_build/package/git.py
@@ -1,16 +1,13 @@
 """Container description for the Git application container."""
 
-from bci_build.container_attributes import BuildType
-from bci_build.container_attributes import PackageType
 from bci_build.container_attributes import SupportLevel
 from bci_build.os_version import ALL_NONBASE_OS_VERSIONS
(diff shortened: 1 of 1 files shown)


dcermak (changes requested):
This won't work:
```
image/config.sh: line 23: rpm: command not found
```

dirkmueller:
> This won't work:
> 
> ```
> image/config.sh: line 23: rpm: command not found
> ```

darn, I already ran into this twice. I think since we don't want to move away from the micro container I think we want to switch away from kiwi.

dcermak:

>> This won't work:
>> 
>> ```
>> image/config.sh: line 23: rpm: command not found
>> ```
>
> darn, I already ran into this twice. I think since we don't want to move away from the micro container I think we want to switch away from kiwi. 

The only alternative would be to use the output of `git --version`,
e.g. `git --version | awk '{print $3}'` or the equivalent with `cut`.

dcermak (approved):
and it saves us half a meg!

PR #2210, opened 2025-01-17
Add Kiosk containers from home:atgracey:wallboardos

dcermak:
https://jira.suse.com/browse/EDGE-271

### DIFF
--- a/src/bci_build/package/firefox.py
+++ b/src/bci_build/package/firefox.py
@@ -1,0 +1,52 @@
+"""Build description for the Firefox container for the containerized kiosk
+containers.
+
+"""
+
+from bci_build.os_version import ALL_NONBASE_OS_VERSIONS
+from bci_build.os_version import CAN_BE_LATEST_OS_VERSION
+from bci_build.package import ApplicationStackContainer
+from bci_build.package import ParseVersion
+from bci_build.package import Replacement
+from bci_build.package.kiosk import KIOSK_EXCLUSIVE_ARCH
+
+FIREFOX_CONTAINERS = [
+    ApplicationStackContainer(
+        name=("kiosk-firefox" if os_version.is_tumbleweed else "kiosk-firefox-esr"),
+        os_version=os_version,
+        is_latest=os_version in CAN_BE_LATEST_OS_VERSION,
+        exclusive_arch=KIOSK_EXCLUSIVE_ARCH,
+        version=(ff_ver_re := "%%ff_ver%%"),
+        version_in_uid=False,
+        pretty_name="Mozilla Firefox",
+        package_list=(
+            [
+                "MozillaFirefox",
+                # for fonts to actually display
+                "xorg-x11-fonts",
+            ]
+            + (
+                ["MozillaFirefox-branding-openSUSE"]
+                if os_version.is_tumbleweed
+                else [
+                    "MozillaFirefox-branding-SLE",
+                    # required by Firefox via /usr/bin/gconftool-2
+                    # to be fixed via FileProvides
+                    "gconf2",
+                ]
+            )
+        ),
+        replacements_via_service=[
+            Replacement(
+                ff_ver_re,
+                package_name="MozillaFirefox",
+                parse_version=ParseVersion.MINOR,
+            )
+        ],
+        cmd=["/bin/bash", "-c", "firefox --kiosk $URL"],
+        custom_end="""RUN useradd -m user -u 1000
+ENV DISPLAY=":0"
+""",
+    )
+    for os_version in ALL_NONBASE_OS_VERSIONS
+]
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -1469,6 +1469,7 @@ from .basecontainers import KERNEL_MODULE_CONTAINERS  # noqa: E402
 from .basecontainers import MICRO_CONTAINERS  # noqa: E402
 from .basecontainers import MINIMAL_CONTAINERS  # noqa: E402
 from .cosign import COSIGN_CONTAINERS  # noqa: E402
+from .firefox import FIREFOX_CONTAINERS  # noqa: E402
 from .gcc import GCC_CONTAINERS  # noqa: E402
 from .git import GIT_CONTAINERS  # noqa: E402
 from .golang import GOLANG_CONTAINERS  # noqa: E402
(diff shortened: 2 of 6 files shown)


src/bci_build/package/kiosk.py
```diff
@@ -0,0 +1,173 @@
...
+from bci_build.os_version import CAN_BE_LATEST_OS_VERSION
+from bci_build.package import ApplicationStackContainer
+from bci_build.package import DevelopmentContainer
+from bci_build.package import ParseVersion
+from bci_build.package import Replacement
+from bci_build.package.helpers import generate_package_version_check
+from bci_build.package.versions import get_pkg_version
+
+_X11_FILES = {
+    "preferences": """ShowThemesMenu=0
```
alexandrevicenzi, line 18: nit: consider this being a file also, as other are already

src/bci_build/package/kiosk.py
```diff
@@ -0,0 +1,173 @@
...
+COPY entrypoint.sh /usr/local/bin/entrypoint.sh
+RUN chmod +x /usr/local/bin/entrypoint.sh
+""",
+    )
+    for os_version in ALL_NONBASE_OS_VERSIONS
+]
+
+
+_PULSE_FILES = {
+    "client.conf": """autospawn = no
```
alexandrevicenzi, line 90: nit: consider this being a file also, as other are already

dcermak:
It sorta kinda works
image

src/bci_build/package/kiosk.py
```diff
@@ -0,0 +1,165 @@
...
+COPY system.pa /etc/pulse/
+""",
+    )
+    for os_version in ALL_NONBASE_OS_VERSIONS
+]
+
+
+FIREFOX_CONTAINERS = [
+    DevelopmentContainer(
+        name="firefox",
```
dirkmueller, line 128: can we name this `firefox-esr` perhaps (see https://jira.suse.com/browse/PED-11731 for context)

src/bci_build/package/kiosk.py
```diff
@@ -0,0 +1,165 @@
...
+        os_version=os_version,
+        tag_version=(ff_ver := get_pkg_version("MozillaFirefox", os_version)),
```
dirkmueller, line 130: maybe use `firefox_ver` ? would be a bit better to understand

src/bci_build/package/kiosk.py
```diff
@@ -0,0 +1,165 @@
...
+            )
+        ),
+        replacements_via_service=[
+            Replacement(
+                ff_ver_re,
+                package_name="MozillaFirefox",
+                parse_version=ParseVersion.MINOR,
+            )
+        ],
+        cmd=["/bin/bash", "-c", "firefox --kiosk $URL"],
```
dirkmueller, line 155: does this work without `$URL` being set? why do we need to pipe this through `bash -c` ?

src/bci_build/package/kiosk.py
```diff
@@ -0,0 +1,165 @@
...
+        custom_end=generate_package_version_check(
+            "MozillaFirefox", ff_ver, ParseVersion.MINOR
```
dirkmueller, line 157: given that we don't actually care about the firefox version, can we just skip this? maintaining the firefox version numbers in the package-versions.json in git will be such a pain..

src/bci_build/package/kiosk.py
```diff
@@ -0,0 +1,165 @@
...
+QuickSwitch=0
+""",
+    "entrypoint.sh": (
+        (_x11_dir := (_pkg_dir := Path(__file__).parent) / "x11") / "entrypoint.sh"
+    ).read_text(),
+    "xinitrc": (_x11_dir / "xinitrc").read_text(),
+    "xorg.conf": (_x11_dir / "xorg.conf").read_text(),
+}
+
+X11_CONTAINERS = [
```
dirkmueller, line 30: can you please split that into a `x11.py` ? given that we have files in x11/ subdir that would be more aligned and easier to maintain going forward imho. 

also, not trying to be too nitpicky, but you're packaging xorg not X11 (that would be from the times before you were born..). can we just call it XORG instead?

src/bci_build/package/kiosk.py
```diff
@@ -0,0 +1,165 @@
...
+]
+
+
+_PULSE_FILES = {
+    "client.conf": ((_pa_dir := _pkg_dir / "pulseaudio") / "client.conf").read_text(),
+    "daemon.conf": (_pa_dir / "daemon.conf").read_text(),
+    "system.pa": (_pa_dir / "system.pa").read_text(),
+}
+
+PULSEAUDIO_CONTAINERS = [
```
dirkmueller, line 95: split into a `pulseaudio.py` for the same reason like x11 (see below) ?

src/bci_build/package/kiosk.py
agracey: It might be worth including an entrypoint.sh that defaults $URL to something sane. I had played with this a couple weeks ago when I was wondering about attaching devtools remotely: https://build.opensuse.org/projects/home:atgracey:wallboardos/packages/firefox-dev-entrypoint/files/entrypoint.sh?expand=1

dirkmueller:
> It sorta kinda works

missing fonts?

src/bci_build/package/kiosk.py
dcermak: I've removed this handling altogether

src/bci_build/package/kiosk.py
dcermak: The version is no longer in git

src/bci_build/package/kiosk.py
dcermak: > can you please split that into a `x11.py` ? given that we have files in x11/ subdir that would be more aligned and easier to maintain going forward imho.

No, these three containers belong together imho

> 
> also, not trying to be too nitpicky, but you're packaging xorg not X11 (that would be from the times before you were born..). can we just call it XORG instead?

I have renamed it `xorg` instead of `x11`

dcermak:
> > It sorta kinda works
> 
> missing fonts?

Yes, I hope it's fixed now. It worked locally, but I have to test the images from OBS too

dcermak:
It works, but the VM doesn't play any sound atm
image

src/bci_build/package/kiosk.py
```diff
@@ -0,0 +1,172 @@
...
+COPY client.conf /etc/pulse/
+COPY system.pa /etc/pulse/
+""",
+    )
+    for os_version in ALL_NONBASE_OS_VERSIONS
+]
+
+
+FIREFOX_CONTAINERS = [
+    DevelopmentContainer(
```
dirkmueller, line 134: Why not an ApplicationStackContainer? one needs to derive from it to  be useful?

dcermak: good catch! fixed

src/bci_build/package/firefox.py
```diff
@@ -0,0 +1,52 @@
...
+            )
+        ),
+        replacements_via_service=[
+            Replacement(
+                ff_ver_re,
+                package_name="MozillaFirefox",
+                parse_version=ParseVersion.MINOR,
+            )
+        ],
+        cmd=["/bin/bash", "-c", "firefox --kiosk $URL"],
```
alexandrevicenzi, line 46: Documentation for this will be added in a later PR?

dirkmueller (approved):

PR #2254, opened 2025-02-11
Switch kiwi version tracking to minor versions only

dirkmueller:
We don't actually conditionalize anything on the patch version, so we can use a service for that part and avoid source-changes for this.


dcermak (approved):
LGTM, ignore the SLFO build failure, the buildroot is broken, again :unamused:

PR #2259, opened 2025-02-11
cleanup rmt- prefix handling for mariadb and mariadb-client

dirkmueller:
The helm chart for RMT has moved on to not use rmt- prefix for quite a while, so we don't need to provide them as additional tags. Also move this into the package name only as otherwise it is incorrectly specified in the READMEs.

### DIFF
--- a/src/bci_build/package/mariadb.py
+++ b/src/bci_build/package/mariadb.py
@@ -9,7 +9,6 @@ from bci_build.container_attributes import BuildType
 from bci_build.container_attributes import SupportLevel
 from bci_build.os_version import ALL_NONBASE_OS_VERSIONS
 from bci_build.os_version import CAN_BE_LATEST_OS_VERSION
-from bci_build.os_version import OsVersion
 from bci_build.package import DOCKERFILE_RUN
 from bci_build.package import ApplicationStackContainer
 from bci_build.package import ParseVersion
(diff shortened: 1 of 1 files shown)


dcermak:
This changes the `BuildName`: https://github.com/SUSE/BCI-dockerfile-generator/commit/2a19facba40575657b1fe269dfa36f3e7c29f75f#diff-9840d91195094a2e16e136c109056d2dc91d9fed7525d1d266b1216ebcc344edR22 
Is that a problem?

dirkmueller:
> This changes the `BuildName`: [2a19fac#diff-9840d91195094a2e16e136c109056d2dc91d9fed7525d1d266b1216ebcc344edR22](https://github.com/SUSE/BCI-dockerfile-generator/commit/2a19facba40575657b1fe269dfa36f3e7c29f75f#diff-9840d91195094a2e16e136c109056d2dc91d9fed7525d1d266b1216ebcc344edR22) Is that a problem?

we'll have to update a few things in other places but I don't expect it to be an issue that would block publishing of new containers.

dcermak (approved):

dcermak:
> > This changes the `BuildName`: [2a19fac#diff-9840d91195094a2e16e136c109056d2dc91d9fed7525d1d266b1216ebcc344edR22](https://github.com/SUSE/BCI-dockerfile-generator/commit/2a19facba40575657b1fe269dfa36f3e7c29f75f#diff-9840d91195094a2e16e136c109056d2dc91d9fed7525d1d266b1216ebcc344edR22) Is that a problem?
> 
> we'll have to update a few things in other places but I don't expect it to be an issue that would block publishing of new containers.

Let's hope that's the case :crossed_fingers:

msmeissn:
i had  the aggregate file updated for the mariadb-image and mariadb-client-image

PR #2260, opened 2025-02-11
Set a stable tag_version for rmt-server

dirkmueller:
rmt-server is updating its minor version on every new release, which means by default that gets added to the buildname which means aggregates and config needs to be changed on every rmt-server release. add the major version to avoid that going forward.


dirkmueller:
Alternative here would be to use `stability_tag=stable` which might be an option but is also a bit weird.

dcermak (approved):

PR #2283, opened 2025-02-17
Change update-alternatives in ruby 3.4 container to match

dirkmueller:


src/bci_build/package/ruby.py
```diff
@@ -58,7 +58,18 @@ def _get_ruby_kwargs(ruby_version: Literal["2.5", "3.4"], os_version: OsVersion)
         },
         # as we only ship one ruby version, we want to make sure that binaries belonging
         # to our gems get installed as `bin` and not as `bin.ruby$ruby_version`
-        "config_sh_script": "sed -i 's/--format-executable/--no-format-executable/' /etc/gemrc",
+        "config_sh_script": (
+            (
+                """update-alternatives --set rake /usr/bin/rake.ruby.ruby3.4; \\
+    update-alternatives --set rdoc /usr/bin/rdoc.ruby.ruby3.4; \\
+    update-alternatives --set ri /usr/bin/ri.ruby.ruby3.4; \\
+    rm -v /usr/bin/ruby && ln -s ruby.ruby3.4 /usr/bin/ruby; \\
```
dcermak, line 66: I thought we're not supposed to mess in `/usr/bin/`?

src/bci_build/package/ruby.py
```diff
@@ -58,7 +58,18 @@ def _get_ruby_kwargs(ruby_version: Literal["2.5", "3.4"], os_version: OsVersion)
         },
         # as we only ship one ruby version, we want to make sure that binaries belonging
         # to our gems get installed as `bin` and not as `bin.ruby$ruby_version`
```
dcermak, line 60: These comments belong to the now last line of the `config_sh_script`

dcermak (commented):
Can you please explain why this is necessary in the code or the commit message? (also your current commit message has been cut off)

src/bci_build/package/ruby.py
dirkmueller: Right, I intended to file a bugreport about the underlying issue. The rm -v is to make sure it fails when the issue is fixed.

But I can change this to /usr/local and add a test case in BCI-tests instead.

dirkmueller:
> Can you please explain why this is necessary in the code or the commit message? (also your current commit message has been cut off)

Fair, will do.

dcermak (approved):

PR #2302, opened 2025-02-20
mark more images as singleton image

dirkmueller:


dirkmueller:
> We have two different kiwi major versions between sle15 and sle16, would this work for that case?

Yes. we only need versions in the build name if the container is maintained in parallel from the same codestream, which isn't the case here. 

> Others might be single images, although I am not hundred percent sure spack will stay this way.

There is no harm with changing it to not be a singleton image anymore once that is the case. we'll have most pain with renaming the source package container (and the loss of build counter etc).

dcermak (approved):

PR #2357, opened 2025-03-17
Remove postfix and Tomcat from the AppCollection

dirkmueller:
The appcollection team decided to package them by themselves, these containers are hence unused and we don't need to be maintaining them anymore.


alexandrevicenzi (approved):

dcermak:
Why are we removing tomcat completely but only moving postfix to registry.suse.com?

dirkmueller:
> Why are we removing tomcat completely but only moving postfix to registry.suse.com?

the other images have never been delivered to appcol in the first place. they're gone now as well. we are not allowed to ship tomcat or any of these containers on registry.suse.com anymore given that it is free to access.

PR #2364, opened 2025-03-24
Add bind9 container image for SUMA/MLM

dcermak:
https://jira.suse.com/browse/PED-12382

### DIFF
--- a/src/bci_build/package/bind.py
+++ b/src/bci_build/package/bind.py
@@ -1,0 +1,93 @@
+"""Build recipe for the ISC Bind 9 container image."""
+
+from pathlib import Path
+
+from bci_build.container_attributes import NetworkPort
+from bci_build.container_attributes import NetworkProtocol
+from bci_build.os_version import ALL_NONBASE_OS_VERSIONS
+from bci_build.os_version import CAN_BE_LATEST_OS_VERSION
+from bci_build.package import DOCKERFILE_RUN
+from bci_build.package import ApplicationStackContainer
+from bci_build.package import ParseVersion
+from bci_build.package import Replacement
+
+_BIND_FILES = {
+    "entrypoint.sh": (
+        (_bind_dir := Path(__file__).parent / "bind") / "entrypoint.sh"
+    ).read_text(),
+    "healthcheck.sh": (_bind_dir / "healthcheck.sh").read_text(),
+}
+
+BIND_CONTAINERS = [
+    ApplicationStackContainer(
+        name="bind",
+        os_version=os_version,
+        is_latest=os_version in CAN_BE_LATEST_OS_VERSION,
+        version=9,
+        pretty_name="ISC BIND 9",
+        version_in_uid=False,
+        package_list=["bind"],
+        exposes_ports=[
+            NetworkPort(53),
+            NetworkPort(53, NetworkProtocol.UDP),
+            NetworkPort(953),
+            NetworkPort(853),
+            NetworkPort(443),
+        ],
+        additional_versions=[
+            (_bind_minor_re := "%%bind_major_minor%%"),
+            (_bind_patch_re := "%%bind_major_minor_patch%%"),
+        ],
+        replacements_via_service=[
+            Replacement(_bind_minor_re, "bind", parse_version=ParseVersion.MINOR),
+            Replacement(_bind_patch_re, "bind", parse_version=ParseVersion.PATCH),
+        ],
+        extra_files=_BIND_FILES,
+        env={
+            # copy-pasta from /etc/sysconfig/named
+            "RNDC_KEYSIZE": 512,
+            "NAMED_ARGS": "",
+            "NAMED_INITIALIZE_SCRIPTS": "",
+            # need to set this one so that we can override it
+            "NAMED_CONF": "/etc/named.conf",
+        },
+        custom_end=rf"""COPY entrypoint.sh {(_entrypoint := "/usr/local/bin/entrypoint.sh")}
+COPY healthcheck.sh {(_healthcheck := "/usr/local/bin/healthcheck.sh")}
+{DOCKERFILE_RUN} \
+    chmod +x {_entrypoint}; \
+    chmod +x {_healthcheck};
+
+# patch named.prep to not call logger (provided by systemd)
+# and just log to stdout
+{DOCKERFILE_RUN} \
+    mkdir -p /usr/local/lib/bind; \
+    cp {os_version.libexecdir}bind/named.prep {(_named_prep := "/usr/local/lib/bind/named.prep")}; \
+    sed -i -e 's|logger "Warning: \$1"|echo "Warning: \$1" >\&2|' -e '/\. \$SYSCONFIG_FILE/d' {_named_prep}
+
+# create directories that tmpfiles.d would create for us
+{DOCKERFILE_RUN} \
+"""
+        + (r" \ " + "\n").join(
+            (
+                f"    mkdir -p {dirname}; chown {user} {dirname}; chmod {mode} {dirname};"
+                for dirname, mode, user in (
+                    ("/run/named", "1775", "root:named"),
+                    ("/var/lib/named", "1775", "root:named"),
+                    ("/var/lib/named/dyn", "755", "named:named"),
+                    ("/var/lib/named/master", "755", "named:named"),
+                    ("/var/lib/named/slave", "755", "named:named"),
+                    ("/var/log/named", "750", "named:named"),
+                )
+            )
+        )
+        + f"""
+# create files that tmpfiles.d would create for us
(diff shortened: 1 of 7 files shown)


src/bci_build/package/bind/README.md.j2
```diff
@@ -0,0 +1,78 @@
+# {{ image.title }}: BIND9 Application Container
+{% include 'badges.j2' %}
+
+## Description
+
+BIND (Berkeley Internet Name Domain) is a suite of software used for interacting
```
dmpop, line 6:
```suggestion
BIND (Berkeley Internet Name Domain) is a suite of software for interacting
```

src/bci_build/package/bind/README.md.j2
```diff
@@ -0,0 +1,78 @@
...
+with the Domain Name System (DNS). Its core component, `named`, serves as both
+an authoritative name server for DNS zones and a recursive resolver for network
+queries.
+
+## Usage
+
+This container image provides the `named` daemon from
+{{ image.os_version.full_os_name }}, including the default configuration shipped
+with the RPM package.
```
dmpop, line 15:
```suggestion
{{ image.os_version.full_os_name }}, including the default configuration that comes
with the RPM package.
```

src/bci_build/package/bind/README.md.j2
```diff
@@ -0,0 +1,78 @@
...
+
+By default, the container entrypoint launches `named` in the foreground mode
+when no additional parameters are specified:
```
dmpop, line 18:
```suggestion
When no additional parameters are specified, the container entrypoint launches `named` in the foreground mode by default:
```

src/bci_build/package/bind/README.md.j2
```diff
@@ -0,0 +1,78 @@
...
+
+### Health Check
+
+The container includes a health check script that performs a simple A-record
+lookup and verifies that a valid IPv4 address is returned.
+
+### Environment Variables
+
+The container entrypoint respects the following environment variables:
+
```
dmpop, line 32:
```suggestion
The container entrypoint accepts the following environment variables:

```

src/bci_build/package/bind/README.md.j2
```diff
@@ -0,0 +1,78 @@
...
+- **`ETC_RNDC_KEY`** (default: `/etc/rndc.key`): Location of the `rndc`
+    configuration file. If not a symlink, it is moved to `NEW_RNDC_KEY`.
+
+- **`NEW_RNDC_KEY`** (default: `/var/lib/named/rndc.key`): Target location where
+    `ETC_RNDC_KEY` is moved if it is not a symlink. If `NEW_RNDC_KEY` does not
+    exist, it is generated using `RNDC_BIN`.
+
+- **`RNDC_BIN`** (default: `/usr/sbin/rndc`): Binary used to generate the `rndc`
+    configuration file if it is missing.
```
dmpop, line 52:
```suggestion
    configuration file if it does not exist.
```

src/bci_build/package/bind/README.md.j2
```diff
@@ -0,0 +1,78 @@
...
+
+- **`RNDC_KEYSIZE`** (default: `512`): Key size passed to `RNDC_BIN` for
+    generating the `rndc` configuration file.
+
+
+### Differences from the RPM Package
```
dmpop, line 58:
```suggestion
### Differences compared to the RPM Package
```

src/bci_build/package/bind/README.md.j2
```diff
@@ -0,0 +1,78 @@
...
+
+Unlike the distribution package, this container does not use `systemd` to manage
+`named`. Instead, `named` is launched directly by the container entrypoint in
+foreground mode, with logs output to `stdout`.
```
dmpop, line 62:
```suggestion
The container does not use `systemd` to manage
`named`. Instead, `named` is launched directly by the container entrypoint in
foreground mode, with logs output to `stdout`.
```

src/bci_build/package/bind/README.md.j2
```diff
@@ -0,0 +1,78 @@
...
+
+Additionally, environment variables from `/etc/sysconfig/named` are not sourced
```
dmpop, line 64:
```suggestion
Moreover, environment variables from `/etc/sysconfig/named` are not sourced
```

src/bci_build/package/bind/README.md.j2
```diff
@@ -0,0 +1,78 @@
...
+or used. Instead, you should set them explicitly using the container runtime,
+for example:
```
dmpop, line 66:
```suggestion
or used. You must set them explicitly using the container runtime,
for example:
```

src/bci_build/package/bind.py
```diff
@@ -0,0 +1,84 @@
+from pathlib import Path
```
dirkmueller, line 1: please add a module docstring at the top

src/bci_build/package/bind.py
```diff
@@ -0,0 +1,84 @@
...
+            # need to set this one so that we can override it
+            "NAMED_CONF": "/etc/named.conf",
+        },
+        custom_end=rf"""COPY entrypoint.sh {(_entrypoint := "/usr/local/bin/entrypoint.sh")}
+COPY healthcheck.sh {(_healthcheck := "/usr/local/bin/healthcheck.sh")}
+{DOCKERFILE_RUN} \
+    chmod +x {_entrypoint}; \
+    chmod +x {_healthcheck};
+
+EXPOSE 53/udp 53/tcp 953/tcp 853/tcp 443/tcp
```
dirkmueller, line 49: didn't we have some exposes_port setting through the template generator ? would prefer that over having it hidden in custom_end as it is less easy to grep.

src/bci_build/package/bind.py
```diff
@@ -0,0 +1,84 @@
...
+
+# patch named.prep to not call logger (provided by systemd)
+# and just log to stdout
+{DOCKERFILE_RUN} \
+    mkdir -p /usr/local/lib/bind; \
+    cp {os_version.libexecdir}bind/named.prep {(_named_prep := "/usr/local/lib/bind/named.prep")}; \
+    sed -i -e 's|logger "Warning: \$1"|echo "Warning: \$1" >\&2|' -e '/\. \$SYSCONFIG_FILE/d' {_named_prep}
+
+# create directories that tmpfiles.d would create for us
```
dirkmueller, line 58: I wonder if there's a way for us to run the tmpfiles.d stuff without having to duplicate it in the dockerfile? this might risk a maintenance burden / incompatibility that we'll eventually miss years down the line.

src/bci_build/package/bind.py
```diff
@@ -0,0 +1,84 @@
...
+    chmod +x {_healthcheck};
+
+EXPOSE 53/udp 53/tcp 953/tcp 853/tcp 443/tcp
+
+# patch named.prep to not call logger (provided by systemd)
+# and just log to stdout
+{DOCKERFILE_RUN} \
+    mkdir -p /usr/local/lib/bind; \
+    cp {os_version.libexecdir}bind/named.prep {(_named_prep := "/usr/local/lib/bind/named.prep")}; \
+    sed -i -e 's|logger "Warning: \$1"|echo "Warning: \$1" >\&2|' -e '/\. \$SYSCONFIG_FILE/d' {_named_prep}
```
dirkmueller, line 56: is there a  request anywhere to fix this in the package, e.g. by reacting to a specific environment variable that we're setting?

dirkmueller (approved):
some nits stuff inline. generally okay with it, can be fixed in a followup as well

src/bci_build/package/bind.py
dcermak: Totally forgot that we had this :facepalm:

src/bci_build/package/bind.py
dcermak: No, I don't think that this is a bug in the package per-se. The package expects to be launched via systemd and environment variables set via the SYSCONFIG_FILE.
However, setting environment variables should be done by the container runtime and thus we cannot source the sysconfig_file as it would take precedence. And we cannot use the `logger` binary, as that is provided by systemd (and fortunately bind does not depend on systemd).

src/bci_build/package/bind.py
dcermak: Well, we can install systemd into the container, run systemd-tmpfiles and then remove systemd afterwards, but that sadly leaves a bunch of crap (mostly empty directories and pam backup files, but those are only around 32kb, so I guess we could live with that).

Nevertheless, I have specifically added tests to https://github.com/SUSE/BCI-tests/pull/771 which verify that the systemd-tmpfiles settings have been applied correctly

alexandrevicenzi (approved):

PR #2389, opened 2025-03-31
Limit concurrency in package-versions

dirkmueller:
Performance is irrelevant here, and the DoS protections of the build service are constantly blocking us. The code probably shouldn't be non-blocking to begin with.


dcermak (commented):
Or we just stop creating tasks and gathering them all in line 203?

dirkmueller:
I don't really understand why the BCI-dockerfile-generator uses nonblocking I/O everywhere, it is provably slower than simply using plain blocking calls, but we had this conversation already two times, so I'm not reopening it again. 

I think my use of the semaphore is not correct as it is binding it too early. will change.

dcermak (approved):

PR #2394, opened 2025-04-01
Ensure that os_version has a stable value in the slm60 container

dcermak:
The value of `os_version` comes from the previous loop and its value is not stable, leading to fun stuff like:
https://github.com/SUSE/BCI-dockerfile-generator/pull/2393 

Defining it using the walrus operator ensures that it has the correct value

### DIFF
--- a/src/bci_build/package/basecontainers.py
+++ b/src/bci_build/package/basecontainers.py
@@ -355,7 +355,7 @@ KERNEL_MODULE_CONTAINERS.append(
         name="slm60-kernel-module-devel",
         pretty_name="SUSE Linux Micro 6.0 Kernel module development",
         logo_url="https://opensource.suse.com/bci/SLE_BCI_logomark_green.svg",
-        os_version=OsVersion.SLE16_0,
+        os_version=(os_version := OsVersion.SLE16_0),
         supported_until=_SUPPORTED_UNTIL_SLE.get(os_version),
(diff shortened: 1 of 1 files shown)


src/bci_build/package/basecontainers.py
```diff
@@ -355,7 +355,7 @@ def _get_minimal_kwargs(os_version: OsVersion):
         name="slm60-kernel-module-devel",
         pretty_name="SUSE Linux Micro 6.0 Kernel module development",
         logo_url="https://opensource.suse.com/bci/SLE_BCI_logomark_green.svg",
-        os_version=OsVersion.SLE16_0,
+        os_version=(os_version := OsVersion.SLE16_0),
         supported_until=_SUPPORTED_UNTIL_SLE.get(os_version),
```
dirkmueller, line 359: so this line is wrong independent of the os_version instability, as the lifecycle of the container does not depend on SLE.

dirkmueller: filed https://github.com/SUSE/BCI-dockerfile-generator/pull/2396 to fix this.

dirkmueller:
with #2396 this change would no longer be necessary..

dcermak:
> with #2396 this change would no longer be necessary..

Hardly, we're still relying on random/unexpected behavior

PR #2405, opened 2025-04-08
Reduce size of minimal image with jdupes

dirkmueller:


dcermak (changes requested):
This had the opposite effect:
```
❯ podman inspect -f '{{.VirtualSize}}' registry.opensuse.org/home/defolos/bci/staging/sle-15-sp7/7-2405/images/bci/bci-minimal:15.7
49465390
❯ podman inspect -f '{{.VirtualSize}}' registry.opensuse.org/devel/bci/sle-15-sp7/images/bci/bci-minimal:15.7
49460690
```

dcermak:
Yeah, this will never work. The problem is that we are using the bci-micro image as the base which **already** includes the license files. The jdupes call happens in the layer on top, which cannot modify the base layer, it only adds additional symlinks. This increases the overall size and doesn't decrease it.

The only way would be to squash the image.

dirkmueller:
> ❯ podman inspect -f '{{.VirtualSize}}' registry.opensuse.org/home/defolos/bci/staging/sle-15-sp7/7-2405/images/bci/bci-minimal:15.7
> 49465390

```
podman inspect -f '{{.VirtualSize}}' registry.opensuse.org/home/defolos/bci/staging/sle-15-sp7/7-2405/images/bci/bci-minimal:15.7
46297482
```

LGTM actually. we're saving 2.5 MB.

dcermak (approved):

PR #2408, opened 2025-04-08
Switch to GCC 15

dirkmueller:


dcermak (commented):
Shouldn't `_is_latest_gcc` be adjusted for Tumbleweed and gcc 15?

dcermak (approved):

PR #2422, opened 2025-04-11
Add entrypoint.sh for pulseaudio container

alexandrevicenzi:
Taken from https://github.com/agracey/kiosk/pull/11.


src/bci_build/package/pulseaudio/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
...
+    echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')] $@"
+}
+
+trap 'log "Received SIGTERM, exiting PulseAudio"; kill -TERM $PA_PID; wait $PA_PID; exit 0' SIGTERM
+trap 'log "Received SIGINT, exiting PulseAudio"; kill -TERM $PA_PID; wait $PA_PID; exit 0' SIGINT
+
+chown root:audio /dev/snd/*
+
+# Start pulseaudio in foreground and capture its PID
+/usr/bin/pulseaudio -vvv --log-target=stderr &
```
dcermak, line 13: please exec pulseaudio, this causes pulseaudio to become a child of bash, which is bad. Bash doesn't forward signals correctly and doesn't reap processes, as it is not meant to be PID1

HVSharma12: Adding traps helps with containers shutdown, but I think Dan’s point is more about best practice using exec so the main process runs as PID 1 and avoid bash to handle sigterm with traps. I remember facing some issue with exec previously with x11. Ill check further to see whats actually causing the SIGKILL, and test with exec so we dont rely on traps.

dirkmueller:
other PR was merged instead

PR #2439, opened 2025-04-25
Remove gcc 12 container

dirkmueller:
Has been dropped from Tumbleweed as it is now unused and superseded by gcc 14/15


dirkmueller:
merging as it is hopefully obvious enough to bypass peer review

dcermak (approved):

PR #2444, opened 2025-04-25
Update poetry.lock

dirkmueller:
it has an aiohttp update which doesn't seem to help but maybe doesn't hurt.


dcermak (approved):
Are you dependabot now?

dirkmueller:
> Are you dependabot now?

dependabot is broken since the poetry v2 update, which is unsupported. see https://docs.github.com/de/code-security/dependabot/working-with-dependabot/dependabot-options-reference#package-ecosystem-

dcermak:

> dirkmueller left a comment (SUSE/BCI-dockerfile-generator#2444)
>
>> Are you dependabot now?
>
> dependabot is broken since the poetry v2 update, which is unsupported. see https://docs.github.com/de/code-security/dependabot/working-with-dependabot/dependabot-options-reference#package-ecosystem-

So instead of telling anyone and looking for a solution, you run updates
manually from now on?

dirkmueller:
> So instead of telling anyone and looking for a solution, you run updates manually from now on?

I'm looking for a solution for my actual issue, namely update-versions not running thanks to aiohttp weirdness (and https://github.com/SUSE/BCI-dockerfile-generator/pull/1482 not being accepted). I'm not interested in running or debugging dependabot - but I wanted to try newer aiohttp version in the hope that it fixes the update-versions not running anymore.

PR #2459, opened 2025-04-28
[bot] Return early when child == ancestor

dcermak:
The `_get_commit_range_between_refs` function, specifically `_recurse_search_for_ancestor` does not terminate if child == ancestor, so we abort early if both commits are the same

### DIFF
--- a/src/staging/git_repo.py
+++ b/src/staging/git_repo.py
@@ -1,0 +1,45 @@
+"""Module containing helper functions interact with git repositories."""
+
+import git
+
+
+def get_commit_range_between_refs(
+    child_ref: str, ancestor_ref: str, repo_path: str = "."
+) -> set[git.Commit] | None:
+    """Returns all commits leading from ``child_ref`` to ``ancestor_ref``,
+    **excluding** ``ancestor_ref`` in the repository specified via `repo_path`.
+
+    If ``ancestor_ref`` is not an ancestor of ``child_ref``, then ``None`` is
+    returned. If ``ancestor_ref`` and ``child_ref`` are the same commit (can be
+    denoted via different names), then an empty set is returned.
+
+    """
+    repo = git.Repo(repo_path)
+
+    def is_ancestor(child_sha: str, ancestor_sha: str) -> bool:
+        """Checks whether the commit identified via ``ancestor_sha`` is an
+        ancestor of the commit identified via ``child_sha``.
+
+        """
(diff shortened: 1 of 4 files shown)


dirkmueller:
`get_packages_without_changelog_addition` is doing this:

```python
        commit_range = self._get_commit_range_between_refs(change_ref, base_ref)
        if not commit_range:
            raise RuntimeError(f"{base_ref} is not an ancestor of {change_ref}!")
```

that needs potentially adjustment to check if it is `None` and then we should return an empty set rather than None?

src/staging/git_repo.py
```diff
@@ -0,0 +1,48 @@
+import git
```
dirkmueller, line 1: module docstring

src/staging/git_repo.py
```diff
@@ -0,0 +1,48 @@
...
+
+    If ``ancestor_ref`` is not an ancestor of ``child_ref``, then ``None`` is
+    returned. If ``ancestor_ref`` and ``child_ref`` are the same commit (can be
+    denoted via different names), then an empty set is returned.
+
+    """
+    repo = git.Repo(repo_path)
+
+    def is_ancestor(child_sha: str, ancestor_sha: str) -> bool:
+        """Checks whether ancestor_sha is a"""
```
dirkmueller, line 18: intentionally unfinished sentence?

tests/test_git_repo.py
```diff
@@ -0,0 +1,88 @@
...
+        # * |   37ebaf51 Merge pull request #2435 from HVSharma12/firefox-audio-fix
+        #
+        #
+        # $ git rev-list e57be2ad ^86990f63
+        # e57be2ad998a6b456a405ad967c9707bd350d0e0
+        # 9ede9a8745d557ad918beda987df8e5acd18e9c2
+        # 9ed4c2d9f4e4a09f4f955d3e18ba3c5bc8d525c8
+        # fe656acb9863a900f47257f9f9a240c399854477
+        #
+        # but fe656acb9863a900f47257f9f9a240c399854477 is not a child of 86990f63…
```
dirkmueller, line 57: but it was merged as part of `9ede9a` for the first time, so imho it should be included?

dcermak: forgot to git push :facepalm:

dirkmueller (approved):

PR #2461, opened 2025-04-29
Pre-create pcpqa & systemd-coredump users with fixed uids

dcermak:
These users get created by systemd-sysusers on container start and their uid is not fixed. To prevent that, we pre-create these users

### DIFF
--- a/src/bci_build/package/appcontainers.py
+++ b/src/bci_build/package/appcontainers.py
@@ -86,6 +86,12 @@ COPY pmcd pmlogger /etc/sysconfig/
 # This can be removed after the pcp dependency on sysconfig is removed
 {DOCKERFILE_RUN} systemctl disable wicked wickedd || :
 
+{DOCKERFILE_RUN} useradd --comment "PCP Quality Assurance" \
+    --create-home --home-dir /var/lib/pcp/testsuite \
+    --shell /bin/bash \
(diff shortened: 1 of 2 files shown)


dirkmueller:
@dcermak is this ready for review?

alexandrevicenzi (approved):

src/bci_build/package/appcontainers.py
```diff
@@ -86,6 +86,8 @@ def _envsubst_pkg_name(os_version: OsVersion) -> str:
 # This can be removed after the pcp dependency on sysconfig is removed
 {DOCKERFILE_RUN} systemctl disable wicked wickedd || :
 
+{DOCKERFILE_RUN} useradd --comment "PCP Quality Assurance" --create-home --home-dir /var/lib/pcp/testsuite --shell /bin/bash --uid {"483" if os_version.is_sle15 else "496"} pcpqa
```
dirkmueller, line 89: this user is only used for testing it seems. so can we just always use 496 rather than having a sle15 specific uid?

dirkmueller: we only need to keep it stable for stuff that is overwritten via configuration or volume mounts, which doesn't seem to be the case here?

dirkmueller (approved):

PR #2463, opened 2025-04-29
Mark the kiosk containers as singleton images

dirkmueller:
Also set "esr" as a rolling tag rather than a package suffix.

### DIFF
--- a/src/bci_build/package/firefox.py
+++ b/src/bci_build/package/firefox.py
@@ -17,6 +17,8 @@ FIREFOX_CONTAINERS = [
         is_latest=os_version in CAN_BE_LATEST_OS_VERSION,
         exclusive_arch=KIOSK_EXCLUSIVE_ARCH,
         version=(ff_ver_re := "%%ff_ver%%"),
+        tag_version=None if os_version.is_tumbleweed else "esr",
+        is_singleton_image=True,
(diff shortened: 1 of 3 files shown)


src/bci_build/package/firefox.py
```diff
@@ -12,11 +12,13 @@
 
 FIREFOX_CONTAINERS = [
     ApplicationStackContainer(
-        name=("kiosk-firefox" if os_version.is_tumbleweed else "kiosk-firefox-esr"),
+        name="kiosk-firefox",
```
dcermak, line 15: Why the change after https://github.com/SUSE/BCI-dockerfile-generator/pull/2210#discussion_r1951323930?

dcermak: > is there any plan to provide a non-esr version on sle15?

No, never. I think I added the `-esr` suffix also because on sle16, we'll call the firefox package `firefox-esr` to avoid confusion of customers who were surprised that `zypper in MozillaFirefox` gave them the ESR version and not the "ordinary" one

dirkmueller: anyhow, I dropped this part of the change, as we have more important things to do.

dcermak (approved):

PR #2500, opened 2025-05-05
Move Kiosk related containers into its own registry path

dirkmueller:
This is needed to implement access control

### DIFF
--- a/src/bci_build/package/firefox.py
+++ b/src/bci_build/package/firefox.py
@@ -9,10 +9,12 @@ from bci_build.package import ApplicationStackContainer
 from bci_build.package import ParseVersion
 from bci_build.package import Replacement
 from bci_build.package.kiosk import KIOSK_EXCLUSIVE_ARCH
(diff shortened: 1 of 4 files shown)


dirkmueller:
reference fails

alexandrevicenzi (approved):

src/bci_build/package/firefox.py
```diff
@@ -9,6 +9,7 @@
 from bci_build.package import ParseVersion
 from bci_build.package import Replacement
 from bci_build.package.kiosk import KIOSK_EXCLUSIVE_ARCH
+from bci_build.package.kiosk import KioskRegistry
 
 FIREFOX_CONTAINERS = [
     ApplicationStackContainer(
```
dcermak, line 15: I'd suggest to strip the `kiosk-` part from the name (at least for non-TW), as the registry url now ends with `kiosk/kiosk-firefox-esr`

PR #2501, opened 2025-05-05
restore executable permissions on bind entrypoint

dirkmueller:
For reasons beyond me this worked locally but not in the build service *shrug*.


dirkmueller:
tests pass again, should be obvious..

dcermak (approved):

PR #2514, opened 2025-05-06
Move projects under devel:BCI for consistency

alexandrevicenzi:


src/staging/project_setup.py
```diff
@@ -95,8 +95,8 @@ def generate_project_name(
 ) -> str:
     res = {
         ProjectType.DEVEL: "devel:BCI:",
-        ProjectType.CR: f"home:{osc_username}:BCI:CR:",
-        ProjectType.STAGING: f"home:{osc_username}:BCI:Staging:",
+        ProjectType.CR: f"devel:BCI:CR:",
+        ProjectType.STAGING: f"devel:BCI:Staging:",
```
dcermak, line 99: There's a catch with this change: you must be a maintainer of devel:BCI to run the bot locally. Previously, it would create a staging project in your home project, which everyone owns.

Then again, the home: scheduler is de-prioritized first.

dirkmueller: I never figured out how to run the bot locally. how would one do that?

dirkmueller: in general I think moving out of "home:defolos" is a good improvement. having staging below devel:BCI means it will inherit signing keys and permissions from the parent project, which might not be a good idea. 

I wonder if we can request a "devel:BCI-CI" and then move staging and CR there.

dirkmueller:
or to run a full build:

OSC_USER=defolos OSC_PASSWORD=(pass secret) poetry run scratch-build-bot \
                             --os-version 6 \
                             --branch-name="6-1184" \
                             -vvvv \
                             scratch_build \
                               --commit-message='Test build for #1184'


that always pushes the code to the origin. I would like it to be in my own personal fork. I guess I need to add an option for that.

dirkmueller:
> 
> that always pushes the code to the origin. I would like it to be in my own personal fork. I guess I need to add an option for that.

src/staging/project_setup.py
alexandrevicenzi: Can we have a subproject, like `devel:BCI:CI`, that does not inherit certain things? Or is that only possible with a new devel project?

dirkmueller: > Can we have a subproject, like `devel:BCI:CI`, that does not inherit certain things? Or is that only possible with a new devel project?

only possible with another project hierarchy

PR #2516, opened 2025-05-06
Replace py_obs with requests

dirkmueller:
Less code to maintain.

### DIFF
--- a/src/bci_build/package/versions.py
+++ b/src/bci_build/package/versions.py
@@ -64,17 +64,16 @@ results in:
 
 """
 
-import asyncio
 import json
+import xml.etree.ElementTree as ET
 from pathlib import Path
 from typing import Literal
 from typing import NoReturn
 from typing import TypedDict
 from typing import overload
 
+import requests
 from packaging import version
-from py_obs.osc import Osc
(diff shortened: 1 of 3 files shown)


src/bci_build/package/versions.py
```diff
@@ -167,7 +166,18 @@ def format_version(ver: str, format: ParseVersion) -> str:
...
-async def update_versions(osc: Osc) -> dict[str, dict[str, str]]:
+def fetch_package_version(prj: str, pkg: str) -> str:
+    """Ask the OBS API to parse the source spec file version and return it."""
+    fetch = requests.get(
+        f"https://api.opensuse.org/public/source/{prj}/{pkg}",
+        params={"view": "info", "parse": "1"},
+        headers={"User-Agent": "BCI update-versions"},
+    )
+    fetch.raise_for_status()
+    return getattr(ET.fromstring(fetch.text).find("version"), "text", None)
```
dcermak, line 177: This means the type hint needs adjustment or raise an exception if `None` is returned

dirkmueller: true, I made it raise an exception

alexandrevicenzi (approved):

PR #2522, opened 2025-05-07
Add procps to pulseaudio container

alexandrevicenzi:
This is to align with the Leap-based image, and also because QA requires it.


dcermak (approved):

src/bci_build/package/pulseaudio.py
```diff
@@ -35,7 +35,7 @@
         is_singleton_image=True,
         version=(pulseaudio_ver_re := "%%pulseaudio_ver%%"),
         pretty_name="Pulseaudio",
-        package_list=["pulseaudio", "pulseaudio-utils"],
+        package_list=["pulseaudio", "pulseaudio-utils", "procps"],
```
dirkmueller, line 38: potentially adding `common_devel_packages` would be good

PR #2525, opened 2025-05-07
Define supportlevel and enddate for containerized kiosk

dirkmueller:


dirkmueller:
@fcrozat any concerns about that?

dcermak (approved):

PR #2535, opened 2025-05-12
Properly cache the poetry venv and drop use of bci-ci container

dirkmueller:


dcermak:
Unit tests are failing

dirkmueller:
> Unit tests are failing

fixed

dcermak (commented):
I think this is a giant NiH and not a good idea: there's existing libraries to parse rpm-md repos and shipping our own will only bite us down the road. Also, this looks to me like a solution looking for a problem, the previous container worked well for years :shrug:

But since I won't be maintaining this, I won't block this.

dirkmueller:
> I think this is a giant NiH and not a good idea: there's existing libraries to parse rpm-md repos

which one? only found https://pypi.org/project/repomd/ which we could use (but hasn't seen a single commit in 6 years..)

> and shipping our own will only bite us down the road. 

the repomd format hasn't changed in the last 20 years, so I think maintenance effort is quite acceptable imho.

> Also, this looks to me like a solution looking for a problem, the previous container worked well for years 🤷

it worked great in the CI, but it is super annoying to get a dnf setup working on any openSUSE distribution, especially when using a non-default python version (leap 15, leap 16, tumbleweed).

dcermak:

> dirkmueller left a comment (SUSE/BCI-dockerfile-generator#2535)
>
>> I think this is a giant NiH and not a good idea: there's existing libraries to parse rpm-md repos
>
> which one? only found https://pypi.org/project/repomd/ which we could use (but hasn't seen a single commit in 6 years..)

libdnf

>> and shipping our own will only bite us down the road. 
>
> the repomd format hasn't changed in the last 20 years, so I think maintenance effort is quite acceptable imho.
>
>> Also, this looks to me like a solution looking for a problem, the previous container worked well for years 🤷
>
> it worked great in the CI, but it is super annoying to get a dnf setup working on any openSUSE distribution, especially when using a non-default python version (leap 15, leap 16, tumbleweed). 

Have you considered opening a bug?

dirkmueller:
> Dirk Mueller ***@***.***> writes:
 only found https://pypi.org/project/repomd/ which we could use (but hasn't seen a single commit in 6 years..)
> libdnf

not packaged on pypi, hence not qualified.

> Have you considered opening a bug?

I did not open a bug but talked to the maintainer, and there is no interest in moving to singlespec (as the specfile is shared with fedora which does not use singlespec).

dirkmueller:
don't have time to make this work atm, so moved to a intermediate solution here: https://github.com/SUSE/BCI-dockerfile-generator/pull/2697

PR #2539, opened 2025-05-13
Merge versioning behavior between Dockerfiles and Kiwi type builds

dirkmueller:
We had two versioning implementations, one for dockerfiles and one for kiwi builds. So it happened what happens whenever there are two implementations: they diverge in an incompatible way. So merge them into one implementation.

- `15.<SP>.0` for OsContainers (whatever type) on SLE15
- `16.<minor>.0` for OsContainers (whatever type) on SLE16
- `<YEAR>` for tumbleweed

Derived containers have "baseversion.derivedversion" (so removing the 3rd component of the OsContainers)

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -402,10 +402,41 @@ class BaseContainerImage(abc.ABC):
 
     @property
     def build_version(self) -> str | None:
-        if self.os_version.is_sle15:
+        """Define the BuildVersion that is used for determining which container build
+        is newer than the other in the build service."""
+        # KIWI used to require an at least 3 component version X.Y.Z. For SLE, we set
+        # X.Y to MAJOR.MINOR and set Z to 0 for OsContainers. Derived
+        # containers inhert the base build_version X.Y and append a suffix .Z.ZZ
+
+        # It is important that the behavior for OsContainers is identical
+        # between KIWI and Dockerfile builds so that we can switch between these
+        # types.
+
+        # TODO: also set it for non-kiwi type (historically we haven't done so)
+        if self.os_version.is_tumbleweed and self.build_recipe_type == BuildType.KIWI:
+            if isinstance(self, OsContainer):
+                return f"{str(datetime.datetime.now().year)}.0.0"
+            return f"{str(datetime.datetime.now().year)}.0"
+        elif self.os_version.is_sle15:
+            if isinstance(self, OsContainer):
+                return f"15.{int(self.os_version.value)}.0"
             return f"15.{int(self.os_version.value)}"
-        if self.os_version.is_sl16:
+        elif self.os_version.is_sl16:
+            if isinstance(self, OsContainer):
+                return f"{str(self.os_version.value)}.0"
             return str(self.os_version.value)
+        return None
+
+    @property
+    def kiwi_version(self) -> str | None:
+        """Return a BuildVersion that is compatible with the requirements that KIWI imposes.
+        https://osinside.github.io/kiwi/image_description/elements.html#preferences-version
+
+        It a version strictly in the format X.Y.Z.
+        """
+        build_ver: str | None = self.build_version
+        if build_ver:
+            return ".".join(build_ver.split(".")[:3])
         return None
 
     @property
@@ -1027,12 +1058,6 @@ exit 0
             + "."
             + (self.custom_labelprefix_end or self.name)
         )
-
-    @property
-    def kiwi_version(self) -> str:
-        if self.os_version in (OsVersion.TUMBLEWEED, OsVersion.SLE16_0):
-            return str(datetime.datetime.now().year)
-        return f"15.{int(self.os_version.value)}.0"
 
     @property
     def kiwi_additional_tags(self) -> str | None:
--- a/src/bci_build/package/basecontainers.py
+++ b/src/bci_build/package/basecontainers.py
@@ -54,7 +54,7 @@ MICRO_CONTAINERS = [
         from_target_image="scratch",
         cmd=["/bin/sh"],
         package_list=[pkg.name for pkg in _get_micro_package_list(os_version)],
-        _min_release_counter=35,  # be newer than the newest kiwi based image on SP6
+        _min_release_counter=41,  # be newer than the newest kiwi based image on SP6
         build_stage_custom_end=(
             (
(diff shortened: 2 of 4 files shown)


dcermak:
>     * for tumbleweed

Do we really want to start setting buildname & buildversion for Tumbleweed? This could cause needless problems, depending how the images are published

dirkmueller:
> Do we really want to start setting buildname & buildversion for Tumbleweed? This could cause needless problems, depending how the images are published

well, if we don't set one, the buidl service picks a "random" one which we do not control and where conversion between kiwi and dockerfile causes other problems. 

so the underlying issue is that kiwi requires a version to be set, and dockerfiles do not. by not setting it, we have to fight "the autoversion selection logic" vs kiwi set version. I think having the versions set will be easier as we have more control?

src/bci_build/package/__init__.py
```diff
@@ -402,12 +402,43 @@ def build_name(self) -> str | None:
...
+        # KIWI used to require an at least 3 component version X.Y.Z. For SLE, we set
+        # X.Y to MAJOR.MINOR and set Z to 0 for OsContainers. Derived
+        # containers inhert the base build_version X.Y and append a suffix .Z.ZZ
+
+        # It is important that the behavior for OsContainers is identical
+        # between KIWI and Dockerfile builds so that we can switch between these
+        # types.
+
+        if self.os_version.is_tumbleweed and isinstance(self, OsContainer):
+            return f"{str(datetime.datetime.now().year)}.0.0"
```
Vogtinator, line 416: Why just the year?

src/bci_build/package/__init__.py
```diff
@@ -402,12 +402,43 @@ def build_name(self) -> str | None:
...
+    @property
+    def kiwi_version(self) -> str | None:
+        """Return a BuildVersion that is compatible with the requirements that KIWI imposes.
+        https://osinside.github.io/kiwi/image_description/elements.html#preferences-version
+
+        It a version strictly in the format X.Y.Z.
+        """
+        build_ver: str | None = self.build_version
+        if build_ver:
+            return ".".join(build_ver.split(".")[:3])
```
Vogtinator, line 439: FWICT the `[:3]` will only truncate, not extend, so if build_ver is just X or X.Y, this won't add `.0` or so

src/bci_build/package/__init__.py
```diff
@@ -402,12 +402,43 @@ def build_name(self) -> str | None:
...
+
+        # It is important that the behavior for OsContainers is identical
+        # between KIWI and Dockerfile builds so that we can switch between these
+        # types.
+
+        if self.os_version.is_tumbleweed and isinstance(self, OsContainer):
+            return f"{str(datetime.datetime.now().year)}.0.0"
+        # TODO: also set it for non-kiwi type (historically we haven't done so)
+        if self.os_version.is_tumbleweed and self.build_recipe_type == BuildType.KIWI:
+            return f"{str(datetime.datetime.now().year)}.0"
```
Vogtinator, line 419: That's just 2025.0, without the .patch component. Intentional?

src/bci_build/package/__init__.py
```diff
@@ -402,12 +402,43 @@ def build_name(self) -> str | None:
...
+        if self.os_version.is_sle15 and isinstance(self, OsContainer):
```
Vogtinator, line 420: I don't quite get the `isinstance(self, OsContainer)` check here for the `.0` suffix

src/bci_build/package/__init__.py
dirkmueller: I'd be fien with Y.Month, but that means a monthy "all tumbleweed containers get resubmitted" change and in the past Dan was very allergic to mass submissions

src/bci_build/package/__init__.py
dirkmueller: yes, but build_version will be always 3 or 4 components (oscontainers or develcontainers). we can add tests for this.

src/bci_build/package/__init__.py
dirkmueller: see the comment above. osContainers have X.Y version numbers, derived containers have `X.Y.Z.ZZ` (`Z.ZZ` being derived from the containers tag_version (e.g. the application software version)

src/bci_build/package/__init__.py
dirkmueller: yes, it is combined with the application version of the derived contaienr, so this leads to `2025.0.<major>` of the derived container version. compatible with how it works for sle15/sle16 so we only have one case to worry about.

src/bci_build/package/__init__.py
Vogtinator: Then I'd argue that the year is also useless and it should just be something like an actual version number that's bumped on $random_occasion or 84.87 (like filesystem).

dirkmueller: I don't get the 84.87 joke of filesystem. using datestamps for version numbers is very popular in various packages in tumbleweed. personally I don#t really care what we come up with but it needs to be `>= 2025.0`

dirkmueller:
> The CI fails because this also added `#!BuildName: opensuse-bci-bci-micro-%OS_VERSION_ID_SP%` as a side effect which makes OBS treat the binaries as separate, leading to "have choice"

the "have choice" isn't supposed to be happening. it happens because the project setup is wrong for tumbleweed. we should not have non-standard projects in "standard". 

if it would be on the same "level" there wouldn#t be a have-choice. I can send another PR for this.

Vogtinator:
> the "have choice" isn't supposed to be happening. it happens because the project setup is wrong for tumbleweed. we should not have non-standard projects in "standard".

Unfortunately it will also affect everyone building against `openSUSE:Containers:Tumbleweed` in OBS as long as we have older binaries available there.

dirkmueller:
> Unfortunately it will also affect everyone building against `openSUSE:Containers:Tumbleweed` in OBS as long as we have older binaries available there.

alright, reverted that part as well.

alexandrevicenzi (approved):

src/bci_build/package/__init__.py
Vogtinator: Apparently trivy-image has `<version>2025.0</version>` now

Vogtinator (approved):
:popcorn: :boom:

PR #2541, opened 2025-05-14
Switch to libopenssl3 for the new dotnet releases

dirkmueller:


dcermak:
fails to build everywhere

dirkmueller:
> fails to build everywhere

for a totally different reason, see https://github.com/dotnet/runtime/issues/115554 . we still need this change.

dirkmueller:
> Sure, but the .Net 9 images were building before.

No, they are not building without this change, because they now require libopenssl3 and we still install libopenssl1_1. an older version of the 9 images were building (e.g. before yesterdays security update).

dirkmueller:
(apparently only scheduled for june release)

PR #2546, opened 2025-05-14
Switch to singleton basename for all 16.0 containers

dirkmueller:


src/bci_build/package/basecontainers.py
```diff
@@ -45,6 +45,10 @@ def _get_micro_package_list(os_version: OsVersion) -> list[Package]:
         supported_until=_SUPPORTED_UNTIL_SLE.get(os_version),
         logo_url="https://opensource.suse.com/bci/SLE_BCI_logomark_green.svg",
         is_latest=os_version in CAN_BE_LATEST_OS_VERSION,
+        is_singleton_image=(
+            # preserve backwards compatibility on already released distributions
+            os_version not in (OsVersion.SP6, OsVersion.SP7)
```
dcermak, line 50: Maybe a bit clearer?
```suggestion
            not os_version.is_sle15
```

dirkmueller: sure. done

PR #2563, opened 2025-05-15
Fix trivys tag_version and add a basic README

dirkmueller:


Vogtinator:
Should be removed from appcontainers.py then?

dirkmueller:
> Should be removed from appcontainers.py then?

yep, I forgot to git add that.

Vogtinator (approved):

PR #2570, opened 2025-05-15
ref the buildcounter after version reset

dirkmueller:
These dockerfile based base containers just had a buildversion change to align with how kiwi base containers are labelled to X.Y.Z style,which resets the buildcounter. ref it again.


Vogtinator (approved):
And so the whack-a-mole begins

PR #2577, opened 2025-05-16
Revert "Remove redundant permission command"

dirkmueller:
The comment was giving a hint. these files and permissions are not set by the rpm script because they come via a volume mount from helm (hence are not initialized from the container like with a podman mount)

Reverts SUSE/BCI-dockerfile-generator#2572


ngetahun (approved):
LGTM!

PR #2769, opened 2025-07-09
Include pgvector extension(jsc#PED-10507)

dirkmueller:


src/bci_build/package/postgres.py
```diff
@@ -41,7 +41,14 @@
             "coreutils",
             "sed",
             "util-linux",  # for setpriv :-(
-        ],
+        ]
+        + (
+            [
+                f"postgresql{ver}-pgvector",
+            ]
+            if os_version.is_tumbleweed
```
alexandrevicenzi, line 49: This does not fulfill the PED ticket, because the extension would need to be in SLE.

dirkmueller: @alexandrevicenzi I'm waiting for product management to decide whether the ticket is still needed because it has been worked on 13 months after the dead line of delivery. However, I think it might be a good idea to include it and develop it for opensuse. 

We can always improve it once product management had taken a decision.

PR #2780, opened 2025-07-15
Change pipx PATH to use standard locations (bsc#1246200)

alexandrevicenzi:


src/bci_build/package/python.py
```diff
@@ -74,8 +74,10 @@ def _get_python_kwargs(py3_ver: _PYTHON_VERSIONS, os_version: OsVersion):
         "env": {
             "PYTHON_VERSION": py3_ver_replacement,
             "PIP_VERSION": pip3_replacement,
-        }
-        | ({"PATH": "$PATH:/root/.local/bin"} if has_pipx else {}),
+            "PIPX_HOME": "/opt/pipx",
```
dirkmueller, line 77: this should only be set if it is a container with pipx. also `/opt/pipx` is nonstandard. please use either `/var/lib/pipx` or `/usr/local/lib/pipx` instead (I think the 2nd is preferred).

src/bci_build/package/python.py
```diff
@@ -103,7 +105,7 @@ def _get_python_kwargs(py3_ver: _PYTHON_VERSIONS, os_version: OsVersion):
 
     config_sh_script = ""
     if has_pipx:
-        config_sh_script += "install -d -o root -g root -m 0700 /root/.local/bin"
+        config_sh_script += "install -d -o root -g root -m 0700 /opt/pipx"
```
dirkmueller, line 108: that makes no sense anymore. the permissions would need to be updated.

src/bci_build/package/python.py
alexandrevicenzi: I took the path from pipx docs, but indeed, `/usr/local/lib/pipx` feels more correct.

src/bci_build/package/python.py
```diff
@@ -74,8 +74,10 @@ def _get_python_kwargs(py3_ver: _PYTHON_VERSIONS, os_version: OsVersion):
         "env": {
             "PYTHON_VERSION": py3_ver_replacement,
             "PIP_VERSION": pip3_replacement,
-        }
-        | ({"PATH": "$PATH:/root/.local/bin"} if has_pipx else {}),
+            "PIPX_HOME": "/opt/pipx",
+            "PIPX_BIN_DIR": "/usr/local/bin",
+            "PIPX_MAN_DIR": "/usr/local/man",
```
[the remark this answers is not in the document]
alexandrevicenzi: yes, in theory, this should work. I was even considering a test case for pipx to ensure this scenario.

src/bci_build/package/python.py
alexandrevicenzi: this entire line is not needed, pipx does create the home folder if it does not exist.

dirkmueller (approved):

PR #2808, opened 2025-07-28
Set permctl

dirkmueller:


dirkmueller:
unfortunately it isn't actually working:

```
[   63s] The command '/bin/sh -c set -euo pipefail;     export PERMCTL_ALLOW_INSECURE_MODE_IF_NO_PROC=1; zypper -n --installroot /target --gpg-auto-import-keys install --no-recommends MozillaFirefox MozillaFirefox-branding-SLE gconf2 libavcodec58_134 libpulse0 noto-sans-cjk-fonts xorg-x11-fonts' returned a non-zero code: 107
```

which means the export is either not respected or not arriving at the rpm(1) invocation that executes the scriptlets.

alexandrevicenzi (approved):

PR #2826, opened 2025-07-31
Add mokutil to the kernel-development containers

dirkmueller:
It makes sense to be included there to be able to load additional uefi signing keys for kernel modules.


alexandrevicenzi (commented):
It fails to build for s390, and I dont think we have a way to distinguish a package per arch, only per OS version.

dirkmueller:
> It fails to build for s390 because the package does not exists, and I dont think we have a way to distinguish a package per arch, only per OS version.

we have, but only via a custom_end, which we can use here because it is a base container.

alexandrevicenzi (approved):

PR #2836, opened 2025-08-06
Change user id and group id to 999 for postgres user

rcmadhankumar:


src/bci_build/package/postgres.py
```diff
@@ -82,7 +82,24 @@
         ],
         volumes=["$PGDATA"],
         exposes_ports=[TCP(5432)],
-        custom_end=rf"""COPY docker-entrypoint.sh /usr/local/bin/
+        custom_end=(
+            rf"""COPY docker-entrypoint.sh /usr/local/bin/
+{DOCKERFILE_RUN} groupmod -g 999 postgres; \
+    usermod -u 999 postgres; \
+    chown -R postgres:postgres /var/lib/pgsql; \
+    chmod +x /usr/local/bin/docker-entrypoint.sh; \
```
dirkmueller, line 90: I think what we should be doing is create a base container layer that includes those users, then you don't need to do `chown -R` and preserve the permissions set by installing the packages.

src/bci_build/package/postgres.py
```diff
@@ -82,7 +82,24 @@
...
+    sed -i -e 's/exec gosu postgres "/exec setpriv --reuid=postgres --regid=postgres --clear-groups -- "/g' /usr/local/bin/docker-entrypoint.sh; \
```
dirkmueller, line 91: this is unrelated isn't it?

src/bci_build/package/postgres.py
```diff
@@ -82,7 +82,24 @@
...
+    mkdir /docker-entrypoint-initdb.d; \
+    install -m 1775 -o postgres -g postgres -d /run/postgresql; \
+    install -d -m 0700 -o postgres -g postgres $PGDATA; \
+    sed -ri "s|^#?(listen_addresses)\s*=\s*\S+.*|\1 = '*'|" /usr/share/postgresql{ver}/postgresql.conf.sample
+
+STOPSIGNAL SIGINT
+HEALTHCHECK --interval=10s --start-period=10s --timeout=5s \
+    CMD pg_isready -U ${{POSTGRES_USER:-postgres}} -h localhost -p 5432
+"""
+            if ver == 16
```
alexandrevicenzi, line 101: `ver` is the psql version,  we should use `os_version` and apply this to SLE 16 only, to any version on SLE 16.

alexandrevicenzi: perhaps also apply to Tumbleweed, but ideally not to SLE 15.

rcmadhankumar:
looking into failures. out for review for the approach and steps.

src/bci_build/package/postgres.py
rcmadhankumar: approach changed.

src/bci_build/package/postgres.py
rcmadhankumar: The stable user and group ids will be supported only from postgres version 18, made changes accordingly.
Added postgres version 17 to the list just for testing right now.

src/bci_build/templates.py
```diff
@@ -53,7 +53,18 @@
...
-    zypper -n {%- if image.from_target_image %} --installroot /target {%- endif %} remove {{ image.packages_to_delete }}{%- endif %}{%- endif %}
+    zypper -n {%- if image.from_target_image %} --installroot /target {%- endif %} remove {{ image.packages_to_delete }}{%- endif %}
+{%- endif %}
+{% if image.stable %}# changing user id and group id created by package installation to stable values
+{{ DOCKERFILE_RUN }} \\
+    OLD_GID=$(chroot /target getent group {{ image.stable.group_name }}| cut -d: -f3);\\
+    OLD_UID=$(chroot /target id -u {{ image.stable.user_name }});\\
+    chroot /target groupmod -g {{ image.stable.group_id }} {{ image.stable.group_name }}; \\
+    chroot /target usermod -u {{ image.stable.user_id }} {{ image.stable.user_name }}; \\
+    chroot /target  find / \\( -gid $OLD_GID -o -uid $OLD_UID \\) -exec chown {{ image.stable.user_id }}:{{ image.stable.group_id }} {} \\;
```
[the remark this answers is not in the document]
rcmadhankumar: tested and fixed.

src/bci_build/templates.py
```diff
@@ -53,7 +53,18 @@
       {%- else %}export PERMCTL_ALLOW_INSECURE_MODE_IF_NO_PROC=1;
       {%- endif %} \\
     {% endif -%} zypper -n {%- if image.from_target_image %} --installroot /target --gpg-auto-import-keys {%- endif %} install {% if image.no_recommends %}--no-recommends {% endif %}{{ image.packages }}{%- if image.packages_to_delete %}; \\
-    zypper -n {%- if image.from_target_image %} --installroot /target {%- endif %} remove {{ image.packages_to_delete }}{%- endif %}{%- endif %}
+    zypper -n {%- if image.from_target_image %} --installroot /target {%- endif %} remove {{ image.packages_to_delete }}{%- endif %}
+{%- endif %}
+{% if image.stable %}# changing user id and group id created by package installation to stable values
+{{ DOCKERFILE_RUN }} \\
+    OLD_GID=$(chroot /target getent group {{ image.stable.group_name }}| cut -d: -f3);\\
```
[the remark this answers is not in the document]
rcmadhankumar: added a condition and make changes only if the user and group exist.

src/bci_build/package/__init__.py
```diff
@@ -77,6 +77,22 @@ class ParseVersion(enum.StrEnum):
     OFFSET = enum.auto()
 
 
+@dataclass
+class Stable:
```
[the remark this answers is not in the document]
rcmadhankumar: changed.

src/bci_build/package/postgres.py
```diff
@@ -20,6 +21,9 @@
 
 # first list the SLE15 versions, then the TW specific versions
 _POSTGRES_MAJOR_VERSIONS = [17, 16, 15, 14] + [13]
+# postgres versions that supports stable user id and stable group ids
+# added 17 just for testing, will be supported only from 18
```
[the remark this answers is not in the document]
rcmadhankumar: this will be shipped with 18. however postgres 18 is not available right now.
Which means for the demo we cannot see the dockerfile from github ci incase if I had added version as 17 here.

Once the PR is approved, then i will change this value to 18 and we can merge this PR. the change will stay and will never get executed until we release postgres 18 image.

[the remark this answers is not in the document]
rcmadhankumar: made changes to the comment and changed to expected value.
PR is ready for merge.

dirkmueller (changes requested):
Please add migration instructions to the README.md that explain what is happening?

src/bci_build/templates.py
```diff
@@ -53,7 +53,16 @@
       {%- else %}export PERMCTL_ALLOW_INSECURE_MODE_IF_NO_PROC=1;
       {%- endif %} \\
     {% endif -%} zypper -n {%- if image.from_target_image %} --installroot /target --gpg-auto-import-keys {%- endif %} install {% if image.no_recommends %}--no-recommends {% endif %}{{ image.packages }}{%- if image.packages_to_delete %}; \\
-    zypper -n {%- if image.from_target_image %} --installroot /target {%- endif %} remove {{ image.packages_to_delete }}{%- endif %}{%- endif %}
+    zypper -n {%- if image.from_target_image %} --installroot /target {%- endif %} remove {{ image.packages_to_delete }}{%- endif %}
+{%- endif %}
+{% if image.stable_user %}# changing user id and group id created by package installation to stable values
+{{ DOCKERFILE_RUN }} \\
+    if {% if image.from_target_image %}chroot /target {% endif %}getent passwd {{ image.stable_user.user_name }} && {% if image.from_target_image %}chroot /target {% endif %}getent group {{ image.stable_user.group_name }}; then\\
```
dirkmueller, line 60: why are we checking for the user to exist? IMHO this should hard fail otherwise we do not find mismatches at container build time - which is dangerous imho.

rcmadhankumar: removed

src/bci_build/package/postgres.py
```diff
@@ -84,7 +88,16 @@
         ],
         volumes=["$PGDATA"],
         exposes_ports=[TCP(5432)],
-        custom_end=rf"""COPY docker-entrypoint.sh /usr/local/bin/
+        stable_user=StableUser(
```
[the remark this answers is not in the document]
rcmadhankumar: added.

src/bci_build/package/__init__.py
```diff
@@ -194,6 +210,9 @@ class BaseContainerImage(abc.ABC):
     #: binary and its parameters as a list
     entrypoint: list[str] | None = None
 
+    # user&group that needs to be stabilized for the package
+    stable_user: StableUser | None = None
```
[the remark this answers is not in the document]
rcmadhankumar: changed.

rcmadhankumar:
> Please add migration instructions to the README.md that explain what is happening?

Added a readme section that explains the user and group id changes in postgres version 18.
(Since, we are not supporting migration on postgress yet, I am assuming user is going to take the backup as sql file and execute it and added info considering the same use case)

PR #2844, opened 2025-08-08
Check if secrets are set

alexandrevicenzi:
If not, skip the build as the job will fail

### DIFF
--- a/.github/workflows/obs_build.yml
+++ b/.github/workflows/obs_build.yml
@@ -5,7 +5,37 @@ on:
   pull_request:
 
 jobs:
+  check-secrets:
+    name: Check project secrets
+    runs-on: ubuntu-latest
+    outputs:
+      have-secrets: ${{ steps.check.outputs.have-secrets }}
+    steps:
+      - id: check
+        run: |
+          missing=()
+
+          if [ -z "${{ secrets.CHECKOUT_TOKEN }}" ]; then
+            missing+=("CHECKOUT_TOKEN")
+          fi
+
+          if [ -z "${{ secrets.OSC_PASSWORD }}" ]; then
+            missing+=("OSC_PASSWORD")
+          fi
+
+          if [ ${#missing[@]} -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/obs_build.yml
```diff
@@ -5,7 +5,37 @@ on:
...
+          if [ -z "${{ secrets.OSC_PASSWORD }}" ]; then
+            missing+=("OSC_PASSWORD")
+          fi
+
+          if [ ${#missing[@]} -eq 0 ]; then
+            echo "✅ All required secrets are set."
+            echo "have-secrets=true" >> $GITHUB_OUTPUT
+          else
+            echo "❌ Missing required secrets: ${missing[*]}"
+            echo "⚠️ PRs must be sent from branches, not forks!"
```
dirkmueller, line 31: this is slightly contradicting. either we tell people to set the secrets (which they could on forks) or we forbid running from forks. that would be simpler to achieve: 

```
  if [ ${{github.repository }} ] != "SUSE/BCI-dockerfile-generator" ]; then
```

alexandrevicenzi: You could run on a fork if you have the proper secrets. It is possible to share secrets with forks, but this is security issue, so a no-go IMHO.

Not everyone can create a branch to send a PR, so yes, the message does not help much.

We could skip the run instead of failing with an error, not sure what is the best user experience here.

.github/workflows/obs_build.yml
```diff
@@ -5,7 +5,37 @@ on:
...
+            echo "have-secrets=false" >> $GITHUB_OUTPUT
+            exit 1
+          fi
+
   obs-build:
+    needs: check-secrets
+    if: needs.check-secrets.outputs.have-secrets == 'true'
```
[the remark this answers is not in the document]
alexandrevicenzi: yes, the if might be removed if we dont write the file on failures perhaps, because we current write a value and read the value.

PR #2858, opened 2025-08-14
Add kubevirt containers

dirkmueller:


alexandrevicenzi:
Overall LGTM, but I lack context. Why are we adding this? Can you describe it in the description or link to a BZ/Jira ticket?

PR #2889, opened 2025-08-30
git was updated to 2.51.0

msmeissn:


alexandrevicenzi:
I just merged #2887, which seems to be the same change as this, but an automated one.

msmeissn:
then we dont need it anymore

PR #2900, opened 2025-09-09
Fix version checks for SL 16.0, introduce 16.1

dirkmueller:


dirkmueller:
> The build for 16.1 is broken, is this expected?

it's a bit of a mess with the staging bot, let me see if I can fix that (rename of `images` repository to `containerkiwi` which we haven't reflected in our automation yet)

dirkmueller:
> > The build for 16.1 is broken, is this expected?
> 
> it's a bit of a mess with the staging bot, let me see if I can fix that (rename of `images` repository to `containerkiwi` which we haven't reflected in our automation yet)

@alexandrevicenzi it is working now (the remaining issue is that bci-base is still building as 16.0 but we're expecting 16.1, that will resolve itself in the future).

PR #2920, opened 2025-09-12
Suffixing fips images to add info about the inclusion of fips certified binaries

rcmadhankumar:


dirkmueller:
Now this reads

` SLE BCI 15 SP7 FIPS-140-3 without FIPS certified binaries Micro: Suitable for deploying static binaries`

which I don't find readable. also, this is already written as such in the README (we could make it bold in there to make it more obvious for users). 

I think we should not have the text "without FIPS certified binaries", because what is included is still going through the vendor assertion tests, so I would suggest to do just nothing. 

for the "with FIPS", I think something like

`SLE BCI 15 SP7 Micro with FIPS-140-3 certified binaries: suitable for deploying static binaries"

instead.

dirkmueller (changes requested):
see prev comment

src/bci_build/package/basecontainers.py
```diff
@@ -190,10 +190,17 @@ def _get_fips_custom_env() -> str:
 
 
 def _get_fips_pretty_name(os_version: OsVersion) -> str:
-    if os_version == OsVersion.SP3:
-        return f"{os_version.pretty_os_version_no_dash} FIPS-140-2"
+    if os_version.is_ltss and os_version == OsVersion.SP3:
```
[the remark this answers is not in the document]
rcmadhankumar: SP3 uses FIPS-140-2 and SP4 uses FIPS-140-3. Fixed the code and modified conditional statements accordingly.

rcmadhankumar:
> > What if we call the certified containers SUSE Linux Enterprise 15 SP3 (LTSS) FIPS-140-2 or SUSE Linux Enterprise 15 SP4 (LTSS) FIPS-140-3, and the non-certified one just SLE BCI 15 SP6 FIPS-Enabled, without mentioning the version, since they are not certified?
> 
> "-2" and "-3" is not a version suffix. it is a different standard.



or Shall we just add `enabled` suffix to the non ltss containers?
for example: SLE BCI 15 SP6 with FIPS-140-3 enabled

src/bci_build/package/base-fips/README.md.j2
```diff
@@ -49,7 +49,7 @@ by NIST.
 
 {% else %}
 This base container image is configured with FIPS mode enabled by default, but
-does not include any certified binaries.
+**does not** include any certified binaries.
```
alexandrevicenzi, line 52: I also suggested before that this sentence alone is not of much help.

I would like to see something like:

This base container image is configured with FIPS mode enabled by default, but does not include any certified binaries (e.g., OpenSSL and libgcrypt modules).

With FIPS mode enabled, only FIPS-approved algorithms are allowed. For legal compliance, FIPS mode enabled is not enough without certified binaries.

If you need a certified FIPS image, you should use the LTSS version, which includes OpenSSL and libgcrypt modules that have been interim validated to FIPS 140-3.

src/bci_build/package/basecontainers.py
```diff
@@ -190,10 +190,15 @@ def _get_fips_custom_env() -> str:
 
 
 def _get_fips_pretty_name(os_version: OsVersion) -> str:
-    if os_version == OsVersion.SP3:
-        return f"{os_version.pretty_os_version_no_dash} FIPS-140-2"
+    if os_version.is_ltss:
```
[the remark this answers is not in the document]
alexandrevicenzi: Sounds good.

src/bci_build/package/basecontainers.py
```diff
@@ -192,10 +190,15 @@ def _get_fips_custom_env() -> str:
...
+        if os_version == OsVersion.SP3:
+            return f"{os_version.pretty_os_version_no_dash} with FIPS-140-2 certified binaries"
+        elif os_version == OsVersion.SP4:
+            return f"{os_version.pretty_os_version_no_dash} with FIPS-140-3 certified binaries"
+
     if os_version.is_sle15 or os_version.is_sl16 or os_version.is_tumbleweed:
         return f"{os_version.pretty_os_version_no_dash} FIPS-140-3"
```
alexandrevicenzi, line 200:
```suggestion
        return f"{os_version.pretty_os_version_no_dash} in FIPS-140-3 mode"
```

src/bci_build/package/basecontainers.py
```diff
@@ -192,10 +190,15 @@ def _get_fips_custom_env() -> str:
 
 
 def _get_fips_pretty_name(os_version: OsVersion) -> str:
-    if os_version == OsVersion.SP3:
-        return f"{os_version.pretty_os_version_no_dash} FIPS-140-2"
+    if os_version.is_ltss:
+        if os_version == OsVersion.SP3:
+            return f"{os_version.pretty_os_version_no_dash} with FIPS-140-2 certified binaries"
```
alexandrevicenzi, line 195:
```suggestion
            return f"{os_version.pretty_os_version_no_dash} FIPS-140-2"
```

src/bci_build/package/basecontainers.py
```diff
@@ -192,10 +190,15 @@ def _get_fips_custom_env() -> str:
...
+        elif os_version == OsVersion.SP4:
+            return f"{os_version.pretty_os_version_no_dash} with FIPS-140-3 certified binaries"
```
alexandrevicenzi, line 197:
```suggestion
            return f"{os_version.pretty_os_version_no_dash} FIPS-140-3"
```

alexandrevicenzi (commented):
If I understood correctly from comments, this is how it should be named.

rcmadhankumar:
resolved the comments.

dirkmueller (approved):

PR #2934, opened 2025-09-24
Run nginx as a non root user

rcmadhankumar:
Nginx can be run as non-root user(nginx user) as follows:
```
 podman run -it -p 8080:8080 -u 499:498 <nginx_bci_image>
 ```
where 499 and 498 are the UID and GID of nginx user.

### DIFF
--- a/src/bci_build/package/appcontainers.py
+++ b/src/bci_build/package/appcontainers.py
@@ -278,8 +278,11 @@ def _get_nginx_kwargs(os_version: OsVersion):
         "build_recipe_type": BuildType.DOCKER,
         "extra_files": _NGINX_FILES,
         "support_level": SupportLevel.L3,
-        "exposes_ports": [TCP(80)],
+        "exposes_ports": [TCP(8080)],
         "custom_end": f"""{version_check_lines}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/bci_build/package/appcontainers.py [hunk 1]
```diff
@@ -278,8 +278,11 @@ def _get_nginx_kwargs(os_version: OsVersion):
         "build_recipe_type": BuildType.DOCKER,
         "extra_files": _NGINX_FILES,
         "support_level": SupportLevel.L3,
-        "exposes_ports": [TCP(80)],
+        "exposes_ports": [TCP(8080)],
         "custom_end": f"""{version_check_lines}
+# Modify the default Nginx config to listen on port 8080 and PID file location to a writable path
+{DOCKERFILE_RUN} sed -i 's/listen  *80;/listen 8080;/g' /etc/nginx/nginx.conf &&\\
```
alexandrevicenzi, line 284: This is a breaking change for anyone using the container as is.

In my view, if you want to change the container port and user, you must provide your own nginx.conf file.

I would just downgrade the nginx process by setting the worker user to nginx.

dirkmueller (changes requested):
see inline comment

PR #2936, opened 2025-09-25
Suppress base.name for single-stage builds

dirkmueller:
There is an issue with generating the baseref which blocks releasing updates, so avoid this for now.


alexandrevicenzi (commented):
The unit test is broken, needs adjustment as well.

alexandrevicenzi (approved):

PR #2937, opened 2025-09-25
Use the correct systemd-defaults package

dirkmueller:
Otherwise zypper/OBS picks a random one which flips between individual builds.


alexandrevicenzi (approved):
There's a failure on TW, but it looks unrelated.

PR #2939, opened 2025-09-25
Switch from Java 24 to 25 for Tumbleweed

alexandrevicenzi:


dirkmueller (changes requested):
either don't drop 24 right away but add 25 first, or update `src/bci_build/package/apache_tomcat.py` to openjdk25 also

PR #2999, opened 2025-10-06
fix: Include diff in kubectl to make kubectl diff work

e-minguez:
TIL kubectl diff uses "diff" https://kubernetes.io/docs/reference/kubectl/generated/kubectl_diff/ and https://github.com/kubernetes/kubernetes/blob/release-1.33/staging/src/k8s.io/kubectl/pkg/cmd/diff/diff.go#L134


alexandrevicenzi (changes requested):
The code is failing formatting checks

e-minguez:
Closed as per @dirkmueller comment.

PR #3034, opened 2025-10-13
Remove last password change from /etc/shadow

alexandrevicenzi:
The last password change field in /etc/shadow changes with every rebuild because the file is generated using the current installation time.

Because this file is created after the installation, there's no easy way to use SOURCE_DATE_EPOCH without affecting the package itself.

Since we don't set any passwords for any users, it is safe to remove it in most cases as the accounts are system accounts, either locked or with no usable password.

See jsc#BCI-397.

### DIFF
--- a/src/bci_build/package/basecontainers.py
+++ b/src/bci_build/package/basecontainers.py
@@ -112,14 +112,17 @@ INIT_CONTAINERS = [
             "usage": "This container should only be used to build containers for daemons. Add your packages and enable services using systemctl."
         },
         logo_url="https://opensource.suse.com/bci/SLE_BCI_logomark_green.svg",
-        custom_end=textwrap.dedent(
+        build_stage_custom_end=textwrap.dedent(
             f"""
             {DOCKERFILE_RUN} install -d -m 0755 /etc/systemd/system.conf.d/ \\
                 && printf "[Manager]\\nLogColor=no" > \\
                     /etc/systemd/system.conf.d/01-sle-bci-nocolor.conf
             {DOCKERFILE_RUN} {_DISABLE_GETTY_AT_TTY1_SERVICE}
             {DOCKERFILE_RUN} useradd --no-create-home --uid 497 systemd-coredump
-
+            """
+        ),
+        custom_end=textwrap.dedent(
+            """
             HEALTHCHECK --interval=5s --timeout=5s --retries=5 CMD ["/usr/bin/systemctl", "is-active", "multi-user.target"]
             """
         ),


src/bci_build/templates.py
```diff
@@ -65,9 +65,13 @@
...
 {{ image.build_stage_custom_end }}
 {%- endif %}
-{% if image.packages %}{{ DOCKERFILE_RUN }} zypper -n {%- if image.from_target_image %} --installroot /target {%- endif %} clean -a; \\
+{% if image.packages %}
+# cleanup logs and temporary files
+{{ DOCKERFILE_RUN }} zypper -n {%- if image.from_target_image %} --installroot /target {%- endif %} clean -a; \\
     {{ LOG_CLEAN }}
 {%- endif %}
+# set the day of last password change to empty
+{{ DOCKERFILE_RUN }} sed -i 's/^\\([^:]*:[^:]*:\\)[^:]*\\(:.*\\)$/\\1\\2/' {% if image.from_target_image %}/target{% endif %}/etc/shadow
```
dirkmueller: what if /etc/shadow doesn't exist, is that possible?

I wonder if we should use `chage -d -1` instead (and only run it if that command exists). 

After a 2nd thought: do we actually need this once passing in a date timestamp for reproducible builds is implemented? because then the container will be always build on the "same day" and hence will always have the same content?

src/bci_build/templates.py
```diff
@@ -65,9 +65,13 @@
 {%- if image.build_stage_custom_end %}
 {{ image.build_stage_custom_end }}
 {%- endif %}
-{% if image.packages %}{{ DOCKERFILE_RUN }} zypper -n {%- if image.from_target_image %} --installroot /target {%- endif %} clean -a; \\
+{% if image.packages %}
+# cleanup logs and temporary files
```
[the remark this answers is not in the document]
alexandrevicenzi: My idea here was to add a reason in the Dockerfile why we are running such a command. Yes, it causes all containers to diff, but all containers are changing already with my sed line.

I don't mind removing it, but as all Dockerfiles are already changing, I did not see a huge churn here.

src/bci_build/templates.py
alexandrevicenzi: I did not see a container that does not have `/etc/shadow`, not sure if it is possible in BCI.

Good point about `chage`, but it requires knowing the user login, while `sed` does not. Perhaps we could run this for known users like `root` and any other stable user.

`/etc/shadow` gets the current OS date to calculate the value. The only way to fix it without sed is by using libfaketime. This file is not reproducible because it is not part of the build of the RPMs, it is created during installation time, similar to log files.

AFAIK, OBS does not change the build host time, and I'm not aware of a way to inject a fake time inside the container during build time. SOURCE_DATE_EPOCH does not work here because `system-user-root` does not respect it. The last password change is not reproducible by design, as it can have side effects from system policies and may have already expired.

PR #3071, opened 2025-10-23
Update nginx image's readme to run as non root

rcmadhankumar:


src/bci_build/package/nginx/README.md.j2
```diff
@@ -42,6 +42,37 @@ The template above is then rendered to `/etc/nginx/conf.d/default.conf` as follo
 listen  80;
 ```
 
+## Running nginx as a non-root user
+It is possible to run the image as a less privileged arbitrary UID/GID. This, however,  requires modification of nginx configuration to use directories writeable by that specific UID/GID pair:
```
alexandrevicenzi: Less wordy, same meaning.

```suggestion
It is possible to run the image as a less privileged user. This requires updating the nginx configuration to use directories writable by a specific UID/GID:
```

src/bci_build/package/nginx/README.md.j2
```diff
@@ -42,6 +42,37 @@ The template above is then rendered to `/etc/nginx/conf.d/default.conf` as follo
...
+```ShellSession
+$ podman run -it --user nginx --rm -p 8080:8080 -v /path/to/html/:/srv/www/htdocs/:Z -v $PWD/nginx.conf:/etc/nginx/nginx.conf:Z {{ image.pretty_reference }}
+```
+
+where nginx.conf in the current directory should have the following directives re-defined:
```
alexandrevicenzi:
```suggestion
Define a new path for the process ID in `nginx.conf`:
```

rcmadhankumar:
> rather than documenting how to fix it, can we just make the image runnable as non-root, e.g. the nginx user?

running as non root requires port changes in the image and we do not want to make port changes in the image.
Also, 

Upstream has 2 options now.
1.running standard nginx image as non root user https://hub.docker.com/_/nginx/#running-nginx-as-a-non-root-user

2. they have a separate non previleged image https://github.com/nginx/docker-nginx-unprivileged/blob/main/Dockerfile-debian.template#L129

I felt option 1 is suitable, because the image is already capable of running as non root without making any more changes to it.

src/bci_build/package/nginx/README.md.j2
```diff
@@ -42,6 +42,37 @@ The template above is then rendered to `/etc/nginx/conf.d/default.conf` as follo
...
 
+## Running nginx as a non-root user
+It is possible to run the image as a less privileged user. This requires updating the nginx configuration to use directories writable by a specific UID/GID:
+```ShellSession
+$ podman run -it --user nginx --rm -p 8080:8080 -v /path/to/html/:/srv/www/htdocs/:Z -v $PWD/nginx.conf:/etc/nginx/nginx.conf:Z {{ image.pretty_reference }}
+```
+
+Define a new path for the process ID in `nginx.conf`:
+
+```ShellSession
```
dirkmueller:
```suggestion
```Nginx
```

src/bci_build/package/nginx/README.md.j2
```diff
@@ -42,6 +42,37 @@ The template above is then rendered to `/etc/nginx/conf.d/default.conf` as follo
...
+pid        /tmp/nginx.pid;
+```
+
+Adjust the HTTP context paths and ports in `nginx.conf`:
+```ShellSession
```
dirkmueller:
```suggestion
```Nginx
```

PR #3096, opened 2025-10-29
Add SL16 family lifecycle information

dirkmueller:


dirkmueller:
> CI failures seem to be unrelated to this change.

yeah, it is the azure outage that was affecting half of the github world yesterday

PR #3116, opened 2025-11-04
added service entry for post-build-checks-containers

msmeissn:


src/bci_build/templates.py
```diff
@@ -232,6 +232,9 @@
   </service>
 {%- endfor -%}
 {% endfor %}
+{%-     if image.post_build_checks_containers %}
```
alexandrevicenzi: why there's so much whitespace here?

PR #3130, opened 2025-11-06
Fix .NET image description

alexandrevicenzi:
Align .NET to also use `based_on_container` as the other images.


alexandrevicenzi:
This should fix the reason https://build.suse.de/request/show/393363 was declined.

PR #3181, opened 2025-11-24
cannot use checkout@v4 together with create-pull-request custom token

dirkmueller:
the remote fails with "duplicate authentication header" then. we probably have to find a better solution for this..


dirkmueller:
7.0.9 works again. so we just need to update the permissions.

PR #3184, opened 2025-11-24
Add post-build-checks-containers for Busybox

alexandrevicenzi:
It is also required for AppCo.


dirkmueller (approved):
let's hope the build failure is a fluke

PR #3198, opened 2025-11-27

bci-pushman:


dirkmueller:
/vc packages=helm-image update for 4.0
Failed to update the changelog, see the [workflow run](https://github.com/SUSE/BCI-dockerfile-generator/actions/runs/19735510089) for further details.

PR #3264, opened 2025-12-18

bci-pushman:


alexandrevicenzi:
The build is failing for x86 on TW, it still picks up 4.22, while on aarch64 it works. Don't understand why, because 4.23 is in Factory. It might not be in the Snapshot yet.

Since the project repos take x86 from Snapshot and other architectures from their respective project, there is a chance for divergence in this case.

dirkmueller:
given that the same containers are dropped, the build failure does not matter anymore.

PR #3331, opened 2026-01-08
Add non-base containers for 16.1

dirkmueller:


src/bci_build/package/appcontainers.py
```diff
@@ -211,7 +211,11 @@ def _generate_prometheus_family_healthcheck(port: int) -> str:
         exposes_ports=[TCP(_ALERTMANAGER_PORT)],
         custom_end=_generate_prometheus_family_healthcheck(_ALERTMANAGER_PORT),
     )
-    for os_version in {v for v in ALL_NONBASE_OS_VERSIONS if v != OsVersion.SL16_0}
+    for os_version in {
```
alexandrevicenzi: nit: this tiny bit of code is in many places, it could become a variable or function somewhere, but I dont have any good suggestions for what to call it.

dirkmueller: I agree. I was thinking about something like `ALL_NONSLFO_NONBASE_OS_VERSIONS` but I found the double negations a bit too much. 

maybe we use simply `(OsVersion.SP7, OsVersion.Tumbleweed)` instead. I'll update this.

PR #3626, opened 2026-03-27
Running kubectl container as a rootless user

rcmadhankumar:


src/bci_build/package/kubectl/README.md.j2
```diff
@@ -29,6 +29,17 @@ podman run --rm --name kubectl \
       -v /localpath/to/customize:/home/user:Z
       {{ image.pretty_reference }} kustomize --enable-helm
 ```
+## Running as a non-root user
 
+By default, this container runs as the `root` user. However, it is prepared to run as a non-privileged user (UID `999`) for enhanced security.
```
alexandrevicenzi:
```suggestion
By default, this container runs as the `root` user. However, the image is also prepared to run as a non-privileged user (UID `999`) for enhanced security.
```

alexandrevicenzi (approved):

src/bci_build/package/kubectl/README.md.j2
```diff
@@ -29,6 +29,17 @@ podman run --rm --name kubectl \
...
+
+To run the container in rootless mode, you must explicitly set the user and the home environment variables:
+
+```ShellSession
+podman run --rm --name kubectl \
+      --user 999:100 \
+      -e HOME=/home/user \
+      -v /localpath/to/kubeconfig:/home/user/.kube/config:Z \
+      {{ image.pretty_reference }} get nodes
```
dirkmueller: why does it list "get nodes" here?

src/bci_build/package/kubectl/README.md.j2
```diff
@@ -29,6 +29,17 @@ podman run --rm --name kubectl \
...
 ```
+## Running as a non-root user
 
+By default, this container runs as the `root` user. However, it is prepared to run as a non-privileged user (UID `999`) for enhanced security.
+
+To run the container in rootless mode, you must explicitly set the user and the home environment variables:
+
+```ShellSession
+podman run --rm --name kubectl \
+      --user 999:100 \
```
dirkmueller: why do we hardcode this here? `--user user` works as well.

src/bci_build/package/kubectl/README.md.j2
```diff
@@ -29,6 +29,17 @@ podman run --rm --name kubectl \
       -v /localpath/to/customize:/home/user:Z
       {{ image.pretty_reference }} kustomize --enable-helm
 ```
+## Running as a non-root user
```
dirkmueller: maybe "regular user" ?

src/bci_build/package/kubectl/README.md.j2
dirkmueller: this is a bit misleading, because how the container is being run is not specified by the container. the only thing we specify is the the user within the container. how you map that is dependent on how you invoke it. 

I think we should simply say 

"The container provides a preconfigured user called "user" which can be explicitly set to lower the permissions of kubectl within the container".

src/bci_build/package/kubectl/README.md.j2
```diff
@@ -29,6 +29,17 @@ podman run --rm --name kubectl \
...
 
+By default, this container runs as the `root` user. However, it is prepared to run as a non-privileged user (UID `999`) for enhanced security.
+
+To run the container in rootless mode, you must explicitly set the user and the home environment variables:
+
+```ShellSession
+podman run --rm --name kubectl \
+      --user 999:100 \
+      -e HOME=/home/user \
```
dirkmueller: do we need this when the -v mount below is explicityly stating /home/user ?

src/bci_build/package/kubectl/README.md.j2
rcmadhankumar: modified.

src/bci_build/package/kubectl/README.md.j2
```diff
@@ -29,6 +29,15 @@ podman run --rm --name kubectl \
...
       {{ image.pretty_reference }} kustomize --enable-helm
 ```
+## Running as a regular user
 
+The container provides a preconfigured user called "user" which can be explicitly set to lower the permissions of kubectl within the container
+
+
+```ShellSession
+podman run --rm --name kubectl \
+      --user user \
```
alexandrevicenzi, line 39: nit: the user could be called kube, because calling a user user is a bit strange.

alexandrevicenzi (approved):

src/bci_build/package/kubectl/README.md.j2
dirkmueller: I don't think we can rename the user as that will be a breaking hange (we already had this user beforehand)

dirkmueller (approved):

PR #3667, opened 2026-04-13
Stop accessing the third party repo outside prepare_template()

dirkmueller:
We don't want to access multiple 3rd party repositories when not rendering a template.


dirkmueller:
obsolete

PR #3671, opened 2026-04-14
Remove hardcoded package version from NVIDIA image

alexandrevicenzi:
Grab the released package version from SCC, the only true source of truth for all codestreams.
Save all known versions to `package_versions.json`.


src/bci_build/package/versions.py
```diff
@@ -35,71 +51,70 @@
...
-#: The value is a dictionary, where the key is the code stream and the value the
-#: version of the package
-_PACKAGE_VERSIONS_T = dict[str, _PKG_VERSION_T]
+
+PackageVersions = Dict[str, PackageInfo]
 
 #: Path to the json file where the package versions are stored
 PACKAGE_VERSIONS_JSON_PATH = Path(__file__).parent / "package_versions.json"
 
+_VER_RE = r"""
```
dirkmueller: this triggers a bit the "YAGNI" vibe for me. I appreciate you added extra unit tests for it, thanks. The previous logic of "split by '-' and then by '.'" seemed way easier to me though. less code that can fail.

src/bci_build/package/package_versions.json
```diff
@@ -1,108 +1,353 @@
...
+                "latest": "3.0"
+            },
+            "7": {
+                "latest": "2.6"
+            },
+            "Tumbleweed": {
+                "latest": "3.0"
+            }
+        },
+        "track_versions": false,
```
dirkmueller: I like way you implemented this, but it gets pretty verbose compared to what we had before. have you thought about using something like this instead?

```

   "kea": {
      "latest": { 
         "16.0": "3.0",
         "7.0": "3.0",
         ...
      },
      "release_history": {
         "16.0": [ "3.0-123", "3.0-124.1"],
     }
```

it would not ever add "release_history" by itself, but if it is there, it would simply amend it if necessary. the pure existence replaces the "track_versions" bool flag which is extra data to keep in sync?

src/bci_build/package/package_versions.json
```diff
@@ -1,108 +1,353 @@
...
+                    "6.12.0-160000.8",
+                    "6.12.0-160000.9",
+                    "6.12.0-160000.26",
+                    "6.12.0-160000.27"
+                ]
+            },
+            "7": {
+                "latest": "6.4.0-150700.53.34",
+                "versions": [
+                    "6.4.0-150700.51",
```
dirkmueller: do we ever need to keep the release numbers for anything other than the same version? just checking. because then we could also automatically cleanup the list as soon as the version part of the ver-release is changing. and we only need to record the release part as well (but that would be optional)

src/bci_build/package/versions.py
alexandrevicenzi: The package versions dont follow any standard. I tried to use functions from libraries we already have like rpm and packaging, but they all failed to parse some.

I assumed a regex was better than adding an extra library, and a simple str split will most likely fail in some cases.

I dont mind postponing until we face such a problem. I can make it work just for what we have today.

src/bci_build/package/package_versions.json
alexandrevicenzi: We can change the json format, I did not think much about it. If we prefer a simple and cleaner json, I can try something as you suggested.

src/bci_build/package/package_versions.json
alexandrevicenzi: I prefer to keep the full version, it is easier to understand. Unfortunately for AMD and NVIDIA we need all kernel versions ever released, but in theory, this will be needed only for a few selected packages, right now only kernel and its flavors.

We could change to only save 6.12.0 once and 160000.27 and else multiple times, but this gets trickier to handle IMHO.

dirkmueller: > I prefer to keep the full version, it is easier to understand. Unfortunately for AMD and NVIDIA we need all kernel versions ever released

ok, but only for the code stream I believe (e.g. 16.0/16.1 share the same kernel, but 16.2 won't)

src/bci_build/package/versions.py
dirkmueller: package versions do follow a standard (the rpm versioning standard ;). they are always "some mess split by dots - some mess splitted by dots". there can only ever be one "-", this is enforced by RPM. Exactly how the mess is ordered etc is difficult (numerically if it looks like a number, otherwise alphabetically), but we don't care about that aspect. we so far only splitted by the number of components. 

I think for kernel we just want "full version" + "everything exceppt the last component of the release part".

alexandrevicenzi: I removed the regex, and reverted to something similar to what existed before.

All tests still passed, so good for now.

src/bci_build/package/package_versions.json
alexandrevicenzi: I changed the format to be less verbose. It looks better indeed.

src/bci_build/package/package_versions.json
alexandrevicenzi: I'm not sure we need to change anything here, the json is cleaner now, and only two packages have the release history, not that bad.

dirkmueller (approved):
Thanks!

PR #3722, opened 2026-04-28
Add Nano base container image

dirkmueller:


src/bci_build/templates.py
```diff
@@ -79,7 +79,7 @@
 {{ DOCKERFILE_RUN }} sed -i 's/^\\([^:]*:[^:]*:\\)[^:]*\\(:.*\\)$/\\1\\2/' {% if image.from_target_image %}/target{% endif %}/etc/shadow
 {% if image.from_target_image %}FROM {{ image.dockerfile_from_target_ref }}
 COPY --from=builder /target /
-{%- if not image.os_version.is_sle15 %}
+{%- if not image.os_version.is_sle15 and image.dockerfile_from_target_ref != "scratch" %}
```
alexandrevicenzi, line 82: This causes Index.db to not be removed from Micro, Micro FIPS and Busybox, it this expected? I dont think so.

dirkmueller: this is redundant, we are removing it generally before:


```
# cleanup logs and temporary files
RUN set -euo pipefail; zypper -n --installroot /target clean -a; \
    t=$(mktemp -d); mv /target/usr/lib/sysimage/rpm/Packages.db $t; rpmdb --rebuilddb --dbpath=$t; \
    rm /target/usr/lib/sysimage/rpm/*.db && mv $t/Packages.db /target/usr/lib/sysimage/rpm/; \
```

I can just remove it alltogether if that makes it clearer.

alexandrevicenzi: no need to remove it multiple times if it was already removed.

alexandrevicenzi (approved):

PR #3753, opened 2026-04-30
Ship a single version for each NVIDIA branch

alexandrevicenzi:
This simplifies things a lot, but requires updating image references (due to name change) everywhere.

### DIFF
--- a/src/bci_build/package/nvidia.py
+++ b/src/bci_build/package/nvidia.py
@@ -5,8 +5,10 @@ NVIDIA drivers are taken from CUDA repositories and use the GA kernel.
 """
 
 import textwrap
+from functools import lru_cache
 from pathlib import Path
 
+import requests
 from jinja2 import Template
 
 from bci_build.container_attributes import Arch
@@ -273,6 +275,11 @@ class NvidiaDriverBCI(ThirdPartyRepoMixin, DevelopmentContainer):
         )
         assert not self.build_stage_custom_end, (
             "Can't use `build_stage_custom_end` for ThirdPartyRepoMixin."
+        )
+
+        # ensure we do not ship versions that are not supported by NVIDIA datacenter
+        assert _is_datacenter_driver(self.version), (
+            f"Version '{self.version}' is not datacenter supported"
         )
 
         pkgs = self.fetch_rpm_packages()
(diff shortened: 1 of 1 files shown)


dirkmueller (changes requested):
this leads to a downgrade in buildversion, which will likely be problematic. can we do this only for future release (596+)? or do we need to do it for all of them for now?

we could just add the 595 as an additional buildtag for now, but keep the first tag/buildversion as it is. then that would not be a problem anymore.

alexandrevicenzi:
> this leads to a downgrade in buildversion, which will likely be problematic. can we do this only for future release (596+)? or do we need to do it for all of them for now?
> 
> we could just add the 595 as an additional buildtag for now, but keep the first tag/buildversion as it is. then that would not be a problem anymore.

We are shipping more versions than we need, and to switch to KMP drivers, it would be nice to cleanup things, because the KMP will be provided for a single 595 version, and if we need to keep it as is, it will require adding hacks because we provide two 595 versions.

I think we can keep the BuildVersion as before, using the full version. Correct?

src/bci_build/package/nvidia.py
```diff
@@ -704,31 +704,44 @@ def _get_kernel_versions(variant: str, os_version: OsVersion):
     return versions
 
 
+def _get_datacenter_driver_versions():
+    """
+    Return the datacenter driver versions supported by NVIDIA.
+    """
+    res = requests.get("https://docs.nvidia.com/datacenter/tesla/drivers/releases.json")
+
+    if res.status_code != 200:
```
dirkmueller: can we do raise_for_status() here instead of silently papering over an error response?

src/bci_build/package/nvidia.py
```diff
@@ -704,31 +704,44 @@ def _get_kernel_versions(variant: str, os_version: OsVersion):
...
+    "575.57.08",
+    "570.211.01",
+    "550.163.01",
     # G05 - Legacy
-    # ("535.288.01", True),
-    # ("535.274.02", False),
+    # 535 and older are not planned
 ]
 
+_NVIDIA_DATACENTER_DRIVER_VERSIONS = _get_datacenter_driver_versions()
```
dirkmueller: this means that we do a request to nvidia on startup. the startup is already incredibly slow as is.. I tried to moe this into prepare_template() in https://github.com/SUSE/BCI-dockerfile-generator/pull/3667 but ran out of time. we shouldn't add more though.

alexandrevicenzi: I deferred the call, just changed the logic and cached the json call to avoid multiple calls.

dirkmueller (approved):

PR #3769, opened 2026-05-04
Bump buildrelease for helm/git so that we can switch to SL16

dirkmueller:


src/bci_build/package/git.py
```diff
@@ -26,6 +26,7 @@
         tag_version=format_version(
             git_version := get_pkg_version("git", os_version), ParseVersion.MINOR
         ),
+        min_release_counter={OsVersion.SL16_0: 80},
```
alexandrevicenzi, line 30: missing import?

`NameError: name 'OsVersion' is not defined. Did you mean: 'ParseVersion'?`

dirkmueller: fixed

alexandrevicenzi (approved):

PR #3782, opened 2026-05-05
Prevent key requests outside prepare_template

alexandrevicenzi:
Requests outside prepare_template slow down startup. This moves the GPG key request to a new helper function used in prepare_template.

Because NVIDIA has its own custom template, it requires invoking prepare_extra_files as well.


alexandrevicenzi:
This is similar to #3667, which can be closed if we merge this PR.

dirkmueller (approved):

PR #3795, opened 2026-05-06
Strip out systemd from the php container

dirkmueller:


alexandrevicenzi (approved):

alexandrevicenzi:
16.1 is already broken, safe to ignore, but we should look up the missing packages.

PR #3814, opened 2026-05-11
Switch from DKMS to KMP for 595+

alexandrevicenzi:
Use the KMP drivers from SUSE for 595+ instead of DKMS drivers.


alexandrevicenzi:
This approach did not work. A new one is available in #3960.

PR #3821, opened 2026-05-12
add missing package for nvlink5 support in nvidia driver container

fcrozat:


src/bci_build/package/nvidia.py
```diff
@@ -616,6 +616,7 @@ def _get_compute_packages(
 
     if driver_branch >= 580:
         packages += [
+            ThirdPartyPackage("infiniband-diags"),
```
dirkmueller: that's not a thirdpartypackage, this is a regular package coming from SUSE. that's why the tests are failing

dirkmueller (changes requested):
see inline

src/bci_build/package/nvidia.py
fcrozat: fixing

dirkmueller (approved):

PR #3824, opened 2026-05-12
Update AMD GPU driver versions

alexandrevicenzi:
Update the images to the newer 30.x and 31.x versions.

### DIFF
--- a/src/bci_build/package/amd.py
+++ b/src/bci_build/package/amd.py
@@ -17,22 +17,12 @@ from bci_build.package.thirdparty import ThirdPartyRepo
 from bci_build.package.thirdparty import ThirdPartyRepoMixin
 from bci_build.repomdparser import RpmPackage
 
-AMD_REPOS = [
-    ThirdPartyRepo(
-        name="radeon-sles15sp7-x86_64",
-        arch=Arch.X86_64,
-        url="https://repo.radeon.com/amdgpu/7.0.3/sle/15.7/main/x86_64/",
-        key_url="https://repo.radeon.com/rocm/rocm.gpg.key",
-    ),
-]
-
-
 CUSTOM_END_TEMPLATE = Template(
     """
 # decompress modules for getKmodsToSign (kmmmodule.go) since it expects uncompressed files
 {{ DOCKERFILE_RUN }} \\
     find /lib/modules/{{ kernel_ga }} -name "*.ko.zst" -exec zstd -d --rm {} \\;; \\
-    depmod {{ kernel_ga }}
+    depmod {% if image.os_version.is_sl16 %}-m /lib/modules {% endif %}{{ kernel_ga }}
 
 # copy modules to /opt since getKmodsToSign (kmmmodule.go) expects it there
 {{ DOCKERFILE_RUN }} \\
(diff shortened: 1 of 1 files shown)


src/bci_build/package/amd.py
```diff
@@ -179,27 +197,92 @@ def fetch_rpm_packages(self) -> list[RpmPackage]:
...
+
+    return packages
+
+
+KERNEL_GA_VERSIONS = {
+    OsVersion.SP7: "6.4.0-150700.51",
+    OsVersion.SL16_0: "6.12.0-160000.5",
+}
+
+_AMD_DRIVER_VERSIONS: list[str] = ["31.20", "31.10", "30.30.3", "30.20.1"]
```
Priyankasaggu11929:
```suggestion
_AMD_DRIVER_VERSIONS: list[str] = ["31.30", "31.20", "31.10", "30.30.3", "30.20.1"]
```

new release: https://repo.radeon.com/amdgpu/31.30/sle/

alexandrevicenzi: added

src/bci_build/package/amd.py
```diff
@@ -179,27 +197,92 @@ def fetch_rpm_packages(self) -> list[RpmPackage]:
...
 
 AMD_CONTAINERS: list[AMDDriverBCI] = []
 
-for os_version in (OsVersion.SP7,):
+for os_version in (OsVersion.SP7, OsVersion.SL16_0):
     for ver in _AMD_DRIVER_VERSIONS:
+        branch = int(ver.split(".")[0])
+
+        if branch < 31 and os_version == OsVersion.SL16_0:
+            # amd gpu drivers for SLE 16.0 start from 31.x
```
dirkmueller: the comment is a bit superfluous because thats what the if condition does, so nicer would be to explain why that is. and maybe put it before the if condition. if we do not know why, the comment could be removed.. :)

src/bci_build/package/amd.py
```diff
@@ -179,27 +197,92 @@ def fetch_rpm_packages(self) -> list[RpmPackage]:
...
+_AMD_DRIVER_VERSIONS: list[str] = ["31.30", "31.20", "31.10", "30.30.3", "30.20.1"]
 
 AMD_CONTAINERS: list[AMDDriverBCI] = []
 
-for os_version in (OsVersion.SP7,):
+for os_version in (OsVersion.SP7, OsVersion.SL16_0):
     for ver in _AMD_DRIVER_VERSIONS:
+        branch = int(ver.split(".")[0])
+
+        if branch < 31 and os_version == OsVersion.SL16_0:
```
dirkmueller: would prefer for this to be "os_version != OsVersion.SP7" or "os_version not in (OsVersion.SP7,)" to make this futureproof for 16.1 and so on.

src/bci_build/package/amd.py
```diff
@@ -179,27 +197,92 @@ def fetch_rpm_packages(self) -> list[RpmPackage]:
...
+}
+
+_AMD_DRIVER_VERSIONS: list[str] = ["31.30", "31.20", "31.10", "30.30.3", "30.20.1"]
 
 AMD_CONTAINERS: list[AMDDriverBCI] = []
 
-for os_version in (OsVersion.SP7,):
+for os_version in (OsVersion.SP7, OsVersion.SL16_0):
     for ver in _AMD_DRIVER_VERSIONS:
+        branch = int(ver.split(".")[0])
```
dirkmueller: NIT: could be `int(partition(".")[0])`

src/bci_build/package/amd.py
```diff
@@ -179,27 +197,92 @@ def fetch_rpm_packages(self) -> list[RpmPackage]:
...
+        packages += [
+            # GA kernel explicitly requires gcc13
+            Package("gcc13", PackageType.BOOTSTRAP),
+            Package("suse-kernel-rpm-scriptlets", PackageType.BOOTSTRAP),
+        ]
+
+    return packages
+
+
+KERNEL_GA_VERSIONS = {
```
dirkmueller: NIT: first usage is line 200 already, maybe move it before the _get_packages function?

src/bci_build/package/amd.py
```diff
@@ -121,56 +111,84 @@ def fetch_rpm_packages(self) -> list[RpmPackage]:
...
+                    "package": "kernel-source",
+                    "no_arch": True,
+                },
+                {
+                    "name": "kernel-macros",
+                    "package": "kernel-source",
+                    "no_arch": True,
+                },
+            ]
+        else:
```
dirkmueller: elsif version == SL_16.0 (or use match/case, we can use that in BCI-dockerfile-generator)

and add an else: assert or raise RuntimeError so that we find when it doesn't match.

dirkmueller (approved):
wfm, left a couple of suggestions inline

PR #3828, opened 2026-05-12
Add a minimal pause command to the nano container

dirkmueller:
This helps with testing

### DIFF
--- a/src/bci_build/package/nano.py
+++ b/src/bci_build/package/nano.py
@@ -1,6 +1,7 @@
 """Base container images maintained by the BCI generator"""
 
 import textwrap
+from pathlib import Path
 
 from bci_build.container_attributes import PackageType
 from bci_build.container_attributes import SupportLevel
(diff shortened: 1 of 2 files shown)


src/bci_build/package/nano.py
```diff
@@ -36,15 +36,38 @@ def _get_nano_package_list(os_version: OsVersion) -> list[Package]:
...
+            "pause.c": textwrap.dedent("""
+            #include <sys/syscall.h>
+            #ifdef __aarch64__
+            #include <signal.h>
+            #endif
+            #include <unistd.h>
+            void _start() {
+
+            #ifdef __aarch64__
+                sigset_t mask = 0;
```
alexandrevicenzi: It seems that this does not build. This is a more elaborate version https://github.com/kubernetes/kubernetes/blob/master/build/pause/linux/pause.c.

src/bci_build/package/nano.py
```diff
@@ -36,15 +36,38 @@ def _get_nano_package_list(os_version: OsVersion) -> list[Package]:
         custom_description="A nano environment for containers {based_on_container}.",
         from_target_image="scratch",
         package_list=[pkg.name for pkg in _get_nano_package_list(os_version)],
+        extra_files={
+            "pause.c": textwrap.dedent("""
```
alexandrevicenzi: can we move to be a regular file in the `nano` folder alongside the README? it is easier to understand and less janky.

src/bci_build/package/nano.py
dirkmueller: @alexandrevicenzi thats basically "catatonit" then. which is like 200kb. this binary is 2kb.

src/bci_build/package/nano.py
dirkmueller: moved

alexandrevicenzi (approved):

PR #3854, opened 2026-05-15
Update registry README (bsc#1265213)

alexandrevicenzi:
Update the example from the now deprecated generate systemd command to the new Quadlet format.

### DIFF
--- a/src/bci_build/package/registry/README.md.j2
+++ b/src/bci_build/package/registry/README.md.j2
@@ -30,14 +30,41 @@ mkdir -p /var/lib/docker-registry
 Then you can start the container with the following command:
 
 ```bash
-podman run -d --restart=always -p 5000:5000 -v /path/to/config.yml:/etc/registry/config.yml \
+podman run -it --restart=always -p 5000:5000 -v /path/to/config.yml:/etc/registry/config.yml \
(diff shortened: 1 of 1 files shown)


src/bci_build/package/registry/README.md.j2
```diff
@@ -30,14 +30,41 @@ mkdir -p /var/lib/docker-registry
...
+ContainerName=registry
+PublishPort=5000:5000
+Volume=/path/to/config.yml:/etc/registry/config.yml:Z
+Volume=/var/lib/docker-registry:/var/lib/docker-registry:Z
+
+[Service]
+Restart=always
+
+[Install]
+WantedBy=multi-user.target
```
dirkmueller, line 56: do we really want an Install section?

dariavladykina (approved):

dirkmueller (approved):

src/bci_build/package/registry/README.md.j2
alexandrevicenzi: I don't know, just found some other examples and followed, it seems to be documented here https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html#enabling-unit-files

dirkmueller: I understand that part, but SUSE pollicy is to not enable services by default.

PR #3874, opened 2026-05-19
Add more containers for 16.1 (same set like 16.0)

dirkmueller:


dirkmueller:
containers build, so I guess it's okay.

PR #3909, opened 2026-05-27
Switch cosign image to bci-nano

dirkmueller:


alexandrevicenzi (approved):

src/bci_build/templates.py
```diff
@@ -79,9 +79,6 @@
 {{ DOCKERFILE_RUN }} sed -i 's/^\\([^:]*:[^:]*:\\)[^:]*\\(:.*\\)$/\\1\\2/' {% if image.from_target_image %}/target{% endif %}/etc/shadow
 {% if image.from_target_image %}FROM {{ image.dockerfile_from_target_ref }}
 COPY --from=builder /target /
-{%- if not image.os_version.is_sle15 and image.dockerfile_from_target_ref != "scratch" %}
```
alexandrevicenzi: nit: this should perhaps be in another commit to explain why it was removed.

dirkmueller: will do. build fails anyway so that's the first step to fix..

PR #3912, opened 2026-05-27
Remove compat-usrmerge-tools from AMD image

alexandrevicenzi:


dirkmueller (approved):
there gotta be a nicer way, but lets not block on this for now

PR #3960, opened 2026-06-11
Switch NVIDIA 595 driver to KMP

alexandrevicenzi:
Switch from DKMS drivers provided by NVIDIA repositories to KMP drivers provided by SLES repositories.


alexandrevicenzi:
This change has been incorporated into #4014.

PR #3975, opened 2026-06-16
Build rmt-server for SL16 as well

dirkmueller:


alexandrevicenzi:
build fails with version mismatch, 16 has 3.0.

dirkmueller:
> build fails with version mismatch, 16 has 3.0.

It's passing now

alexandrevicenzi (approved):

PR #3982, opened 2026-06-17
set a CMD for kiwi so that the README is actually valid

dcermak:


src/bci_build/package/kiwi.py
```diff
@@ -37,6 +37,7 @@ def generate_kiwi_10_config():
                 get_pkg_version("python-kiwi", os_version), ParseVersion.MINOR
             )
         ),
+        cmd=["/usr/bin/kiwi"],
```
dirkmueller, line 40: I think we should just fix the README. the readme is incorrect anyway, because /image is not the current working dir ('/') is, so this doesn't fix anything. 

worse it breaks existing documented Edge documentation on how to build refreshed micro images.

PR #3993, opened 2026-06-18

bci-pushman:


alexandrevicenzi:
The package fails to install on PPC with:

```
[   50s] /usr/sbin/useradd -r -c MySQL database admin -d /var/lib/mysql -U -u 60 mysql -s /usr/sbin/nologin
[   50s] Cannot open audit interface - aborting.
```

Not sure if a real package issue or just a problem with 16.1.

And on aarch64 the package is outdated, still on the older version, not sure how that is possible.

dirkmueller:
> And on aarch64 the package is outdated, still on the older version, not sure how that is possible.

that's just because in OBS we build against the codestream (so the very latest state of everything), and thats just in flux and broken right now.

dirkmueller (approved):

PR #3995, opened 2026-06-18
Cleanup /run leftovers

alexandrevicenzi:
Not all images have a target layer, this cleans up specific files instead of entire /run.


src/bci_build/package/__init__.py
```diff
@@ -47,7 +47,7 @@
 #: Remove various log files and temporary files. While it is possible to just ``rm -rf /var/log/*``,
 #: that would also remove some package owned directories (not %ghost)
 LOG_CLEAN: str = textwrap.dedent("""rm -rf {/target,}/var/log/{alternatives.log,lastlog,suseconnect.log,tallylog,zypper.log,zypp/history,YaST2}; \\
-    rm -rf /target/run/*; \\
+    rm -rf {/target,}/target/run/{zypp.pid,regenerate-initrd,systemd}; \\
```
dirkmueller: thanks for addressing the regression, but isn't that one `/target/` too much? am I overlooking something?

alexandrevicenzi: yes, it was a mistake, it should be fixed now.

dirkmueller (approved):

PR #3996, opened 2026-06-18
Reenable mariadb containers on riscv64

andreas-schwab:


dirkmueller (approved):
thanks!

alexandrevicenzi:
We don't have risc64 enabled in any of our projects, should we enable it for Tumbleweed @dirkmueller?

dirkmueller:
> We don't have risc64 enabled in any of our projects, should we enable it for Tumbleweed @dirkmueller?

It's actually enabled by default, we just implicitly disabled it here because of some containers being only available for 64bit platforms. like mariadb. I'll rework this PR afterwards.

PR #4014, opened 2026-06-19
Nvidia kmp driver update only

fcrozat:
Use KMP driver for SLE 16.x for open driver, enabling Secure Boot on this platform.

Due to KMP being currently built against some update kernel, running container on kernels older than the update kernel used for KMP build is not supposed.

### DIFF
--- a/src/bci_build/package/nvidia.py
+++ b/src/bci_build/package/nvidia.py
@@ -10,6 +10,7 @@ from pathlib import Path
 
 import requests
 from jinja2 import Template
+from version_utils import rpm
 
 from bci_build.container_attributes import Arch
 from bci_build.container_attributes import BuildType
@@ -31,7 +32,7 @@ from bci_build.package.thirdparty import ThirdPartyRepoMixin
 from bci_build.package.versions import get_all_pkg_version
 from bci_build.repomdparser import RpmPackage
 
-NVIDIA_REPOS = {
+_NVIDIA_REPOS = {
     OsVersion.SP7: [
         ThirdPartyRepo(
             name="cuda-sles15-x86_64",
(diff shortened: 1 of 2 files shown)


src/bci_build/package/nvidia.py
```diff
@@ -277,196 +304,43 @@ def prepare_template(self) -> None:
...
+                kmp_rpms = _get_nvidia_kmp_rpms(
+                    self.version, self.os_version, self.kernel_variant, self.exclusive_arch
+                )
+                if kmp_rpms:
+                    kmp_rpm_version = kmp_rpms[0].evr[1]
+                    if "_k" in kmp_rpm_version:
+                        built_kernel_raw = kmp_rpm_version.split("_k")[-1]
+                        built_kernel = "-".join(built_kernel_raw.split("_", 1))
+            except ValueError:
+                pass
```
alexandrevicenzi: This is not good. Such errors should usually stop the generator.

src/bci_build/package/nvidia.py
```diff
@@ -827,15 +855,33 @@ def _is_datacenter_driver(version: str):
...
+                kmp_rpms = _get_nvidia_kmp_rpms(
+                    ver, os_version, kernel_variant, exclusive_arch
+                )
+                if kmp_rpms:
+                    kmp_rpm_version = kmp_rpms[0].evr[1]
+                    if "_k" in kmp_rpm_version:
+                        built_kernel_raw = kmp_rpm_version.split("_k")[-1]
+                        built_kernel = "-".join(built_kernel_raw.split("_", 1))
+            except ValueError:
+                pass
```
alexandrevicenzi: Same as above, could this be a help function so we don't duplicate code?

alexandrevicenzi:
It needs a rebase. Could you also squash a few commits?

alexandrevicenzi (approved):

alexandrevicenzi (approved):

PR #4015, opened 2026-06-20
Use coreutils-single on SLFO and Tumbleweed

dirkmueller:
This saves about 5MB which is huge for bci-micro and bci-base


alexandrevicenzi (dismissed):

alexandrevicenzi:
it seems that there's an issue

> nothing provides /usr/bin/coreutils needed by coreutils-single

alexandrevicenzi (approved):

PR #4022, opened 2026-06-23
Fix NVIDIA driver generator

alexandrevicenzi:


pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
     "requests >= 2.34.2",
     "urllib3 >= 2.7.0",
     "version-utils >= 0.3.2",
+    "zstandard (>=0.25.0,<0.26.0)",
```
dirkmueller, line 27: no, we should be using https://pypi.org/project/backports.zstd/ for this, as this was moved into stdlib in 3.14.

src/bci_build/repomdparser.py
```diff
@@ -76,16 +77,24 @@ def parse(self):
...
-        primary_xml: ET.Element[str] = ET.fromstring(
-            gzip.decompress(primary_xml_response.content)
-        )
+
+        if primary_xml_location.endswith(".gz"):
+            xml_data = gzip.decompress(primary_xml_response.content)
+        elif primary_xml_location.endswith(".zst"):
+            size = int(primary_xml_response.headers["Content-Length"])
+            dc = zstd.ZstdDecompressor()
+            xml_data = dc.decompress(primary_xml_response.content, max_output_size=size)
```
dirkmueller, line 93: what's the point of max_output_size here? first of all Content-Length is optional as a header, so the code above might crash, and then max_output_size is on the *decompressed* length, while content-length is the *compressed* length, so these two don't match.

dirkmueller:
I made an alternative at https://github.com/SUSE/BCI-dockerfile-generator/pull/4023

src/bci_build/repomdparser.py
alexandrevicenzi: the main reason is that it was throwing an error: `could not determine content size in frame header`

PR #4057, opened 2026-07-03
remove usbredir on the SLES containers - not available

dirkmueller:


src/bci_build/package/kubevirt.py
```diff
@@ -195,7 +195,6 @@ def _get_kubevirt_kwargs(
                     "libvirt-client",
                     "qemu-hw-usb-host",
                     "qemu-hw-usb-redirect",
```
oakban, line 197: Should be fine to do the same with qemu driver

oakban: (keep usb-host for USB passtrhough)

dirkmueller: usb-redirect is on sles though, but I can remove it if you want as well? what do you recommend?

oakban: Alright. Let's keep it. I'll later fix this in the kubevirt package

PR #4085, opened 2026-07-07
Bump AMD 30.30 driver to 30.30.4

alexandrevicenzi:


src/bci_build/package/amd.py
```diff
@@ -260,7 +260,7 @@ def _get_packages(os_version: OsVersion):
     return packages
 
 
-_AMD_DRIVER_VERSIONS: list[str] = ["31.30", "31.20", "31.10", "30.30.3", "30.20.1"]
+_AMD_DRIVER_VERSIONS: list[str] = ["31.30", "31.20", "31.10", "30.30.4", "30.20.1"]
```
Priyankasaggu11929, line 263: question - will the `30.30.3` driver version tag go away from the registry?

alexandrevicenzi: yes and no, the old tag will still be there, but no new tags will be published

Priyankasaggu11929: meaning the `30.30.3` will remain, just won't be udpated/refreshed.

if so, then changes lgtm

alexandrevicenzi: Yes, you are correct. Previously published tags are not removed, but are not updated either.

dirkmueller (approved):

PR #4087, opened 2026-07-07
Cleanup code

dirkmueller:


src/bci_build/package/nvidia.py
alexandrevicenzi: If we are consuming the json file to update versions, I'm not sure if we still need this validation here, WDYT?

alexandrevicenzi (approved):

src/bci_build/package/nvidia.py
dirkmueller: @alexandrevicenzi good point, removed as well

PR #4159, opened 2026-07-22
Also switch to version-suffixed version on SL16.0

dirkmueller:


alexandrevicenzi:
build is unresolvable, is that expected?

dirkmueller:
> build is unresolvable, is that expected?

yes, it will work in IBS

PR #4160, opened 2026-07-22
set SOURCE_DATE_EPOCH on target-installation

dirkmueller:
In those cases, we do want reproducible timestamps in the RPM database.


alexandrevicenzi:
is the tw libc dependency issue expected?

dirkmueller:
> is the tw libc dependency issue expected?

it works on a rebuild, no idea what happend :shrug:

alexandrevicenzi (approved):

PR #4170, opened 2026-07-23
Convert python 3.13 to micro/base also for TW

dirkmueller:
We need this for the MCP container sets. also tighten latest tag


alexandrevicenzi:
there is a bug in the Dockerfile, where the image name is wrong, causing `unresolvable: nothing provides container:opensuse/bci/None:latest`

dirkmueller:
> there is a bug in the Dockerfile, where the image name is wrong, causing `unresolvable: nothing provides container:opensuse/bci/None:latest`

great catch, fixed..

alexandrevicenzi (approved):

PR #4187, opened 2026-07-23
Nvidia driver listed in SBOM

fcrozat:
This ensures they will be listed in SBOM and on
package list on registry web ui


src/bci_build/package/nvidia.py
```diff
@@ -360,6 +369,17 @@ def prepare_template(self) -> None:
...
                 and pkg.arch in ARCH_FILENAME_MAP[arch]
             ],
+            get_justdb_packages_for_arch=lambda arch: [
+                pkg
+                for pkg in pkgs
+                if (
+                    pkg.name in kmp_packages
+                    or pkg.name in closed_packages
+                    or pkg.name.startswith("nvidia-open-driver")
+                    or "-kmp-" in pkg.name
```
alexandrevicenzi: do we need this if we already use `pkg.name in kmp_packages`?

src/bci_build/package/nvidia.py
```diff
@@ -360,6 +369,17 @@ def prepare_template(self) -> None:
                 if pkg.name not in ignore_in_target_packages
                 and pkg.arch in ARCH_FILENAME_MAP[arch]
             ],
+            get_justdb_packages_for_arch=lambda arch: [
+                pkg
+                for pkg in pkgs
+                if (
+                    pkg.name in kmp_packages
+                    or pkg.name in closed_packages
+                    or pkg.name.startswith("nvidia-open-driver")
```
alexandrevicenzi: why not `pkg.name in open_packages`?

alexandrevicenzi (approved):

src/bci_build/package/nvidia.py
fcrozat: indeed, fixing

dirkmueller (changes requested):
please update https://github.com/SUSE/BCI-dockerfile-generator/blob/main/update-files.sh#L49 as well (and potentially the license is no longer needed? not sure)

dirkmueller (approved):

PR #4222, opened 2026-07-27
Libguestfs tools image

oakban:
This one was missing: https://build.opensuse.org/package/show/openSUSE:Factory/virt-libguestfs-tools-container

Equivalent of quay.io/kubevirt/libguestfs-tools

First commit was the initial port and then some fixes on top, feel free to squash.


dirkmueller:
moved to #4223

dirkmueller:
```nothing provides kubevirt1.8-manifests, nothing provides zerofree```

can we avoid using -manifests? this will not work as it comes from another project

PR #4223, opened 2026-07-27
[recreated for CI] Libguestfs tools image

dirkmueller:


oakban:
``` 
> nothing provides kubevirt1.8-manifests, nothing provides zerofree

can we avoid using -manifests? this will not work as it comes from another project
```

Sure

oakban:
Will probably require more changes. I'll update the other branch and let you know when ready.

src/bci_build/package/kubevirt/entrypoint.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -xe
+
+/bin/bash
```
dirkmueller: you could just set `entrypoint_script=` to `/usr/bin/bash` instead of having this wrapper script?

dirkmueller: (which is btw the default)

oakban:
The new [changes](https://github.com/SUSE/BCI-dockerfile-generator/compare/158d1d98f4bfee54c0bd40c9265c245ab9e8534d..0eb1ca2f96eb8b36a77ca52d9c87ef34a335798e) include previous review plus adding sidecar-shim image that we were missing as well.

This, to work, needs kubevirt1.8 [request](https://build.opensuse.org/request/show/1368067) to land on Factory as I moved some requirements there.

src/bci_build/package/kubevirt/entrypoint.sh
oakban: You can say it is a limitation but the file is needed. See [guestfs.go:159](https://github.com/kubevirt/kubevirt/blob/main/pkg/virtctl/guestfs/guestfs.go#L159). But this is now part of the kubevirt1.8 so it shouldn't really be a problem here, I hope.

src/bci_build/package/kubevirt.py
```diff
@@ -135,6 +135,28 @@ def _get_kubevirt_kwargs(
...
+    """
+    kwargs = _get_kubevirt_kwargs(
+        "libguestfs-tools",
+        os_version,
+        custom_end=False,
+        custom_service_pkg_name=f"{_kubevirt_pkg(os_version)}-libguestfs-tools",
+    )
+    kwargs["name"] = "libguestfs-tools"
+    kwargs["pretty_name"] = "KubeVirt libguestfs-tools"
+    del kwargs["from_target_image"]
```
dirkmueller: this makes it a single stage container build again

src/bci_build/package/kubevirt.py
```diff
@@ -280,6 +302,78 @@ def _get_kubevirt_kwargs(
...
+                + textwrap.dedent(f"""
+                # build the appliance once at image build so pods do not run
+                # supermin at startup; store the root as compressed qcow2
+                # because image layers do not preserve sparseness
+                {DOCKERFILE_RUN} mkdir -p /usr/local/lib/guestfs/appliance && \\
+                    cd /usr/local/lib/guestfs/appliance && \\
+                    LIBGUESTFS_BACKEND=direct LIBGUESTFS_DEBUG=1 libguestfs-make-fixed-appliance . && \\
+                    qemu-img convert -c -O qcow2 root root.qcow2 && \\
+                    mv root.qcow2 root && \\
+                    touch done && \\
```
dirkmueller: what is this good for?

oakban: [Ugh](https://build.opensuse.org/projects/openSUSE:Factory/packages/virt-libguestfs-tools-container/files/Dockerfile?expand=1). I just tested without it and works.

oakban: https://github.com/SUSE/BCI-dockerfile-generator/pull/4222/changes/cf95b4b62007b1cefaed4423dc25f7c7504d59a9

src/bci_build/package/kubevirt.py
oakban: https://github.com/SUSE/BCI-dockerfile-generator/pull/4222/changes/c7a834512dffcaa97d1f2068f7b5c64ee8a7c57c

PR #4234, opened 2026-07-29
Ensure blkid is set to scan

dirkmueller:


alexandrevicenzi:
Missing import? `Undefined name SET_BLKID_SCAN`

alexandrevicenzi (approved):

PR #4250, opened 2026-07-31
nginx: allow to run with any user

alexandrevicenzi:
Make 40-unprivileged-mode.sh more explicit on what is doing and double check before making any changes.

Change most nginx paths to /tmp/nginx and chmod it to 777 to give permission to any user, even the ones that do not exist in the image.

### DIFF
--- a/src/bci_build/package/nginx.py
+++ b/src/bci_build/package/nginx.py
@@ -80,17 +80,24 @@ def _get_nginx_kwargs(os_version: OsVersion):
         ),
         "build_stage_custom_end": generate_package_version_check(
             "nginx", nginx_version, use_target=True
-        ),
+        )
+        + textwrap.dedent(f"""
+            {DOCKERFILE_RUN} mkdir /target/docker-entrypoint.d
+            COPY [1-4]0-*.sh /target/docker-entrypoint.d/
+            COPY docker-entrypoint.sh /target/usr/local/bin
+            COPY index.html /target/srv/www/htdocs/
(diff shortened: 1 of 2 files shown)


alexandrevicenzi:
Test case available in https://github.com/SUSE/BCI-tests/pull/1124, the test will fail until this is merged.

src/bci_build/package/nginx.py
```diff
@@ -80,17 +80,24 @@ def _get_nginx_kwargs(os_version: OsVersion):
...
+            COPY index.html /target/srv/www/htdocs/
+            {DOCKERFILE_RUN} chmod +x /target/docker-entrypoint.d/*.sh /target/usr/local/bin/docker-entrypoint.sh
+            {DOCKERFILE_RUN} \\
+                mkdir -p /target/tmp/nginx/client_body_temp \\
+                         /target/tmp/nginx/proxy_temp \\
+                         /target/tmp/nginx/fastcgi_temp \\
+                         /target/tmp/nginx/uwsgi_temp \\
+                         /target/tmp/nginx/scgi_temp; \\
+                chmod -R 777 /target/etc/nginx \\
+                             /target/var/log/nginx \\
```
dirkmueller, line 97: why do we need 777 on //var/log/nginx when we just symlink dev/stdout and dev/stderr there?

dirkmueller: In my testing this seems to be enough:

```
RUN chmod o+rx /var/log/nginx; chown root: /var/log/nginx
```

dirkmueller: and for the pid file problem it appears

```pid /dev/null;``` 

is enough.

alexandrevicenzi: I'll test this out.

PR #4261, opened 2026-08-10
amd: switch to L3 support

alexandrevicenzi:


Priyankasaggu11929 (approved):
LGTM - both version bump + support level change.

(just adding a note here for record keeping - Support level change (`TechPreview` to `L3`) is per the approval of BCI PM)

PR #4275, opened 2026-08-14
python: always install python3-base

alexandrevicenzi:
Always install python3-base for system versions.
Otherwise, /usr/bin/python3 is not provided by any packages.

### DIFF
--- a/src/bci_build/package/python.py
+++ b/src/bci_build/package/python.py
@@ -85,6 +85,7 @@ def _get_python_kwargs(
         (["lifecycle-data-sle-module-python3"] if os_version.is_sle15 else [])
         + os_version.lifecycle_data_pkg
         + [py3]
+        + (["python3-base"] if is_system_py else [])
         + [pip3]
     )
     if build_flavor != "micro":


mloviska (approved):

src/bci_build/package/python.py
```diff
@@ -85,6 +85,7 @@ def _get_python_kwargs(
         (["lifecycle-data-sle-module-python3"] if os_version.is_sle15 else [])
         + os_version.lifecycle_data_pkg
         + [py3]
+        + (["python3-base"] if is_system_py else [])
```
dirkmueller, line 88: see https://github.com/SUSE/BCI-dockerfile-generator/pull/4275/changes#diff-ea8ac374a0fc310180130d34c25ffc85104859635385669aa4cbd163d49e0fe3L53

which OS version broke now? -base should always be required by -devel, which we already have here. I think you might want to just extend the condition above.

dirkmueller: made an experimental change here: https://github.com/SUSE/BCI-dockerfile-generator/pull/4277

alexandrevicenzi:
Closing this in favor of https://github.com/SUSE/BCI-dockerfile-generator/pull/4277.

PR #4277, opened 2026-08-17
prefer python3 for the system python on SLFO

dirkmueller:


alexandrevicenzi (approved):

dirkmueller:
still needs a change for version determination

PR #4278, opened 2026-08-18
micro-fips: add openssl.cfg (bsc#1274584)

alexandrevicenzi:
On 15.7, openssl.cnf is provided by openssl-3, however, this package is not installed on Micro. Instead, copy the file from the builder stage.


alexandrevicenzi:
Reproducer available at https://github.com/SUSE/BCI-tests/pull/1123, will fail until this is merged.

dirkmueller (approved):
LGTM. using install(1) or  `cp -p` might be warranted here though but for now it's good (crb will filter the redundancy)

PR #4287, opened 2026-08-21
Pr helper name fix

oakban:
Since [this commit](https://github.com/kubevirt/kubevirt/commit/ef6c8fbe78faf94100d5b3efe55c5079b0b67bc7) we required the entrypoint.sh.

Also let's not prefix with virt- the image name.


oakban:
Note that this particular fix still require underlying packages to ship the entrypoint.sh, which is ongoing at the moment.

dirkmueller:
rebased and moved to #4308

PR #4288, opened 2026-08-21
Restore world-readable rpmdb after the reproducibility rebuild

oakban:
Caught by the BCI test suite on the kubevirt libguestfs-tools image, where it also masqueraded as a compat-usrmerge-tools removal failure - the package is absent, but the test could not read the rpm database to see it.

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -55,9 +55,12 @@ LOG_CLEAN: str = textwrap.dedent("""rm -rf {/target,}/var/log/{alternatives.log,
     rm -f {/target,}/var/cache/ldconfig/aux-cache
 """)
 
-#: Rebuild the rpm database to make it reproducible
+#: Rebuild the rpm database to make it reproducible. The rebuilt Packages.db
+#: comes back mode 0600 (created inside the private mktemp dir); restore the
+#: stock 0644 so rpm keeps working for non-root users of the image.
 TARGET_REBUILDDB: str = """t=$(mktemp -d); mv /target/usr/lib/sysimage/rpm/Packages.db $t; rpmdb --rebuilddb --dbpath=$t; \\
-    rm /target/usr/lib/sysimage/rpm/*.db && mv $t/Packages.db /target/usr/lib/sysimage/rpm/"""
+    rm /target/usr/lib/sysimage/rpm/*.db && mv $t/Packages.db /target/usr/lib/sysimage/rpm/ && \\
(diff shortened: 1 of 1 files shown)


src/bci_build/package/__init__.py
```diff
@@ -55,9 +55,12 @@
     rm -f {/target,}/var/cache/ldconfig/aux-cache
 """)
 
-#: Rebuild the rpm database to make it reproducible
+#: Rebuild the rpm database to make it reproducible. The rebuilt Packages.db
+#: comes back mode 0600 (created inside the private mktemp dir); restore the
+#: stock 0644 so rpm keeps working for non-root users of the image.
 TARGET_REBUILDDB: str = """t=$(mktemp -d); mv /target/usr/lib/sysimage/rpm/Packages.db $t; rpmdb --rebuilddb --dbpath=$t; \\
-    rm /target/usr/lib/sysimage/rpm/*.db && mv $t/Packages.db /target/usr/lib/sysimage/rpm/"""
+    rm /target/usr/lib/sysimage/rpm/*.db && mv $t/Packages.db /target/usr/lib/sysimage/rpm/ && \\
```
Vogtinator, line 62: Why does it have 0600 in the first place? Wrong umask of 077?

src/bci_build/package/__init__.py
```diff
@@ -55,9 +55,12 @@
     rm -f {/target,}/var/cache/ldconfig/aux-cache
 """)
 
-#: Rebuild the rpm database to make it reproducible
+#: Rebuild the rpm database to make it reproducible. The rebuilt Packages.db
+#: comes back mode 0600 (created inside the private mktemp dir); restore the
+#: stock 0644 so rpm keeps working for non-root users of the image.
 TARGET_REBUILDDB: str = """t=$(mktemp -d); mv /target/usr/lib/sysimage/rpm/Packages.db $t; rpmdb --rebuilddb --dbpath=$t; \\
```
alexandrevicenzi, line 61: This affects all images. I don't think this is the issue, the libguestfs container needs a review.

alexandrevicenzi (changes requested):

src/bci_build/package/__init__.py
dirkmueller: no, the umask is correct, the issue does indeed happen in every container also.

dirkmueller:
thanks, this looks correct, I've moved a slightly adapted version to #4345

PR #4297, opened 2026-08-24
Add golang-oldoldstable

dirkmueller:


src/bci_build/package/golang.py
```diff
@@ -16,6 +16,7 @@
 
 _GO_VER_T = Literal["1.23", "1.24", "1.25", "1.26", "1.27"]
 _GOLANG_VERSIONS: list[_GO_VER_T] = ["1.26", "1.27"]
+_GOLANG_OLDOLDSTABLE_VERSION = "1.25"
```
alexandrevicenzi, line 19: 1.25 is EOL, 1.25.14 was released on 2026-08-19, but at the same time it reached EOL. Download is under Archived versions.

oldoldstable should perhaps be called eol.

alexandrevicenzi (approved):

src/bci_build/package/golang.py
dirkmueller: I chose the name that debian uses for "older than oldstable".. there might be another reason to reuse this in the future, so didn't just want to call it eol..

PR #4303, opened 2026-08-25
Add SL 16.2 support - switch over to SLFO 1.3

dirkmueller:


src/staging/project_setup.py
```diff
@@ -157,9 +157,17 @@ def generate_meta(
...
             last_prj = "SUSE:SLFO:Main"
-            if os_version == OsVersion.SL16_0:
+            if os_version in (OsVersion.SL16_0,):
                 first_prj = "SUSE:Registry"
-                last_prj = "SUSE:SLFO:Products:SLES:16.0"
+            match os_version:
+                case OsVersion.SL16_0:
+                    last_prj = "SUSE:SLFO:Products:SLES:16.0"
+                case OsVersion.SL16_1:
+                    last_prj = "SUSE:SLFO:1.3"
```
alexandrevicenzi: Should it be `SUSE:SLFO:Products:SLES:16.1`?

dirkmueller: we can try. the goal of this was to have the codestream available, as there won't be regular snapshots, and that way containers are fresher on the testing / devel projects.

alexandrevicenzi (approved):

PR #4308, opened 2026-08-25
Pr helper name fix

dirkmueller:


alexandrevicenzi:
it seems that there is a build issue:

`install: cannot stat '/usr/share/kube-virt-1.9/pr-helper/entrypoint.sh': No such file or directory`

oakban:
> it seems that there is a build issue:
> 
> `install: cannot stat '/usr/share/kube-virt-1.9/pr-helper/entrypoint.sh': No such file or directory`

There was a race between submission here and Factory. I do both more or less at same time but Factory takes a bit longer to be available.

https://build.opensuse.org/request/show/1373648

PR #4321, opened 2026-08-27
bot: Fix 16.1 project meta

alexandrevicenzi:
A duplicated key causes a 400 error when updating the project meta.

### DIFF
--- a/src/staging/project_setup.py
+++ b/src/staging/project_setup.py
@@ -157,10 +157,9 @@ def generate_meta(
             assert os_version.is_sl16
             first_prj = f"SUSE:SLFO:Products:SLES:{os_version}"
             last_prj = "SUSE:SLFO:Main"
-            if os_version in (OsVersion.SL16_0,):
-                first_prj = "SUSE:Registry"
             match os_version:
                 case OsVersion.SL16_0 | OsVersion.SL16_1:
+                    first_prj = "SUSE:Registry"
                     last_prj = f"SUSE:SLFO:Products:SLES:{os_version}"
                 case OsVersion.SL16_2:
                     last_prj = "SUSE:SLFO:Main"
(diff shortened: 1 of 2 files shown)


alexandrevicenzi:
The bot works with this change, but the build has choices, and I believe it is caused by `devel:BCI:16.1` not having these lines in the project config:

```
# Avoid the unreleased containers
Prefer: -container:base-image -container:bci-micro -container:bci-bci-micro -container:bci-bci-init
```

But I'm not 100% sure that this is the case.

src/staging/project_setup.py
```diff
@@ -157,10 +157,9 @@ def generate_meta(
             assert os_version.is_sl16
             first_prj = f"SUSE:SLFO:Products:SLES:{os_version}"
             last_prj = "SUSE:SLFO:Main"
-            if os_version in (OsVersion.SL16_0,):
-                first_prj = "SUSE:Registry"
             match os_version:
                 case OsVersion.SL16_0 | OsVersion.SL16_1:
+                    first_prj = "SUSE:Registry"
```
dirkmueller, line 162: not sure we want to build in released mode given that 16.1 is not yet released. 

I would prefer to go back to SLFO:1.3 as last project like discussed in https://github.com/SUSE/BCI-dockerfile-generator/pull/4303#discussion_r3851428342

alexandrevicenzi: While not released, the images are in the registry in tech preview, so this, in theory, works.

We could also go back to SLFO:1.3, it should work because it removes the duplicated entries.

dirkmueller: the issue is that there is only irregular snapshots, like once a month or so, which means we do not see problems as they get introduced, but only after the snapshot gets released. that's why we build against codestreams for unreleased so that we can find problems *before* they get introduced. 

made https://github.com/SUSE/BCI-dockerfile-generator/pull/4324 to go back to codestream builds for unreleased.

alexandrevicenzi: good point

alexandrevicenzi:
Closing this in favor of https://github.com/SUSE/BCI-dockerfile-generator/pull/4324.

PR #4322, opened 2026-08-27
dotnet: Drop coreutils and openssl-3

alexandrevicenzi:
These dependencies are already installed in bci-base.


dirkmueller:
The build service cannot detect them being there and then you get choice errors or worse, BusyBox packages for dependency resolution.  So that was intentional

dirkmueller:
imho a feature and blocks mismatches between base container and devel container. so lets not merge this (it is onyl a problem when the openssl stack in the base container is broken, and not releasing the devel container is a good thing)

PR #4342, opened 2026-08-31
Remove golang-openssl from SLFO

dirkmueller:
We'll stop providing it with go 1.27 onwards, and 1.25/1.26 are currently published from SLE15


alexandrevicenzi (approved):

alexandrevicenzi:
Can we also publish something in the [announcments](https://github.com/SUSE/bci/discussions)?

dirkmueller:
> Can we also publish something in the [announcments](https://github.com/SUSE/bci/discussions)?

sure, but the exact wording is not yet clear (and should be positive, so rather than removing, we're adding fips native support..)

PR #4349, opened 2026-09-01
feat: add opencode sandbox

Lioli7k:
Add opencode sandbox image in two flavors:
  base - hardened image containing just opencode and basic tools.
  devel - tumbleweed based development image.


dirkmueller:
thanks!

dirkmueller (approved):

PR #4352, opened 2026-09-01
golang: include the new FIPS 140-3 mode

alexandrevicenzi:
Add information on how to use the new FIPS 140-3 mode.
Add a migration notice for Go-openssl.

### DIFF
--- a/src/bci_build/package/golang/README.md.j2
+++ b/src/bci_build/package/golang/README.md.j2
@@ -19,6 +19,41 @@ for cryptographic operations, if available at runtime.
 Therefore, you can use FIPS 140-2/140-3 validated routines, provided by the OpenSSL
 library, for cryptographic operations in the container environment.
 
+{%- else %}
+
+## FIPS 140-3
+
+Go binaries can natively operate in a mode that facilitates
+[FIPS 140-3 Compliance](https://go.dev/doc/security/fips140).
+
+### The GOFIPS140 environment variable
+
+The `GOFIPS140` environment variable can be used with `go build`, `go install`,
+and `go test` to enable FIPS 140-3 mode.
+
+The recommended values are `off` (default value) or `certified` to select the
+corresponding NIST CMVP certified release.
+
+### The fips140 GODEBUG option
+
+The `fips140` option in `GODEBUG` controls FIPS 140-3 mode during runtime execution.
+
+It defaults to `off` unless `GOFIPS140` was set at build time.
+
+The possible values for `fips140` are:
+
+- **off**: The Go Cryptographic Module does not operate in FIPS 140-3 mode.
+- **on**: The Go Cryptographic Module operates in FIPS 140-3 mode.
+- **only**: The Go Cryptographic Module operates in FIPS 140-3 mode, but cryptographic algorithms not approved by FIPS 140-3 return an error or panic.
+
+### Migration from Go-openssl
+
+Starting with Go 1.27, the Go-openssl variant—which routes cryptographic
+operations through OpenSSL—will no longer be supported in {{ short_bci }}.
+
+To maintain compatibility with prior Go-openssl behavior, during build time
+set `GOFIPS140` to `certified` or set `fips140=on` during runtime in `GODEBUG`.
+
 {%- endif %}
 
 


src/bci_build/package/golang/README.md.j2
```diff
@@ -19,6 +19,28 @@ for cryptographic operations, if available at runtime.
 Therefore, you can use FIPS 140-2/140-3 validated routines, provided by the OpenSSL
 library, for cryptographic operations in the container environment.
 
+{%- else %}
+
+## FIPS 140-3
+
+Since Go 1.24, a new `fips140` setting facilitates
```
dirkmueller: I think that information isn't erlevant since we will only include it in 1.26 and newer :-) would be better to explain that SUSEs golang fips certification

src/bci_build/package/golang/README.md.j2
```diff
@@ -19,6 +19,28 @@ for cryptographic operations, if available at runtime.
...
+[FIPS 140-3 Compliance](https://go.dev/doc/security/fips140).
+
+The possible values for `fips140` are:
+
+- **auto**: FIPS activates only when the host is in FIPS mode (added in 1.27).
```
dirkmueller: lets put that behind a if version != "1.26"

src/bci_build/package/golang/README.md.j2
```diff
@@ -19,6 +19,28 @@ for cryptographic operations, if available at runtime.
...
+- **off**: The Go Cryptographic Module does not operate in FIPS 140-3 mode.
+- **on**: The Go Cryptographic Module operates in FIPS 140-3 mode.
+- **only**: The Go Cryptographic Module operates in FIPS 140-3 mode, but cryptographic algorithms not approved by FIPS 140-3 return an error or panic.
+
+**NOTE:** The default value for `fips140` since 1.27 is `auto`, and from 1.24 to 1.26 is `off`.
```
dirkmueller: we  could render the README accordingly..

src/bci_build/package/golang/README.md.j2
```diff
@@ -19,6 +19,28 @@ for cryptographic operations, if available at runtime.
...
 library, for cryptographic operations in the container environment.
 
+{%- else %}
+
+## FIPS 140-3
+
+Since Go 1.24, a new `fips140` setting facilitates
+[FIPS 140-3 Compliance](https://go.dev/doc/security/fips140).
+
+The possible values for `fips140` are:
```
dirkmueller, line 43: I think the variable is called GOFIPS140 and it would be nice to explain that it is an environment variable to be set

src/bci_build/package/golang/README.md.j2
```diff
@@ -19,6 +19,28 @@ for cryptographic operations, if available at runtime.
...
+- **on**: The Go Cryptographic Module operates in FIPS 140-3 mode.
+- **only**: The Go Cryptographic Module operates in FIPS 140-3 mode, but cryptographic algorithms not approved by FIPS 140-3 return an error or panic.
+
+**NOTE:** The default value for `fips140` since 1.27 is `auto`, and from 1.24 to 1.26 is `off`.
+
+### Migration from Go-openssl
+
+Starting with Go 1.27, the Go-openssl variant—which routes cryptographic operations through OpenSSL—will no longer be supported in {{ short_bci }}.
+
+To maintain compatibility with prior Go-openssl behavior, set `fips140=auto` in `GODEBUG`.
```
dirkmueller: should this be referring to GOFIPS instead?

src/bci_build/package/golang/README.md.j2
alexandrevicenzi: Good point, I missed this option.

src/bci_build/package/golang/README.md.j2
alexandrevicenzi: Good point.

src/bci_build/package/golang/README.md.j2
alexandrevicenzi: Good point.

src/bci_build/package/golang/README.md.j2
alexandrevicenzi: I will double-check. I think both options are supported.

src/bci_build/package/golang/README.md.j2
alexandrevicenzi: Updated.

src/bci_build/package/golang/README.md.j2
alexandrevicenzi: Both options are supported, and I tried to capture and document both.

src/bci_build/package/golang/README.md.j2
alexandrevicenzi: Documented GOFIPS140.

src/bci_build/package/golang/README.md.j2
```diff
@@ -19,6 +19,45 @@ for cryptographic operations, if available at runtime.
...
+
+### Migration from Go-openssl
+
+Starting with Go 1.27, the Go-openssl variant—which routes cryptographic
+operations through OpenSSL—will no longer be supported in {{ short_bci }}.
+
+To maintain compatibility with prior Go-openssl behavior, during build time
+set `GOFIPS140` to `certified` or set `fips140=on` during runtime in `GODEBUG`.
+
+Unlike OpenSSL, the native Go FIPS has no support for automatic detection of
```
alexandrevicenzi: I am not sure if I should call automatic detection, that does not feel 100% correct, but I don't have a better idea.

Go does not care if the host is FIPS-enforcing or not.

I can leave this entire section off, if it sounds better.

dirkmueller (approved):

PR #4402, opened 2026-09-09
fix: remove /dev/null

alexandrevicenzi:
A few images ship /dev/null, and if a command pipes to /dev/null, the file can differ between builds.

### DIFF
--- a/src/bci_build/package/__init__.py
+++ b/src/bci_build/package/__init__.py
@@ -52,7 +52,8 @@ LOG_CLEAN: str = textwrap.dedent("""rm -rf {/target,}/var/log/{alternatives.log,
     rm -f {/target,}/usr/lib/sysimage/rpm/.rpm.lock; \\
     rm -f {/target,}/var/lib/zypp/AnonymousUniqueId; \\
     rm -f {/target,}/var/lib/zypp/AutoInstalled; \\
-    rm -f {/target,}/var/cache/ldconfig/aux-cache
+    rm -f {/target,}/var/cache/ldconfig/aux-cache \\
+    rm -f {/target,}/dev/null
 """)
 
 #: Rebuild the rpm database to make it reproducible


dirkmueller:
so these are normal files then rather than a devnode? removing the devnode sounds wrong to me (and likely fails?).

alexandrevicenzi:
> so these are normal files then rather than a devnode? removing the devnode sounds wrong to me (and likely fails?).

I did check a few images, and `/dev` is empty on most of the images, except for `/dev/null` on some images, and at least on Kea, the file differs between builds.

The [OCI spec](https://specs.opencontainers.org/runtime-spec/config-linux/#RUNTIME-SPEC-CONFIG-LINUX-23) defines these devices as something that should be provided by the runtime.

alexandrevicenzi:
@dirkmueller it seems that the Kea issue might be solved by #4403.

If that is the case, we might not need to remove `/dev/null` from all images, but I don't think there is any harm in removing it, since it looks like `/dev/null` is just a text file and not a real device.

dirkmueller:
> @dirkmueller it seems that the Kea issue might be solved by #4403.

ok, great. 

> If that is the case, we might not need to remove `/dev/null` from all images, but I don't think there is any harm in removing it, since it looks like `/dev/null` is just a text file and not a real device.

right, we should have a test that alerts us of unexpected text files in /dev rather than just deleting it (papering over the problem). we need to fix the root causes underneath instead. 

also was pondering about testing all application containers with --read-only, just to be sure.

alexandrevicenzi:
> > @dirkmueller it seems that the Kea issue might be solved by #4403.
> 
> ok, great.
> 
> > If that is the case, we might not need to remove `/dev/null` from all images, but I don't think there is any harm in removing it, since it looks like `/dev/null` is just a text file and not a real device.
> 
> right, we should have a test that alerts us of unexpected text files in /dev rather than just deleting it (papering over the problem). we need to fix the root causes underneath instead.
> 
> also was pondering about testing all application containers with --read-only, just to be sure.

I can investigate if this is some expected behavior, because I find it a bit odd to be honest, `/dev/null` should not be a text file, but also not clear what in the build could be causing this.

I can write a test case, this will at least show all images that have the file, I found a few manually, but unclear how many are affected and if there are any other files in `/dev`.

dirkmueller:
imho: let's fix the rootcause of the cases where /dev/null is being created.

alexandrevicenzi:
> imho: let's fix the rootcause of the cases where /dev/null is being created.

yes, it is on my TODO list to investigate.

PR #4403, opened 2026-09-09
kea: require grep

alexandrevicenzi:
grep is used by the package post scripts but fails silently.

### DIFF
--- a/src/bci_build/package/kea.py
+++ b/src/bci_build/package/kea.py
@@ -29,7 +29,7 @@ KEA_DHCP_CONTAINERS = [
         version_in_uid=False,
         support_level=SupportLevel.L3,
         pretty_name="Kea DHCP Server",
-        package_list=sorted(["kea", "sed", "util-linux"]),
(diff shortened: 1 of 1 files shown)


alexandrevicenzi:
Found this while looking at `/dev/null` inside the image, which contained: `/var/tmp/rpm-tmp.0q1ICU: line 18: grep: command not found`

This most likely causes the diff in the file that is removed by #4402.

dirkmueller:
> grep is used by the package post scripts but fails silently.

did you fix the package as well?

dirkmueller (approved):

dirkmueller:
> > grep is used by the package post scripts but fails silently.
> 
> did you fix the package as well?

submitted as https://src.opensuse.org/dhcp/kea/pulls/17

alexandrevicenzi:
> > > grep is used by the package post scripts but fails silently.
> > 
> > 
> > did you fix the package as well?
> 
> submitted as https://src.opensuse.org/dhcp/kea/pulls/17

Thanks, I did not, because I was going to test the image first.

PR #4416, opened 2026-09-10
feat: allow to disable the container support

alexandrevicenzi:


alexandrevicenzi:
This seems to solve the logs from Python and PHP images, but I wonder if we should disable it for all application/developer images, since they don't install anything on the final layer, AFAIK.

dirkmueller (commented):
code wise this looks fine (although I'm not a big fan of obs_disable vs disable_obs in the naming), but I'm wondering if we should rather avoid having RUN statement in the final stage. 

here it is just about creating two symlinks, which we can perfectly do at build stage and then the issue is gone as well.

alexandrevicenzi:
> code wise this looks fine (although I'm not a big fan of obs_disable vs disable_obs in the naming), but I'm wondering if we should rather avoid having RUN statement in the final stage.
> 
> here it is just about creating two symlinks, which we can perfectly do at build stage and then the issue is gone as well.

For some reason, the PHP containers still have logs, and there is no RUN command. I can investigate more and try to solve by changing the image if possible.

I am open to better naming, I can think a bit more.

dirkmueller:
> For some reason, the PHP containers still have logs, and there is no RUN command. I can investigate more and try to solve by changing the image if possible.

ok, would be good to get to the bottom of this. I'd be also fine with always disabling dockersupport in the last stage of a multistage build, and then fix the fallout (there might be 1-2 images affected by this). then we don't need yet another flag.

alexandrevicenzi:
> > For some reason, the PHP containers still have logs, and there is no RUN command. I can investigate more and try to solve by changing the image if possible.
> 
> ok, would be good to get to the bottom of this. I'd be also fine with always disabling dockersupport in the last stage of a multistage build, and then fix the fallout (there might be 1-2 images affected by this). then we don't need yet another flag.

Yes, I suspect that the issue is that `patchdockerfile` seems to take the previous stage into account, and it injects `obs-docker-support` even without `RUN` statements.

In https://github.com/openSUSE/obs-build/blob/master/patchdockerfile#L49 it checks if the previous section did a `--upload-packages`.

PR #4429, opened 2026-09-14
ruby: add micro variant (jsc#PED-16817)

alexandrevicenzi:


dirkmueller (approved):
LGTM. a README update would have been nice but I haven't done so in the other containers either.

alexandrevicenzi:
> LGTM. a README update would have been nice but I haven't done so in the other containers either.

Yes, I was considering reviewing all READMEs. I'll try to do it in another PR.

PR #4434, opened 2026-09-15
php: add micro variant (jsc#PED-16817)

alexandrevicenzi:


src/bci_build/package/php.py
alexandrevicenzi, line 182: Does it make sense to remove Composer for the micro image? Python and Ruby still ship with their package manager.

The other packages, I dont think it makes too much sense to remove, since there are common PHP dependencies.

dirkmueller (approved):
added a few fixes on top to compensate for my delay in a timely review

PR #4443, opened 2026-09-17
ci: check if branches are different

alexandrevicenzi:
Check if target and for-deploy are different instead of using the last commit hash.

This allows to run the job again in case of failures.


alexandrevicenzi:
@dirkmueller I think there's no way to test this until we merge it. The command works locally, it finds a diff between branches.

dirkmueller (approved):

PR #4445, opened 2026-09-18
kubevirt: build the 1.9 stack for SLE 16.0 too

dirkmueller:
kubevirt1.9, containerized-data-importer1.66 and virt-template0.2 are being added to SLE 16.0 (SLFO 1.2), so the three image sets need an SL16_0 row next to their Tumbleweed one. virt-template had no registry override yet: virt-operator derives the virt-template image names from its own registry path, so on 16.0 the two images must be published under suse/sles/16.0 like the KubeVirt ones.

Only KubeVirt 1.8 claims :latest in suse/sles/16.0 today, and nothing changes there: the SL16 rows never set is_latest.


alexandrevicenzi:
packages are missing, is this expected?

dirkmueller:
> packages are missing, is this expected?

yeah, this part is wip

PR #4450, opened 2026-09-18
Avoid bundler on 16.0

dirkmueller:


dirkmueller:
need this to get the testing back to green..

